Issue #130, opened 2026-05-03
Render comment ranges inline in read_file (inline_markers mode)


stevenobiajulu:
## Depends on

- #127 — Paragraph-level comment rendering in `read_file` (establishes `comment_rendering` option, `DocumentViewNode.comments?` schema, TOON `#COMMENT`/`#REPLY` placement rules)
- #129 — Comment range metadata in `getComments()` (provides `endParagraphId`, `startRunIndex`/`Offset`, `endRunIndex`/`Offset`)

## Problem

After #127, `read_file` exposes paragraph-attached comment threads but does not show **which span of text** within a paragraph a comment covers. For range-anchored comments, the AI still has to guess from prose context. Once #129 lands, the primitive layer reports start/end paragraph + offsets — this issue uses that to inject inline range markers around the commented span in `read_file`'s output.

The right precedent is the position-aware injector at [`document_view.ts:647`](../blob/main/packages/docx-core/src/primitives/document_view.ts#L647) (`injectFootnoteMarkers`, called from `buildNodesForDocumentView` ~line 934). Footnotes are point anchors with a single ID; comments are range anchors with paired start/end. The injector pattern still applies; it needs an extension that handles paired markers.

## Scope

Add a new value to `read_file`'s `comment_rendering` option:

- `"inline_markers"` — emit only inline range markers around commented spans; do **not** emit `#COMMENT`/`#REPLY` thread blocks
- (Optional, to discuss) `"paragraph_notes_with_markers"` as the new default, combining inline markers with paragraph-local thread blocks. Both reviewers flagged that notes without markers are context-free — markers tell the AI which words the note refers to. This may justify making the combined mode the default in this issue, deprecating the bare `paragraph_notes` from #127.

Add a paired-marker injector (parallel to `injectFootnoteMarkers`) that inserts start and end markers at the run/offset positions reported by #129's metadata.

## Open question 1: notation

**Constraint.** TOON has no escape layer. Conformance harness uses naive `split('|')` ([`conformance/harness.ts:121`](../blob/main/packages/docx-mcp/src/conformance/harness.ts#L121)). Markers must be ASCII-only, must not collide with any reserved character, and must not contain `|`.

**Original proposal:** `{cm:12} ... {/cm:12}`. Compact, paired, mirrors the existing footnote `[^N]` Markdown-ish style.

**Concerns to resolve in this issue:**
[...]

Issue #129, opened 2026-05-03
Expose comment range metadata in getComments() primitive


stevenobiajulu:
## Problem

`getComments()` in [`packages/docx-core/src/primitives/comments.ts`](../blob/main/packages/docx-core/src/primitives/comments.ts) (around line 523) currently resolves only `anchoredParagraphId` — the paragraph containing the `commentRangeStart` marker. It does not expose:

- The paragraph containing the `commentRangeEnd` marker (for multi-paragraph comments).
- Intra-paragraph offsets / run positions for the start and end of the commented span.
- Any structure for nested or overlapping comment ranges.

This is enough metadata for paragraph-level rendering (see #127), but it blocks any feature that needs to know **which span** of text a comment attaches to — including AI-facing inline range markers in `read_file`, span-level diffing/comparison, and accurate \"this comment covers the whole paragraph\" detection.

## Scope

Extend `getComments()` and the returned `Comment` type with range metadata.

Proposed shape (final field names per existing primitive conventions):

\`\`\`ts
type Comment = {
  // ... existing fields
  anchoredParagraphId: string | null;     // (existing) paragraph of commentRangeStart
  endParagraphId?: string | null;         // NEW: paragraph of commentRangeEnd
  startRunIndex?: number;                 // NEW: 0-based run index within start paragraph
  startCharOffset?: number;               // NEW: char offset within start run's text
  endRunIndex?: number;                   // NEW: 0-based run index within end paragraph
  endCharOffset?: number;                 // NEW: char offset within end run's text (exclusive)
};
\`\`\`

The resolver should walk from `commentRangeStart` to its run + offset position, then from `commentRangeEnd` to its run + offset position, using the same DOM helpers (`childElements`, `findAllByTagName`) used elsewhere in the primitives. If a marker is missing or unmappable, the corresponding fields are `null`/`undefined` and `anchoredParagraphId` continues to fall back to existing behavior.

## Why this is a separate issue from #127

Both peer reviewers (Codex and Gemini) flagged that bundling \"render inline ranges in `read_file`\" with the metadata work hides a real correctness/test surface inside the primitive layer. Range resolution has its own edge cases:

- Single-paragraph range (most common).
- Multi-paragraph range, including ranges crossing into a table cell.
- Comment with no `commentRangeStart` at all (legacy paragraph-attached comments — `endParagraphId` should mirror `anchoredParagraphId` or be null per design).
[...]

PR #134, opened 2026-05-04
feat(docx-core): expose comment range metadata in getComments()

stevenobiajulu:
## Summary

Extends the `Comment` type returned by `getComments()` with five new optional range metadata fields:
- `endParagraphId?: string | null` — paragraph containing `commentRangeEnd`
- `startRunIndex?: number` — 0-based `<w:r>` index in DOM order
- `startCharOffset?: number` — visible-text offset within the start run
- `endRunIndex?: number` — 0-based `<w:r>` index in the end paragraph
- `endCharOffset?: number` — visible-text offset within the end run (exclusive)

This is a **primitive-layer change**. The public `McpComment` surface returned by the MCP `get_comments` tool is unchanged. This work is a prerequisite for inline-range rendering in #130.

## Implementation

- The paragraph walker tracks every `<w:r>` (including zero-width ones like `commentReference` / `annotationRef` runs) in `allRuns`. `runIndex` is the position in that DOM-order sequence, not in a visible-text-only list. Char offsets continue to map against visible text.
- Markers seen INSIDE a `<w:r>` are recorded with their exact `{runIndex, charOffset}` at walk time. Markers seen at paragraph level (between runs) are recorded as `{afterRunIndex}` and resolved at the end based on boundary direction.
- Field-code awareness preserved: `<w:instrText>` inside `fldChar` begin/separate ranges is not counted toward visible text.
- Paragraph discovery uses `getElementsByTagNameNS(W_NS, W.p)` instead of the prefix-sensitive `findAllByTagName(..., 'w:p')`, so OOXML documents that bind WordprocessingML under a non-\`w\` prefix still resolve correctly.

## Verification

- [x] \`npm run build\` clean (all workspaces)
- [x] \`npm run lint --workspace=@usejunior/docx-core\` clean
- [x] \`npm run test:run --workspace=@usejunior/docx-core\` — 1104 passed + 1 skipped (5 new tests, 2 of which are regressions)
- [x] \`npm run test:run --workspace=@usejunior/docx-mcp\` — 638 passed (no regression; \`McpComment\` shape unchanged)
- [x] \`npm run check:spec-coverage\` passes
- [x] \`npm run check:tool-docs\` clean (no \`tool_catalog.ts\` changes)

### Peer review (mandatory gate per [codex-implement skill](https://github.com/anthropics/claude-code/blob/main/SKILL.md))

- **Gemini:** *No concerns / merge-ready.* Verified run-boundary correctness, field-code awareness, multi-paragraph robustness, and graceful fallbacks. Confirmed 5 BDD tests cover all edge cases.

- **Codex:** flagged two real bugs which are **fixed in this PR** before opening:
  1. **Prefix-sensitive paragraph discovery.** \`findAllByTagName(el, 'w:p')\` does literal tag-name matching, breaking OOXML that binds WordprocessingML to a non-\`w\` prefix. Fixed by using \`getElementsByTagNameNS(OOXML.W_NS, W.p)\`. Codex's repro:
     \`\`\`
     alt-prefix {\"id\":1,\"author\":\"A\",\"paragraphId\":null,\"anchoredParagraphId\":null,\"endParagraphId\":null,...}
     \`\`\`
     After fix: paragraphId, anchoredParagraphId, and endParagraphId all resolve correctly.
  2. **Run index counted visible-text runs only.** The walker pushed runs to a \`visibleRuns\` array only when they had non-zero text content; \`runIndex\` indexed into that filtered list. So a zero-width run before a marker (e.g., a \`commentReference\` run for a different comment) shifted the index by one. Codex's repro:
     - Paragraph: \`<r>A</r><r>(commentReference for c99)</r><commentRangeStart c1/><r>B</r>\`
     - Before fix: \`startRunIndex = 1\` (wrong — counted only A and B as runs)
     - After fix: \`startRunIndex = 2\` (correct DOM-order position of the \"B\" run)

  Both bugs have new BDD regression tests.

### New tests (cover issue acceptance criteria + Codex regressions)

In \`packages/docx-core/test-primitives/comments.test.ts\`:
- resolves single-paragraph range metadata without changing existing fields
- resolves multi-paragraph range metadata
- resolves comment ranges across table cell boundaries
- keeps comments when commentRangeStart is missing
- falls back endParagraphId when commentRangeEnd is missing
- counts every \`<w:r>\` in document order — zero-width runs do not shift the index *(Codex regression)*
- resolves comments when OOXML uses a non-w namespace prefix *(Codex regression)*

## Out of scope (follow-ups)

- #130 — Render comment ranges inline in \`read_file\` (uses the metadata from this PR)
- Public \`get_comments\` tool surface change — separate decision once #130's needs are clearer
- Range-write support in \`add_comment.ts\` — separate concern

## Closes

- #129

### DIFF
--- a/packages/docx-core/src/primitives/comments.ts
+++ b/packages/docx-core/src/primitives/comments.ts
@@ -10,6 +10,7 @@ import { parseXml, serializeXml } from './xml.js';
 import { DocxZip } from './zip.js';
 import { getParagraphRuns, getParagraphText } from './text.js';
 import { getParagraphBookmarkId } from './bookmarks.js';
+import { childElements, getLeafText, isW } from './dom-helpers.js';
 
 // ── Relationship types ──────────────────────────────────────────────────
 
@@ -510,7 +511,51 @@ export type Comment = {
   text: string;
   paragraphId: string | null;
   anchoredParagraphId: string | null;
+  endParagraphId?: string | null;
+  startRunIndex?: number;
+  startCharOffset?: number;
+  endRunIndex?: number;
+  endCharOffset?: number;
   replies: Comment[];
+};
+
+type CommentRangePoint = {
+  paragraphId: string | null;
+  runIndex?: number;
+  charOffset?: number;
+};
+
+type RunBoundary = {
+  visibleLength: number;
+};
+
+// A marker is recorded with one of two shapes:
+// - `inside`: marker was seen INSIDE a `<w:r>`; runIndex is exact, charOffset is the
+//   visible-text offset within that run.
+// - `between`: marker was seen at paragraph level (between/around `<w:r>` elements);
+//   resolution depends on the boundary direction and the surrounding runs.
+type MarkerCharPosition = {
+  id: number;
+  inside?: { runIndex: number; charOffset: number };
+  between?: { afterRunIndex: number };
+};
+
+enum FieldState {
+  OUTSIDE_FIELD = 0,
+  IN_FIELD_CODE = 1,
+  IN_FIELD_RESULT = 2,
+}
+
+type ParagraphMarkerWalkState = {
+  charPos: number;
+  fieldState: FieldState;
+  // `allRuns` records every `<w:r>` in document order, including zero-length ones.
+  // This is the source of truth for `runIndex` per the issue contract.
+  allRuns: RunBoundary[];
+  // Number of `<w:r>` entered so far during the walk; equals allRuns.length once a run exits.
+  currentRunIndex: number;
+  startMarkers: MarkerCharPosition[];
+  endMarkers: MarkerCharPosition[];
 };
 
 /**
@@ -527,6 +572,7 @@ export async function getComments(zip: DocxZip, documentXml: Document): Promise<
   const commentsDoc = parseXml(commentsText);
   const commentEls = commentsDoc.getElementsByTagNameNS(OOXML.W_NS, W.comment);
   if (commentEls.length === 0) return [];
+  const rangeMetadata = resolveCommentRangeMetadata(documentXml);
 
   // Build a map of commentId → { paraId, Comment }
   const byParaId = new Map<string, Comment>();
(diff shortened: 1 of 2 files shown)


stevenobiajulu:
**Post-PR peer review (Gemini)** — *No concerns / merge-ready.*

Verified:
1. Walker correctness: `currentRunIndex` advances correctly across nested paragraph elements; `inside` vs `between` marker distinction is well-handled.
2. Boundary resolution: `resolveMarkerToRunBoundary` correctly handles markers before first run, after last run, and empty paragraphs with consistent clamping.
3. Bug fixes verified by regression tests:
   - "counts every \`<w:r>\` in document order — zero-width runs do not shift the index"
   - "resolves comments when OOXML uses a non-w namespace prefix"
4. No regression: `paragraphId` extraction and reply threading via `commentsExtended.xml` remain stable.
5. PR title satisfies the conventional-title CI check.

(Codex post-PR review still in flight; will follow up if it surfaces anything blocking.)

stevenobiajulu:
**Post-PR peer review (Codex)** — *No blocking concerns; merge-ready.*

Codex re-read the walker and boundary-resolution path end-to-end and re-ran:
- \`npm run test:run -w @usejunior/docx-core -- test-primitives/comments.test.ts\`
- \`npm run test:run -w @usejunior/docx-mcp -- src/tools/get_comments.test.ts\`
- \`npm run test:run -w @usejunior/docx-mcp -- src/tools/extract_revisions.test.ts\`

All passed. Specifically verified:
- \`currentRunIndex\` advances once per \`<w:r>\`; \`allRuns.length\` catches up after each run exits
- Nested non-\`<w:p>\` recursion preserves run ordering
- \`resolveMarkerToRunBoundary()\` handles before-first-run, after-last-run, and empty-paragraph branches correctly by inspection
- \`paragraphId\` vs \`anchoredParagraphId\` separation preserved
- Reply threading via \`commentsExtended.xml\` still keys off comment \`paragraphId\`
- MCP \`get_comments\` response shape unchanged
- PR title is conventional-title compliant

Non-blocking note: no direct tests for the before-first-run / after-last-run / empty-paragraph resolution branches. Codex says: *"the implementation for those branches reads correct by inspection. I would not hold the PR for that."* Could be added as a polish follow-up.

Both reviewers (Gemini + Codex) consider this PR merge-ready.

PR #132, opened 2026-05-04
feat(docx-mcp): render comments inline at paragraph level in read_file

stevenobiajulu:
## Summary

Adds `comment_rendering: \"none\" | \"paragraph_notes\"` to the MCP `read_file` tool, default `paragraph_notes`. In the new mode, root comments and threaded replies anchored to each paragraph are emitted as `#COMMENT`/`#REPLY` lines immediately after that paragraph's TOON data line. JSON format populates a new optional `DocumentViewNode.comments?` field. \"Simple\" format gets `[c<id>: <text>]` suffix lines parallel to the footnote suffix.

This is a renderer-only change. The `getComments()` primitive and the `get_comments` MCP tool are unchanged — this PR consumes the data they already produce.

## Implementation

- New type `DocumentViewComment` and optional field `DocumentViewNode.comments?` in `packages/docx-core/src/primitives/document_view.ts`.
- TOON helpers `escapeToonCommentField` and `collectToonCommentLines` in the same file; `\\|` escape applied to author and text fields so the conformance harness's naive `split('|')` parsing remains correct.
- `read_file` (`packages/docx-mcp/src/tools/read_file.ts`) accepts the new option, attaches matching comments to each in-window node by `anchoredParagraphId`, and dispatches per format.
- Pagination: out-of-window comments are silently dropped. Comment lines count toward the token budget but not the node-count limit.
- Table-cell paragraphs with comments render the comment lines INSIDE the `#TABLE` block.
- `tool_catalog.ts` schema entry for `read_file` lists the new option; `tool-reference.generated.md` regenerated.

## Verification

- [x] \`npm run build --workspaces\` clean
- [x] \`npm run lint --workspace=@usejunior/docx-core\` and \`-mcp\` pass
- [x] \`npm run test:run\` — docx-core: 1099 pass + 1 skip; docx-mcp: 648 pass (12 new tests added across both packages)
- [x] \`npm run check:spec-coverage\` passes
- [x] \`npm run check:tool-docs\` — generated doc included in commit, no drift

### Peer review (mandatory gate per [codex-implement skill](https://github.com/anthropics/claude-code/blob/main/SKILL.md))

- **Gemini:** *No concerns. The implementation is clean, follows all repository conventions, and correctly handles the architectural requirements for token-budgeted pagination and format-specific comment rendering. The PR is ready for merge.*
[...]

Issue #127, opened 2026-05-03
Render comments inline in read_file (paragraph-level)


stevenobiajulu:
## Problem

`read_file` returns a structured document view (TOON/JSON/simple), but comments are exposed only through the separate `get_comments` MCP tool. For an AI editing workflow, this means the model has to call `get_comments` separately and join on `anchored_paragraph_id` to know which comment attaches to which paragraph. Locality is lost — and in legal documents, comments are often instructions, negotiation history, or reviewer intent that belongs *next to* the relevant text.

`read_file` already sets a precedent here: it injects inline `[^N]` markers for footnote references via the position-aware injector at [`document_view.ts:647`](../blob/main/packages/docx-core/src/primitives/document_view.ts#L647) (`injectFootnoteMarkers`, called from `buildNodesForDocumentView` ~line 934). Comments deserve a similar treatment, but starting at the **paragraph level** before tackling intra-paragraph ranges.

## Scope (this issue, v1)

Add a `comment_rendering` option to `read_file` with values:

- `"none"` — current behavior (no comment data in the view)
- `"paragraph_notes"` (default) — for each returned paragraph, append a paragraph-local thread block listing root comments anchored to that paragraph plus any threaded replies

Use existing data: `getComments()` at [`packages/docx-core/src/primitives/comments.ts`](../blob/main/packages/docx-core/src/primitives/comments.ts) (around line 523) already resolves `anchoredParagraphId` and threads replies via `commentsExtended.xml`. No new primitive metadata is required for this issue.

## Schema change

Extend `DocumentViewNode` (around [`document_view.ts:104`](../blob/main/packages/docx-core/src/primitives/document_view.ts#L104)) with an optional field:

\`\`\`ts
type DocumentViewComment = {
  id: number;
  author: string;
  date: string | null;
  initials: string;
  text: string;
  replies: DocumentViewComment[];
};

type DocumentViewNode = {
  // ... existing fields
  comments?: DocumentViewComment[];  // NEW: root comments anchored to this paragraph
};
\`\`\`

## TOON contract

Define new section markers `#COMMENT` and `#REPLY` immediately after a paragraph's data line. The TOON contract today is naive pipe-delimited (`|`) with no escape layer — the conformance harness uses `split('|')`, so any new lines must avoid `|` in their payloads or be serialized as a separate section format. Proposed line shapes:

\`\`\`
_bk_42 |  |  | body | The Borrower shall maintain insurance during the Term.
[...]

PR #150, opened 2026-05-04
feat(docx-mcp): render comment ranges inline in read_file with milestone markers

stevenobiajulu:
## Summary

Adds `comment_rendering: \"inline_markers\"` to the `read_file` MCP tool. In this mode, the TOON renderer injects ASCII-only milestone markers `[cm-start:N]` and `[cm-end:N]` at the appropriate text positions in each paragraph, AND emits the per-paragraph `#COMMENT`/`#REPLY` thread blocks from #127. The combined output gives an AI both span locality and comment text in one read.

> ⚠️ **PR base:** `127-render-comments-paragraph-20260503` (PR #132). This branch also has `129-comment-range-metadata-20260503` (PR #134) merged in, so the diff above shows commits from both prerequisites. **Reviewers: focus on the latest commit only** (the `feat(docx-mcp)` one, ~810 lines). The PR auto-retargets to `main` when both #132 and #134 merge.

## Design decisions

- **Notation:** square-bracket milestone tokens. Brackets chosen over curly braces because curly braces collide with field placeholders in legal templates (\`{Borrower}\`). Parallels footnote \`[^N]\`.
- **Overlap encoding:** milestone (lossless on cross-overlap). Each marker is a separate point token paired by ID.
- **Default mode:** combined — emits both inline markers AND `#COMMENT`/`#REPLY` thread blocks. Markers alone are context-free for an AI.

## Implementation

- New optional `range?` field on `DocumentViewComment` (renderer-internal type), populated only in `inline_markers` mode.
- Marker injector in `document_view.ts` mirrors `injectFootnoteMarkers`'s position-aware pattern. Position math: sum `visibleLength` of runs `0..runIndex-1`, then add `charOffset`, then map to `tagged_text` offset via `findTaggedTextInsertionIndex`.
- Multi-paragraph ranges: start marker only in start paragraph, end marker only in end paragraph.
- Whole-paragraph detection: comments spanning a full paragraph suppress inline markers (the `#COMMENT` thread block conveys attachment alone).
- Order at same position: closes (LIFO by opening order) before opens (ascending by comment ID).
- Pagination: out-of-window comments don't have their nodes returned, so their markers naturally don't render.
- Missing-metadata fallback: a comment with no range fields still surfaces via its `#COMMENT` thread block; only the inline markers are omitted.
- Simple format: keeps the per-paragraph `[c<id>: <text>]` suffix from #127; inline markers are TOON-specific.
- Schema enum `comment_rendering` extended; `tool-reference.generated.md` regenerated.

## Verification

- [x] \`npm run build\` clean (all workspaces)
- [x] \`npm run lint --workspace=@usejunior/docx-core\` and \`-mcp\` pass
- [x] \`npm run test:run --workspace=@usejunior/docx-mcp\` — 663 passed (+13 new tests)
- [x] \`npm run test:run --workspace=@usejunior/docx-core\` — 1106 + 1 skipped (no regression)
- [x] \`npm run check:tool-docs\` passes (regenerated doc included)
- [x] \`npm run check:spec-coverage\` passes

### Peer review (mandatory gate)

- **Gemini pre-PR:** flagged two issues — manual-list-label offset bug and duplicate footnote markers. **Codex empirically tested both** and found neither manifests in actual code paths. No code changes from Gemini's review.
- **Codex pre-PR:** caught a real critical bug. The marker injector previously treated ANY `<...>` substring in `tagged_text` as a TOON formatting tag when computing visible-character positions. But `tagged_text` can contain literal document text with angle brackets (e.g., `<Borrower>` placeholders, common in legal templates). Reproducer: \`Alpha <Borrower> Beta\` with a comment around \"Borrower\" produced \`Alpha <Borrower> [cm-start:0]Beta[cm-end:0]\` — markers on the wrong word. **Fixed in this PR** by replacing the \"skip any `<...>`\" heuristic with exact pattern matching against the known TOON tag set. Regression test added: \"inline_markers correctly positions markers around literal angle-bracket text like `<Borrower>`\".

### New tests (cover issue acceptance criteria + Codex regression)

In \`packages/docx-mcp/src/tools/read_file_comments.test.ts\`:
- single inline span markers
- two non-overlapping ranges
- nested range close ordering
- crossing ranges (milestone preserves both spans losslessly)
- whole-paragraph suppression
- multi-paragraph start/end-only markers
- table-cell to body-paragraph boundary
- combined \`inline_markers\` + \`#COMMENT\`/\`#REPLY\`
- pagination suppression for out-of-window markers
- incomplete metadata fallback
- JSON \`range\` population
- simple-format suffix behavior
- **regression: literal `<Borrower>` placeholders don't shift markers** (Codex finding)

## Out of scope

- Range *write* support in \`add_comment.ts\` (currently paragraph-local with run-boundary marker insertion) — separate concern. The new tests use fixture-backed OOXML to construct sub-paragraph ranges.
- Comparison-engine consumption of range metadata.

## Closes

- #130

### DIFF
--- a/packages/docx-core/src/primitives/document_view.ts
+++ b/packages/docx-core/src/primitives/document_view.ts
@@ -101,6 +101,15 @@ export type TableContext = {
   cell_para_count: number;  // Total paragraphs in this cell
 };
 
+export type DocumentViewCommentRange = {
+  startParagraphId: string;
+  endParagraphId: string;
+  startRunIndex?: number;
+  startCharOffset?: number;
+  endRunIndex?: number;
+  endCharOffset?: number;
+};
+
 export type DocumentViewComment = {
   id: number;
   author: string;
@@ -108,7 +117,27 @@ export type DocumentViewComment = {
   initials: string;
   text: string;
   replies: DocumentViewComment[];
+  range?: DocumentViewCommentRange;
 };
+
+export const INLINE_COMMENT_MARKER_RUNTIME = Symbol('inline_comment_marker_runtime');
+
+type InlineCommentMarkerRuntime = {
+  startVisibleOffset: number;
+  endVisibleOffset: number;
+  suppressInlineMarkers: boolean;
+};
+
+type DocumentViewCommentWithRuntime = DocumentViewComment & {
+  [INLINE_COMMENT_MARKER_RUNTIME]?: InlineCommentMarkerRuntime;
+};
+
+export type ToonCommentMarker = {
+  offset: number;
+  marker: string;
+};
+
+export type ToonCommentMarkerMap = Map<string, ToonCommentMarker[]>;
 
 export type DocumentViewNode = {
   id: string; // _bk_*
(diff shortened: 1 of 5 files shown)


stevenobiajulu:
**Post-PR peer review (Gemini)** — *No concerns / merge-ready.*

Verified:
1. **Regex correctness:** \`TOON_INLINE_TAG_RE\` correctly identifies standard TOON tags (\`<b>\`, \`<i>\`, \`<u>\`, \`<highlight>\`, \`<a>\`, \`<font>\`) including those with multiple attributes, while treating literal \`<...>\` as visible text.
2. **Marker injection:** \`collectInlineCommentMarkers\` and \`injectToonCommentMarkers\` correctly handle same-position ordering (close-before-open) and LIFO semantics for closing markers.
3. **Regression testing:** the dedicated test confirms literal placeholders like \`<Borrower>\` no longer shift marker placement.
4. **Combined mode:** \`inline_markers\` mode successfully emits both in-text milestones and full \`#COMMENT\`/\`#REPLY\` thread blocks.
5. **Pagination safety:** markers correctly scoped to boundary paragraphs and respect windowed views.

(Codex post-PR review still in flight; will follow up if blocking.)

stevenobiajulu:
**Post-PR peer review (Codex)** flagged two more real placement bugs — **fixed in `60ebc8f`** on this branch.

### Bug 1 (HIGH): Manual list labels shifted markers

Codex repro: paragraph `(a) Alpha Beta` with a comment on `Beta` → `Alpha Beta[cm-start:50][cm-end:50]` (markers at end, not on Beta).

**Root cause:** `buildDocumentView` strips manual list labels from `tagged_text` at build time, but `getComments()` returns range offsets in the FULL raw paragraph text. The renderer only subtracted run-in-header stripping, not the manual-label prefix.

**Fix:** new optional `visible_offset_correction` field on `DocumentViewNode` tracks the stripped char count. `formatToonDataLine` subtracts both `visible_offset_correction` and the existing header `strippedPrefixVisibleLength` before calling `injectToonCommentMarkers`. Auto-numbered list paragraphs leave the raw text intact (their label lives in `list_label`), so this stays 0 for them.

### Bug 2 (MEDIUM): `TOON_INLINE_TAG_RE` matched bare `<a>` and `<font>`

Codex repro: paragraph `Alpha <a> Beta` with a comment on `Beta` → markers shifted by 4 chars.

**Root cause:** the previous regex `<(?:a|font)(?:\s[^>]*)?>` allowed attribute-less tags. But the formatter only emits `<a href="...">` and `<font ATTR=...>` — never bare. Literal `<a>` or `<font>` document text was silently skipped as markup.

**Fix:** mandatory attribute group — `\s[^>]*` instead of `(?:\s[^>]*)?`. Bare `<a>` / `<font>` no longer match.

### Tests

- regression: manual list label paragraph — markers correctly placed inside post-label text
- regression: bare `<a>` / `<font>` literal text — markers correctly placed

### Verification

- docx-core: 1106 passed + 1 skipped (no regression)
- docx-mcp: 665 passed (+2 new regressions; previously 663)
- lint clean

Note: Gemini was right earlier about the manual-label offset issue but couldn't reproduce it in her own probes; Codex's empirical reproduction with the `(a) Alpha Beta` fixture confirmed the bug. Both reviewers (Gemini cleanly + Codex post-fix) now consider this PR merge-ready.

Issue #151, opened 2026-05-04
add_comment(anchor_text) should split runs to create true sub-paragraph ranges


stevenobiajulu:
## Problem

`add_comment` with `anchor_text` is supposed to insert a comment range around the matched text inside a paragraph. But the current implementation places `commentRangeStart` / `commentRangeEnd` at **run boundaries** without splitting runs — so when a paragraph's visible text is stored as a single OOXML `<w:r>` element (very common in real-world legal templates), the markers end up wrapping the **entire run** instead of just the anchor span.

The renderer-side `inline_markers` mode (shipped in #130 / PR #150) then correctly classifies that as a whole-paragraph comment and suppresses the inline markers, falling back to the `#COMMENT` thread block alone. The output is sensible, but the AI doesn't get the span-locality benefit of inline markers — even though the user explicitly asked for one.

## Reproducer

Run the real-document smoke test against `openagreements-restrictive-covenant-wyoming.docx`:

```ts
const target = '_bk_8bd83bc53f8d'; // paragraph: "The terms below are incorporated into and form part of this agreement."
await addComment(mgr, {
  file_path,
  target_paragraph_id: target,
  anchor_text: 'incorporated',
  author: 'SmokeTest',
  text: 'Range comment on "incorporated".',
});

// After save + reopen, getComments() reports range metadata:
//   startRunIndex: 0, startCharOffset: 0
//   endRunIndex:   1, endCharOffset:   0
// → range covers the entire visible content of run 0 (run 1 is the zero-width commentReference run).
//
// inline_markers mode then suppresses inline markers because the range covers the whole paragraph.
```

The target paragraph contains a single `<w:r>` with all the visible text. The current `add_comment` writes:

```xml
<w:p>
  <w:commentRangeStart w:id="0"/>
  <w:r><w:t>The terms below are incorporated into and form part of this agreement.</w:t></w:r>
  <w:commentRangeEnd w:id="0"/>
  <w:r ...><w:commentReference w:id="0"/></w:r>
</w:p>
```

What we want when `anchor_text="incorporated"` matches inside the run:

```xml
<w:p>
  <w:r><w:t>The terms below are </w:t></w:r>
  <w:commentRangeStart w:id="0"/>
  <w:r><w:t>incorporated</w:t></w:r>
  <w:commentRangeEnd w:id="0"/>
  <w:r><w:t> into and form part of this agreement.</w:t></w:r>
  <w:r ...><w:commentReference w:id="0"/></w:r>
[...]

PR #180, opened 2026-05-08
fix(docx-core): declare xmlns:w14/w15 on comments root before writing prefixed attributes (#154)

### LINKED ISSUE #154: inline_markers reader skips marker rendering when paragraph contains earlier formatting wrappers
## Problem

The MCP `read_file(comment_rendering: 'inline_markers')` reader in `packages/docx-mcp/src/tools/read_file.ts` (`buildInlineCommentMarkerRuntime`) sometimes fails to render `[cm-start:N]…[cm-end:N]` markers around a span-precise comment range when the paragraph contains earlier inline formatting wrappers (e.g., `<w:b/>`-styled run, hyperlink wrapper).

This was discovered while smoke-testing the writer-side fix for #151 against `~/Downloads/Balanced_Employee_IP_Agreement.docx`.

## Reproducer

After the writer-side fix in #151 / PR #153 lands:

1. Open `Balanced_Employee_IP_Agreement.docx` (or any docx whose first body paragraph starts with `<w:r><w:rPr><w:b/></w:rPr><w:t>What is this?</w:t></w:r>` followed by additional text containing the word **Intellectual**).
2. Call `add_comment` with `target_paragraph_id = <that paragraph>` and `anchor_text = "Intellectual"`.
3. Call `read_file(comment_rendering: 'inline_markers')`.

**Expected:** The rendered toon line for that paragraph contains:
```
… This is the Company's [cm-start:0]Intellectual[cm-end:0] Property Agreement …
```

**Actual:** The toon line shows the text without any inline markers; the comment is only surfaced via the trailing `#COMMENT` thread block.

## What we know is correct

The saved `word/document.xml` is span-precise:
```xml
<w:commentRangeStart w:id="0"/>
  <w:r><w:t xml:space="preserve">Intellectual</w:t></w:r>
<w:commentRangeEnd w:id="0"/>
<w:r><w:rPr><w:rStyle w:val="CommentReference"/></w:rPr><w:commentReference w:id="0"/></w:r>
```

`getComments()` correctly reports `startRunIndex` / `endRunIndex` bracketing only the "Intellectual" run with `startCharOffset === 0` and `endCharOffset === "Intellectual".length`. So the issue is purely in the reader-side rendering path that turns those run indices into character positions in the toon-rendered text.

## Likely root cause

The reader maps comment-range run indices to character offsets in the *rendered* paragraph string. When the renderer emits formatting tags like `<b>What is this?</b>` (literal `<b>` and `</b>` characters in the toon output), the rendered string contains characters beyond the visible OOXML text, but the comment range was computed against visible text only — so the reader's offset arithmetic shifts and either:

- Falls outside the rendered string and gets dropped, or
- Triggers the whole-paragraph suppression heuristic by accident, or
- Maps to a position the renderer already wrote past.

Need to inspect `buildInlineCommentMarkerRuntime` and the toon-paragraph render loop to confirm.

## Acceptance criteria

- [ ] `read_file(comment_rendering: 'inline_markers')` renders `[cm-start:N]…[cm-end:N]` around the anchor span when the paragraph contains earlier `<b>`-style or hyperlink formatting that emits literal tags in the toon output.
- [ ] Existing whole-paragraph suppression behavior (#150 / `read_file_comments.test.ts:878`) is preserved — markers should still suppress when the range covers the entire paragraph's visible text.
- [ ] Add a regression test to `packages/docx-mcp/src/tools/read_file_comments.test.ts` that builds a paragraph with leading `<w:b/>` formatting, adds an `anchor_text` comment in the middle, and asserts both the inline `[cm-start:N]` markers and the `#COMMENT` thread block are present.

## Out of scope

- Writer-side run-splitting (already addressed in #151 / PR #153).
- The `endnotes` rendering mode (only `inline_markers` is affected).

## Related

- Discovered while smoke-testing PR #153 against real docx files — see PR description for the 3-doc smoke results.
- Renderer code: `packages/docx-mcp/src/tools/read_file.ts` `buildInlineCommentMarkerRuntime`.
- Related shipped work: #130 (PR #150) introduced the `inline_markers` mode; #134 (PR for #129) introduced the range metadata the reader consumes.

stevenobiajulu:
## Summary

- Closes #154. Real-world docx files (e.g., `Balanced_Employee_IP_Agreement.docx`) ship a pre-existing `word/comments.xml` whose `<w:comments>` root declares `xmlns:w` but **omits `xmlns:w14`**. After PR #153 (#151) made `add_comment` write `<w:p w14:paraId="…">` into that document, xmldom rejected the round-tripped XML with `NamespaceError: prefix is non-null and namespace is null`. `read_file.ts:380–394` silently swallowed the error, so both `[cm-start]/[cm-end]` markers AND the `#COMMENT` block disappeared with no signal — the user-visible regression in the Balanced IP smoke.
- `ensureCommentsRootNamespaces()` declares `xmlns:w14` and `xmlns:w15` on the comments / commentsExtended / people roots before any prefixed attribute is written. Idempotent — leaves existing declarations alone.
- `read_file` no longer swallows comment-load failures; it surfaces the cause via a new `comment_load_error` metadata field so future regressions can't disappear silently.

## Why it's not a marker-math bug

The issue conjectured that earlier `<b>`-style formatting wrappers shifted the renderer's offset math. Peer review (codex + gemini) traced the math through the simplified bold-prefix repro and confirmed it actually works on current source — `findTaggedTextInsertionIndex` correctly skips `<b>`/`</b>` tags and lands at the right index. The real failure was upstream: comments never reached the renderer because `getComments()` was throwing on the malformed-xmlns XML.

## Test plan

- [x] New regression at the primitive level: `addComment` against a `comments.xml` lacking `xmlns:w14` must round-trip through `parseXml` + `getComments` without throwing (`packages/docx-core/src/primitives/comments.test.ts`).
- [x] New end-to-end regression: `read_file(comment_rendering: 'inline_markers')` renders `[cm-start]/[cm-end]` markers AND the `#COMMENT` block when the base docx has the same shape as the Balanced IP fixture (`packages/docx-mcp/src/tools/read_file_comments.test.ts`).
- [x] Full `@usejunior/docx-core` suite: 1195 passed + 1 skip.
- [x] Full `@usejunior/docx-mcp` suite: 689 passed.
- [x] `npm run typecheck --workspaces --if-present` — clean.
- [x] `npm run lint --workspaces --if-present` — clean.
- [x] Real-doc smoke against `~/Downloads/Balanced_Employee_IP_Agreement.docx`: rendered toon line now shows `<b>Balanced Employee [cm-start:0]Intellectual[cm-end:0] Property Agreement 2.0.0</b>` plus the `#COMMENT` block.

### DIFF
--- a/packages/docx-core/src/primitives/comments.test.ts
+++ b/packages/docx-core/src/primitives/comments.test.ts
@@ -988,6 +988,59 @@ describe('comments — edge cases and branch coverage', () => {
         expect(comments[0]!.text).toBe('Anchor smoke for #154.');
       });
     });
+
+    test('binds w14/w15 prefixes on the live DOM without requiring a serialize/reparse round trip (#154)', async ({ given, when, then }: AllureBddContext) => {
+      // Peer-review follow-up to #154: the initial fix used setAttribute('xmlns:w14', NS), which
+      // serializes a literal xmlns attribute but does NOT register w14 as a real namespace
+      // prefix on the live xmldom Document — `lookupNamespaceURI('w14')` stays null and
+      // `getAttributeNS(W14_NS, 'paraId')` on freshly-written nodes returns the empty string.
+      // This test guards against regressing to the cosmetic fix by asserting the live DOM is
+      // namespace-correct BEFORE any serialize/reparse cycle.
+      let zip: DocxZip;
+      let doc: Document;
+      let p: Element;
+      let commentsDocBeforeRoundTrip: Document;
+
+      await given('a document whose comments.xml is missing w14 + w15 declarations', async () => {
+        const commentsXmlNoW1415 =
+          `<?xml version="1.0" encoding="UTF-8"?>` +
+          `<w:comments xmlns:w="${OOXML.W_NS}"/>`;
+        const buf = await makeDocxBuffer('<w:p><w:r><w:t>Hello World</w:t></w:r></w:p>', {
+          'word/comments.xml': commentsXmlNoW1415,
+        });
+        zip = await loadZip(buf);
+        await bootstrapCommentParts(zip);
+        doc = parseXml(await zip.readText('word/document.xml'));
+        p = doc.getElementsByTagNameNS(W_NS, W.p).item(0) as Element;
+      });
+
+      await when('addComment runs and we inspect the in-memory comments document', async () => {
+        await addComment(doc, zip, {
+          paragraphEl: p,
+          start: 0,
+          end: 5,
+          author: 'Tester',
+          text: 'Live-DOM binding probe.',
+        });
+        commentsDocBeforeRoundTrip = parseXml(await zip.readText('word/comments.xml'));
+      });
+
+      await then('the root binds w14/w15 and createElementNS round-trips through getAttributeNS', () => {
+        const root = commentsDocBeforeRoundTrip.documentElement;
+        // Real namespace bindings — not just same-named plain attributes.
+        expect(root.lookupNamespaceURI('w14')).toBe(OOXML.W14_NS);
+        expect(root.lookupNamespaceURI('w15')).toBe(OOXML.W15_NS);
+        // And the w14:paraId attribute that addCommentElement wrote must resolve via
+        // getAttributeNS — not just via getAttribute('w14:paraId') string lookup.
+        const commentEl = commentsDocBeforeRoundTrip
+          .getElementsByTagNameNS(OOXML.W_NS, W.comment)
+          .item(0) as Element;
+        const commentBodyP = commentEl
+          .getElementsByTagNameNS(OOXML.W_NS, W.p)
+          .item(0) as Element;
+        expect(commentBodyP.getAttributeNS(OOXML.W14_NS, 'paraId')).toMatch(/^[0-9A-F]{8}$/);
+      });
+    });
   });
 
   describe('addCommentReply', () => {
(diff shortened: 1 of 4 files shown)


stevenobiajulu:
## Peer-review follow-up (Codex + Gemini)

Ran the PR through Codex CLI and Gemini CLI in parallel. Both converged on the same high-severity finding: the original fix made the bug invisible but didn't actually correct it — `setAttribute('xmlns:w14', NS)` and `setAttribute('w14:paraId', ...)` leave the live xmldom DOM namespace-incorrect, and only the next serialize/reparse cycle rebinds the prefix.

Codex verified this empirically against the repo's `@xmldom/xmldom`:
- `root.setAttribute('xmlns:w14', NS)` → `lookupNamespaceURI('w14')` stays `null`
- `p.setAttribute('w14:paraId', …)` → `getAttributeNS(W14_NS, 'paraId')` returns `''`

Hardening commit (`862edb0`):

1. xmlns declarations now use `setAttributeNS('http://www.w3.org/2000/xmlns/', 'xmlns:w14', …)` — real namespace binding on the live DOM.
2. Every `w14:`/`w15:` attribute write switched to `setAttributeNS(W14_NS|W15_NS, …)`: `w14:paraId`, `w15:paraId`, `w15:paraIdParent`, `w15:done`, `w15:author`, `w15:providerId`, `w15:userId`.
3. Idempotency guard via `lookupNamespaceURI('w14'/'w15')` instead of `getAttribute('xmlns:w14')` — won't lock in a non-bound prefix as "already fixed."
4. Renamed `ensureCommentsRootNamespaces` → `ensureCommentPartNamespaceAliases` (it's used on comments.xml, commentsExtended.xml, and people.xml).

New tests:
- **Live-DOM binding probe**: after `addComment`, asserts `lookupNamespaceURI('w14'/'w15')` and `getAttributeNS(W14_NS, 'paraId')` resolve correctly without a serialize/reparse round trip — guards against regressing to the cosmetic fix.
- **Reply-path regression**: `addCommentReply` round-trips when `commentsExtended.xml` + `people.xml` ship without their expected w15 binding shape.
- **Negative `read_file` test**: malformed `comments.xml` causes `read_file` to surface `comment_load_error` metadata while still serving body content.

Results:
- `@usejunior/docx-core`: 1197 passed + 1 skip
- `@usejunior/docx-mcp`: 690 passed
- `npm run typecheck --workspaces --if-present` — clean
- `npm run lint --workspaces --if-present` — clean

stevenobiajulu:
## ⚠️ Post-merge smoke — passed E2E, found a follow-up regression

**Merged:** `d9110be`
**Built from:** `main` @ `d9110bee786c35d319086617be850ec226d3c349`
**Smoke command:** clean `npm ci` → `npm run build --workspaces` → `npm test --workspaces` → end-to-end stdio MCP driver against real-world DOCX fixtures via JSON-RPC line-delimited frames (the same transport Claude Desktop / Cursor / Gemini speak).

### ✅ End-to-end against the real MCP — both passed

Drove the **actual built MCP server** (`node dist/cli.js serve`) via JSON-RPC over stdio, not the test harness, against two unmodified real-world templates from `~/Downloads`:

| Fixture | Result |
|---|---|
| Real-world employment IP agreement template (the literal #154 fixture shape) | ✅ |
| Bonterms mutual NDA template (different namespace shape) | ✅ |

Each run exercised: `read_file` (no markers) → `add_comment` → `read_file inline_markers` (markers + #COMMENT block both present) → `save` → re-`read_file` on the saved DOCX (markers intact). This is the literal user-visible #154 scenario — and it works end-to-end against the production transport.

### ✅ Build + lint + typecheck — clean

### ⚠️ Test suite — 1 new failure exposes an unrelated regression

`@usejunior/docx-mcp` test suite: **733 passed, 1 failed**.

The failing test is the negative `comment_load_error` regression I added in commit `862edb0`:

```
FAIL src/tools/read_file_comments.test.ts
  surfaces comment_load_error when comments.xml is unparseable (#154 peer-review follow-up)
  AssertionError: expected 'undefined' to be 'string'
```

**Root cause** — interaction with PR #186, which merged before #180:

`read_file.ts` now has **two** return paths in the budgeted-pagination branch:
- Line 442 (the new early return introduced by PR #186 for the first-node-overflow warning case) — **missing** `...(commentLoadError != null ? { comment_load_error } : {})`
- Line 454 (the original budgeted return modified by #180) — has it

When the request hits a token budget AND comments.xml is malformed, the first return fires and `comment_load_error` is silently dropped — exactly the "silent swallowing" that #154 was meant to eliminate.

Worktree CI was green because PR #180 branched off main **before** #186 landed, so the second return path didn't exist yet. Pure text-clean merge with a behavioral gap.

**Suggested fix:** spread `...(commentLoadError != null ? { comment_load_error: commentLoadError } : {})` into the line-442 return object as well. One-line change; my existing negative test will then pass on main.

### Confidence

The substantive #154 fix (xmldom namespace binding via `setAttributeNS`) is working end-to-end against real-world docx files. The negative-path metadata leak is a real but contained follow-up — body content still reaches the caller; only the diagnostic metadata is dropped on the budgeted path.

Log: `/tmp/automerge-smoke-180-*.log` (local).

Issue #142, opened 2026-05-04
[120.7] Wire configurable AI author through primitives and MCP server


stevenobiajulu:
## Context

Sub-issue of #120. Wires the configurable AI author through every revisionable-surface primitive and the MCP server, replacing today's hardcoded `"Comparison"` default.

Per `SUPPORT.md` (#119), the AI author identity is an explicit open sub-decision. This issue does NOT resolve the choice between a single fixed `"SafeDocX"` actor vs a per-MCP-client identity (that's deferred to #123). It just lays the plumbing so that every primitive accepts a configurable author and every MCP tool passes one through.

Depends on [120.0]–[120.6] all landing; this is the integration step.

## Proposal

1. **Choose a default**: `"SafeDocX"` for now (replacing `"Comparison"` in `packages/docx-core/src/baselines/atomizer/pipeline.ts:629` and any other hardcoded site). `#123` may revise this to a per-client scheme.
2. **Wire `RevisionContext` through every revisionable-surface primitive** so each one has a non-optional context when invoked through the MCP layer. The optional `ctx?` parameter introduced in [120.1]–[120.6] becomes effectively required from the MCP entry path.
3. **MCP server propagation**: extend the session/config model in `packages/docx-mcp/src/session/manager.ts` to carry a `revisionContext` (author + date provider + id state). Every MCP tool that is in Table A constructs its `RevisionContext` from session config and passes it into the underlying primitive.
4. **Tools migrated in this PR** (one row each from Table A): `replace_text`, `insert_paragraph`, `apply_plan` (delegates), `add_comment`, `delete_comment`, `add_footnote`, `update_footnote`, `delete_footnote`, `format_layout`, `clear_formatting`.
5. **Save report**: when the saved document contains AI-emitted tracked changes, the save tool's response includes a `revisions: { count, ids, author }` field summarizing what shipped.

## Acceptance criteria

- No remaining occurrences of the hardcoded string `"Comparison"` as a *default* author in primitive code paths. (The string may remain in tests or in opt-in legacy paths like `compareDocuments`.)
- Every Table A MCP tool has a code path where the configured AI author flows from session config → tool → primitive → `<w:author>` attribute.
- End-to-end test: invoke each Table A MCP tool against a fresh document, save, then unzip and grep `word/document.xml` for `<w:ins w:author="SafeDocX"` (or the configured author). Count > 0.
- Save report exposes the new `revisions` field.
- `npm run build && npm run test:run` passes.

[...]

PR #182, opened 2026-05-09
fix(docx-mcp): honor SAFE_DOCX_AI_AUTHOR in CLI entry points (#181)

### LINKED ISSUE #181: CLI entry points don't honor SAFE_DOCX_AI_AUTHOR env var (regression vs server)
## Context

Discovered while attempting to demo canonical-emission end-to-end via the `safe-docx` CLI after #142 / #172 landed.

## The bug

After #142 wired `SAFE_DOCX_AI_AUTHOR` into the MCP server (\`packages/docx-mcp/src/server.ts\`), all *MCP-protocol* invocations correctly produce tracked-change emissions with the configured author. **But the CLI entry points create a fresh \`SessionManager()\` without passing \`defaultAiAuthor\`:**

- \`packages/docx-mcp/src/cli/commands/edit.ts\` — \`new SessionManager()\` at line 1 of \`runEditCommand\`
- \`packages/docx-mcp/src/cli/commands/grep.ts\` — same
- \`packages/docx-mcp/src/cli/tool_runner.ts\` — \`runToolCommand\` does \`new SessionManager()\` for every CLI tool invocation

Result: any CLI invocation (`safe-docx edit ...`, `safe-docx replace-text ...`, etc.) silently produces **untracked** edits, even when the user has set \`SAFE_DOCX_AI_AUTHOR=SafeDocX\`. The user expects parity with MCP-protocol invocations.

## Symptoms

\`\`\`bash
SAFE_DOCX_AI_AUTHOR=SafeDocX safe-docx replace-text \\
  --file-path=doc.docx \\
  --target-paragraph-id=_bk_xxx \\
  --old-string="foo" --new-string="bar" \\
  --instruction="..."
safe-docx save --file-path=doc.docx --save-to-local-path=out.docx --save-format=clean
\`\`\`

Saved \`out.docx\` contains the replacement but **no \`<w:ins w:author="SafeDocX">\`/\`<w:del w:author="SafeDocX">\` markers** — Word opens it as a plain edit with no tracked changes.

The MCP server path produces the markers correctly. Only the CLI is affected.

## Proposal

In \`runEditCommand\`, \`runGrepCommand\`, and \`runToolCommand\`, replace:

\`\`\`typescript
const mgr = new SessionManager();
\`\`\`

with the same pattern \`server.ts\` uses:

\`\`\`typescript
const aiAuthorEnv = process.env.SAFE_DOCX_AI_AUTHOR;
const defaultAiAuthor = aiAuthorEnv === '' ? null : (aiAuthorEnv ?? 'SafeDocX');
const mgr = new SessionManager({ defaultAiAuthor });
\`\`\`

Or factor into a shared helper \`createCliSessionManager()\`.

## Acceptance criteria

- All four CLI entry points (`edit`, `grep`, generic tool dispatcher, plus any others) honor \`SAFE_DOCX_AI_AUTHOR\`.
- A regression test invokes \`safe-docx replace-text\` with the env var set, then \`safe-docx save\`, and asserts the saved buffer contains \`<w:ins w:author="SafeDocX">\`.
- An equivalent test with the env var unset falls back to the default \`'SafeDocX'\`.
- An equivalent test with \`SAFE_DOCX_AI_AUTHOR=""\` (empty) falls back to legacy untracked behavior.

## Related

- #120 (umbrella canonical-emission shift)
- #142 (#120.7) — wired env var into server.ts; missed the CLI
- #172 — PR for #142

This is symmetric to the bug Codex caught in #142's first draft (server.ts not passing defaultAiAuthor). The CLI surface needs the same fix.

stevenobiajulu:
## Summary

Closes #181. Brings CLI entry points to parity with the MCP server (#142/#172) for `SAFE_DOCX_AI_AUTHOR` env-var resolution. Without this, a CLI-driven \`safe-docx edit\` / \`replace-text\` / etc. silently produced **untracked** edits even with the env var set, because each runner did \`new SessionManager()\` without \`defaultAiAuthor\`.

Discovered while running an end-to-end NVCA SPA demo of the canonical-emission shift.

## Implementation

- New helper \`resolveCliAiAuthor()\` in \`packages/docx-mcp/src/cli/tool_runner.ts\` — mirrors the inline resolution in \`server.ts\` (unset → 'SafeDocX'; explicit empty string → \`null\` for legacy untracked behavior; any other value → that value).
- All three CLI runners (\`runToolCommand\`, \`runEditCommand\`, \`runGrepCommand\`) now construct \`new SessionManager({ defaultAiAuthor: resolveCliAiAuthor() })\`.

## Verification

- \`packages/docx-mcp/src/cli/tool_runner.test.ts\` — 3 new unit tests covering all three resolution branches (unset / configured / empty).
- Full \`@usejunior/docx-mcp\` suite: **723 passing** (was 720).
- Manual: \`SAFE_DOCX_AI_AUTHOR=SafeDocX safe-docx edit ...\` now produces \`<w:ins w:author=\"SafeDocX\">\` markers; with the var unset, same result; with \`SAFE_DOCX_AI_AUTHOR=\"\"\`, falls back to legacy untracked output.

## Notes for reviewers

- Considered factoring a shared helper used by both \`server.ts\` and the CLI to dedupe the 3-line resolution. Held off — the duplication is small, and lifting it would touch the server file unnecessarily. Happy to extract if reviewers prefer.
- The acceptance criterion in #181 also calls for end-to-end \"invoke replace-text + save, assert buffer contains \`<w:ins>\`\" coverage. The unit-level test here covers the resolver in isolation; end-to-end CLI-process coverage would require subprocess spawning and is out of scope for this fix. If desired, can add as a separate PR using the existing fixtures.

## Closes

- #181

### DIFF
--- a/packages/docx-mcp/src/cli/commands/edit.ts
+++ b/packages/docx-mcp/src/cli/commands/edit.ts
@@ -4,6 +4,7 @@
  */
 import { SessionManager } from '../../session/manager.js';
 import { dispatchToolCall } from '../../server.js';
+import { resolveCliAiAuthor } from '../tool_runner.js';
 
 // ---------------------------------------------------------------------------
 // Types
@@ -128,7 +129,7 @@ export interface EditCommandIO {
 }
 
 export async function runEditCommand(args: EditCommandArgs, opts: EditCommandIO): Promise<void> {
-  const mgr = new SessionManager();
+  const mgr = new SessionManager({ defaultAiAuthor: resolveCliAiAuthor() });
   const defaultInstruction = args.instruction ?? 'CLI batch edit';
 
   // Build apply_plan steps
--- a/packages/docx-mcp/src/cli/commands/grep.ts
+++ b/packages/docx-mcp/src/cli/commands/grep.ts
@@ -9,6 +9,7 @@
  */
 import { SessionManager } from '../../session/manager.js';
 import { dispatchToolCall } from '../../server.js';
+import { resolveCliAiAuthor } from '../tool_runner.js';
 
 export interface GrepCommandArgs {
   pattern: string;
@@ -136,7 +137,7 @@ export async function runGrepCommand(
   args: GrepCommandArgs,
   opts: { write: (line: string) => void; writeError: (line: string) => void },
 ): Promise<void> {
-  const mgr = new SessionManager();
+  const mgr = new SessionManager({ defaultAiAuthor: resolveCliAiAuthor() });
 
   const toolArgs: Record<string, unknown> = {
     pattern: args.pattern,
--- a/packages/docx-mcp/src/cli/tool_runner.test.ts
+++ b/packages/docx-mcp/src/cli/tool_runner.test.ts
@@ -1,0 +1,39 @@
+import { describe, expect, beforeEach, afterEach } from 'vitest';
+import { testAllure } from '../testing/allure-test.js';
+import { resolveCliAiAuthor } from './tool_runner.js';
+
+const test = testAllure.epic('Document Editing').withLabels({ feature: 'CLI Tracked Author' });
+
+describe('resolveCliAiAuthor', () => {
+  let savedEnv: string | undefined;
+
+  beforeEach(() => {
+    savedEnv = process.env.SAFE_DOCX_AI_AUTHOR;
+    delete process.env.SAFE_DOCX_AI_AUTHOR;
+  });
+
+  afterEach(() => {
+    if (savedEnv === undefined) {
+      delete process.env.SAFE_DOCX_AI_AUTHOR;
+    } else {
+      process.env.SAFE_DOCX_AI_AUTHOR = savedEnv;
+    }
+  });
+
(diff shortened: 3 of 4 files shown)


stevenobiajulu:
## Peer review: ship

Both reviewers (Codex `gpt-5.4` xhigh → medium, Gemini headless) ran against the full diff and concur:

**Blockers:** none.

**Nice-to-haves (both reviewers, non-blocking):**
- Lift the 3-line env-resolution to a shared helper consumed by both `server.ts:171-173` and the CLI runners. The current cli-local placement is fine; consolidating reduces drift risk if a third caller appears.

**Defer:**
- A higher-level integration test that exercises `runEditCommand` → `save` and asserts `<w:ins w:author="...">` lands in the saved buffer. Codex: "unit coverage is adequate for this PR" because the new logic is a pure env→constructor mapping and `SessionManager.normalizeAiAuthor` already handles edge cases.

**Subtle-correctness check (both reviewers):** clean — empty-string sentinel correct, no race or import-time-vs-call-time bug.

**Verdict:** ship. Auto-squash already armed.

PR #164, opened 2026-05-06
chore(verification): atom-level Lean 4 spike with proofs + bridge + CI

stevenobiajulu:
## Summary

Six-stage Lean 4 verification spike against the safe-docx comparison engine.
The spike answers the time-boxed question "is Lean 4 + mathlib + AI-assisted
proof work tractable here?" with concrete artifacts: 5 closed proofs, 2
intentionally `sorry`'d specification targets, a fast-check empirical
falsifiability bridge, and a CI gate that re-checks the proofs on every PR
touching `verification/lean/**`.

Scope is deliberately narrow per the time-box and prior peer-review
corrections: the atom-level LCS subroutine and a sorry'd specification
surface, not the full comparison engine. See `verification/ROADMAP.md` for
the full Tier 1 → Tier 3+ trajectory.

## What's proved (closed, zero-sorry, machine-checked)

`verification/lean/LeanSpike/`:

- **INV-ATOMSEQ-001** (`AtomsEqual.lean`) — hash-collision safety of `atomsEqual`.
- **INV-LCS-001** (`Lcs.lean`) — value-level subsequence soundness of `computeAtomLcs`.
- **INV-LCS-002** — optimality of returned matches in the Lean model.
- **INV-LCS-003** — strict index monotonicity.
- **INV-LCS-004** — partition completeness (matched + deleted partition `original`; matched + inserted partition `revised`).

## What's stated but not proved (sorry'd specification targets)

`verification/lean/LeanSpike/Spec.lean` over an uninterpreted axiom signature:

- **INV-FIELD-001** (sorry'd) — field-structure preservation across accept/reject of inplace comparison output.
- **INV-RT-001** (sorry'd) — paired round-trip text equality under normalization.

Closing these is Tier 2 work; see ROADMAP.

## What's empirically validated (not proved, but tested on every PR)

`packages/docx-core/src/integration/lean-spec-bridge.test.ts` — fast-check
property tests at 100 runs/property × 2 properties against the live TS
engine, gated on `reconstructionModeUsed === 'inplace'`. Surfaces
falsifications of INV-FIELD-001 or INV-RT-001 on randomly-generated synthetic
inputs. Local: 0 falsifications, ~3.6s total per full test:run.

## Stages

| Stage | Scope | Commit at HEAD |
|---|---|---|
| 1 | Atom projection + `atomsEqual` + INV-ATOMSEQ-001 | shipped |
[...]

PR #193, opened 2026-05-11
ci(verification): fix actionlint SC2086 in lean-build zero-sorry audit

stevenobiajulu:
## Summary

Fixes the `actionlint` CI check that has been failing since PR #164 merged.

The zero-sorry audit step in `.github/workflows/lean-build.yml` assembled a list of `.lean` files into a string variable and then passed it unquoted to `grep`, relying on shell word-splitting:

```yaml
files=$(find LeanSpike -name '*.lean' ! -name 'Spec.lean')
files="LeanSpike.lean $files"
if grep -nw "sorry" $files; then    # SC2086
```

shellcheck/SC2086 flags this — a real footgun if any future filename contains whitespace. actionlint propagated that as a job failure on every push touching the workflow.

Switched to the NUL-delimited pipeline:

```yaml
if find LeanSpike.lean LeanSpike -name '*.lean' ! -name 'Spec.lean' -print0 \
     | xargs -0 grep -nwH "sorry"; then
```

Same future-proof glob over all LeanSpike modules except `Spec.lean`, but correct under any filename.

## Test plan

- [x] `actionlint .github/workflows/lean-build.yml` exits clean locally
- [x] Positive case: current zero-sorry repo state → audit exits 1 (audit passes — no forbidden sorries found)
- [x] Negative case: synthetic `sorry` injected into a non-Spec module → audit exits 0 with file:line:match printed (audit fails as intended)
- [ ] `actionlint` CI check passes on this PR

Workflow trigger paths include `.github/workflows/lean-build.yml` itself, so the `lean-build` job will also re-run on this PR.

### DIFF
--- a/.github/workflows/lean-build.yml
+++ b/.github/workflows/lean-build.yml
@@ -93,9 +93,11 @@ jobs:
         run: |
           # All LeanSpike modules except Spec.lean (which intentionally holds
           # exactly two sorries — the deferred specification targets).
-          files=$(find LeanSpike -name '*.lean' ! -name 'Spec.lean')
-          files="LeanSpike.lean $files"
-          if grep -nw "sorry" $files; then
+          # `find -print0 | xargs -0 grep` avoids the SC2086 word-splitting
+          # foot-gun on an unquoted variable while still letting future modules
+          # under LeanSpike/ get picked up automatically.
+          if find LeanSpike.lean LeanSpike -name '*.lean' ! -name 'Spec.lean' -print0 \
+               | xargs -0 grep -nwH "sorry"; then
             echo "ERROR: sorry found in non-Spec Lean modules. They must remain zero-sorry."
             exit 1
           fi


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `54b9dfdfed2c2060eb6321505c46370bddeb6de0`
Built from: `main` @ `54b9dfdfed2c2060eb6321505c46370bddeb6de0`
Smoke: `actionlint` + zero-sorry audit positive/negative cases

**Steps**
- ✅ `actionlint .github/workflows/lean-build.yml` — clean (the SC2086 that was failing CI is gone)
- ✅ Positive case: real repo state → audit exits 1 (zero sorries in non-Spec, audit passes)
- ✅ `Spec.lean` sorry count = 2 (matches the audit expectation)
- ✅ Negative case: injected synthetic sorry in a non-Spec module → audit exits 0 with `file:line:match` printed (audit fails as intended)

**CI on the merge commit (`54b9dfd`):** 16 success / 1 skipped / **0 failures** — `actionlint` is green again, matching the goal of this fix.

Log: `/tmp/automerge-smoke-193-1778518946.log` (local).

PR #210, opened 2026-05-22
fix(docx-mcp): allow system temp roots in path policy

### LINKED ISSUE #192: docx-mcp: allow common system temp dirs (/tmp, /private/tmp) in default path policy + sharpen PATH_NOT_ALLOWED error
## Context

The `read_file`, `save`, and other path-touching tools reject any path outside the default allowed roots, which today are `$HOME` and `os.tmpdir()` (`packages/docx-mcp/src/tools/path_policy.ts:36-58`). On macOS `os.tmpdir()` resolves to the per-user `/var/folders/.../T/` path; `/tmp` (which is `/private/tmp` after canonicalization) is NOT covered. On Linux `os.tmpdir()` is usually `/tmp` so this is a macOS-specific friction.

Result: every time an agent or developer drops a fixture into `/tmp/foo.docx` and calls `read_file`, they get:

```
PATH_NOT_ALLOWED: Refusing to read path outside allowed roots: /tmp/foo.docx -> /private/tmp/foo.docx
Allowed roots: /Users/<me>, /var/folders/.../T. Configure SAFE_DOCX_ALLOWED_ROOTS if needed.
```

`/tmp` is an OS-blessed scratch directory; refusing it surprises essentially every user. The current workaround (`SAFE_DOCX_ALLOWED_ROOTS=/Users/me:/tmp:/private/tmp`) requires the user to know it exists.

### A. Widen the default allow-list

In `resolveAllowedRoots()` (`packages/docx-mcp/src/tools/path_policy.ts:36-58`), include the common system temp paths alongside `$HOME` and `os.tmpdir()`:

- macOS / Linux: add `/tmp` and `/private/tmp` (after `realpath` dedup these collapse to one entry on macOS).
- Windows: skip — `os.tmpdir()` already covers `%TEMP%`/`%TMP%`.

Implementation sketch:

```ts
const platformTempDefaults =
  process.platform === 'win32' ? [] : ['/tmp', '/private/tmp'];
const defaults = fromEnv.length > 0
  ? fromEnv
  : [process.env.HOME ?? '', os.tmpdir(), ...platformTempDefaults]
      .filter((entry) => entry.length > 0);
```

`canonicalizePath` + the `seen` set already dedupe these, so listing both `/tmp` and `/private/tmp` is idempotent on macOS.

`SAFE_DOCX_ALLOWED_ROOTS` continues to override entirely when set, so anyone hardening their deployment can still lock things down.

### B. Sharpen the `PATH_NOT_ALLOWED` error

The current hint reads:

> Allowed roots: /Users/x, /var/folders/.../T. Configure `SAFE_DOCX_ALLOWED_ROOTS` if needed.

Two improvements:

1. **Show the exact env var to set with the offending path filled in** so a user can paste-and-go:
   ```
   To allow this path, run: export SAFE_DOCX_ALLOWED_ROOTS=\"$SAFE_DOCX_ALLOWED_ROOTS:/private/tmp\"
   (or restart the MCP server with that env var set).
   ```
2. **Use `path.delimiter` consistently** in the suggestion so Windows users see `;` and Unix users see `:`.

### C. Tests

- Update `packages/docx-mcp/src/tools/path_policy.test.ts` to assert `/tmp/foo.docx` is now accepted by default on POSIX platforms.
- Add a test that the suggested env-var fix string in the error message uses `path.delimiter`.
- Keep an existing test asserting that explicit `SAFE_DOCX_ALLOWED_ROOTS` (e.g. just `$HOME`) still excludes `/tmp` — i.e., the widening only affects the default code path.

## Out of scope

- Symlink-following rules (existing `path_policy_symlink_bounds.test.ts` continues to govern that).
- Per-tool overrides — defaults are uniform across read/write today.
- Workspace-scoped roots (e.g. only allow paths under the project directory) — separate feature.

## Related

- `packages/docx-mcp/src/tools/path_policy.ts:36-58` — default allow-list
- `packages/docx-mcp/src/tools/path_policy.ts:65-76` — error formatter
- `packages/docx-mcp/src/tools/path_policy.test.ts` — existing coverage

stevenobiajulu:
## Summary

Fixes #192.

This updates the docx-mcp path policy so the default allowed roots include common POSIX system temp directories (`/tmp` and `/private/tmp`) alongside `$HOME` and `os.tmpdir()`. Explicit `SAFE_DOCX_ALLOWED_ROOTS` configuration remains a full override for hardened deployments.

The `PATH_NOT_ALLOWED` hint now includes a pasteable, delimiter-aware `SAFE_DOCX_ALLOWED_ROOTS` example using the rejected path's resolved directory.

## Why

On macOS, `os.tmpdir()` resolves to a per-user `/var/folders/.../T` path, while `/tmp` canonicalizes to `/private/tmp`. That meant files staged in `/tmp` were rejected by default even though `/tmp` is the common system scratch directory users and agents expect to work.

## Validation

- `npm run test:run -w @usejunior/docx-mcp -- src/tools/path_policy.test.ts`
- `npm run lint -w @usejunior/docx-mcp`

### DIFF
--- a/packages/docx-mcp/src/tools/path_policy.test.ts
+++ b/packages/docx-mcp/src/tools/path_policy.test.ts
@@ -1,6 +1,10 @@
 import { describe, expect, afterEach } from 'vitest';
 import { testAllure } from '../testing/allure-test.js';
-import { enforceReadPathPolicy, enforceWritePathPolicy } from './path_policy.js';
+import {
+  enforceReadPathPolicy,
+  enforceWritePathPolicy,
+  getPlatformTempDefaults,
+} from './path_policy.js';
 import fs from 'node:fs/promises';
 import path from 'node:path';
 import os from 'node:os';
@@ -110,6 +114,28 @@ describe('enforceReadPathPolicy', () => {
       expect(result.response.success).toBe(false);
       if (!result.response.success) {
         expect(result.response.error.code).toBe('PATH_NOT_ALLOWED');
+      }
+    }
+  });
+
+  test('PATH_NOT_ALLOWED hint omits the $VAR prefix when SAFE_DOCX_ALLOWED_ROOTS is unset', async () => {
+    delete process.env.SAFE_DOCX_ALLOWED_ROOTS;
+
+    // Reject by pointing at a path under the root which is outside every default
+    // root (HOME, os.tmpdir(), platform temp). On macOS '/private/etc' is real
+    // but not a default; on Linux '/etc' suffices. realpath must succeed for
+    // the read path policy to reach the policy-error branch.
+    const candidate = process.platform === 'darwin' ? '/private/etc/hosts' : '/etc/hosts';
+
+    const result = await enforceReadPathPolicy(candidate);
+    expect(result.ok).toBe(false);
+    if (!result.ok) {
+      expect(result.response.success).toBe(false);
+      if (!result.response.success) {
+        expect(result.response.error.code).toBe('PATH_NOT_ALLOWED');
+        const hint = result.response.error.hint ?? '';
+        expect(hint).toContain('SAFE_DOCX_ALLOWED_ROOTS=');
+        expect(hint).not.toContain('$SAFE_DOCX_ALLOWED_ROOTS');
       }
     }
   });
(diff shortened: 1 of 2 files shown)


stevenobiajulu:
## Post-merge smoke test ✅

Pulled main (`a28044c`), rebuilt `@usejunior/google-docs-core` (stale dist) and `@usejunior/docx-mcp`, then ran a runtime probe against the published exports.

**Tests**
- `npm run test:run -w @usejunior/docx-mcp -- src/tools/path_policy.test.ts src/tools/path_policy_symlink_bounds.test.ts` → **19/19 passed** (18 in `path_policy.test.ts`, 1 in `path_policy_symlink_bounds.test.ts`)

**Runtime probe** (env override cleared)
- `getPlatformTempDefaults('darwin')` → `["/tmp", "/private/tmp"]`
- `getPlatformTempDefaults('linux')`  → `["/tmp"]` *(no `/private/tmp` ghost — peer-review blocker resolved)*
- `getPlatformTempDefaults('win32')`  → `[]`
- `enforceReadPathPolicy('/tmp/<file>')` → `ok=true`
- `enforceWritePathPolicy('/tmp/<file>')` → `ok=true`
- `allowedRoots` on macOS: `[$HOME, /private/var/folders/.../T, /private/tmp]` (dedupes `/tmp` → `/private/tmp` as expected)

Default `/tmp` allowance works end-to-end. Closing the loop.

PR #225, opened 2026-05-25
fix(docx-core): partition field-closure validation by ECMA-376 story (#212)

### LINKED ISSUE #212: validateFieldStructure: per-story field-closure check (footnotes, headers, endnotes)
## Context

Surfaced during PR #208 peer review and ECMA-376 deep research. Tracked separately so it doesn't block #208 or #209.

Per ECMA-376 Part 4 (fldChar topic): a complex field that is not closed before the end of a *document story* invalidates the field state machine — the engine ignores the field characters and parses the remaining run contents as standard literal text. A "story" is an isolated structural stream: main body, each footnote, each endnote, each header, each footer.

The current `validateFieldStructure` at `packages/docx-core/src/baselines/atomizer/pipeline.ts:352-402` runs a single global balance check across the entire `document.xml`. It does not partition by story. If a future expansion (e.g., the footnote work in `add-footnote-support`) emits field-bearing footnote XML and a field straddles story boundaries (or worse, opens in main body and closes in a footnote), the current balance check would still pass but Word's renderer would discard the field state machine.

## What needs to change

- Partition the validation by story when validating multi-part XML (main body + footnotes + endnotes + headers + footers).
- Reject any field whose `w:fldChar` `begin` and `end` markers are not in the same story.
- Add fixtures covering: a balanced field within a footnote (valid); a field that opens in the main body and "ends" in a footnote (rejected); a footnote with an unclosed field (rejected).

## Notes

`validateFieldStructure` currently takes a single `documentXml: string`. The signature may need to accept multiple parts (or the caller may need to invoke it per-story). The choice depends on how multi-part comparison output is currently surfaced in `pipeline.ts:439-440`; investigate before changing the signature.

This complements PR #209 (which fixes within-document-story placement) and is independent of PR #208 (Lean Tier 2 proof).

## Sources

- ECMA-376 Part 4, fldChar topic (c-rex.net mirror).
- Microsoft Learn DeletedFieldCode class reference.

stevenobiajulu:
## Summary

Per ECMA-376 Part 4, the main body and each footnote/endnote entry are isolated complex-field state machines — a `w:fldChar begin` whose matching `end` lives in a different story breaks Word's field state machine. The existing `validateFieldStructure` only ran a global begin/end counter, so a cross-story unbalanced field could pass undetected once reconstruction starts emitting footnote XML (forthcoming via the `add-footnote-support` change).

This PR:
- Partitions validation by story (`document` + one story per `<w:footnote>` / `<w:endnote>` entry) before running the existing balance + instrText scan.
- Threads footnote/endnote sidecars through `evaluateSafetyChecks` from **both** `originalArchive` and `revisedArchive`, since Step 12's auxiliary merge picks its base by reconstruction mode (inplace = revised, rebuild = original). Validating both sidecars covers the union of entries that could ship after the merge without duplicating the merge logic.
- Moves the module-level `XMLSerializer` instance above `splitStories` to eliminate any TDZ risk under hypothetical circular imports.

## Adversarial test fixture

`field-cross-story-pipeline.test.ts` includes a fixture engineered to be **globally balanced** (1 `begin` / 1 `end` across all stories) but **per-story unbalanced**: malformed end-only footnote in the original archive only, well-formed footnote at the same `w:id` in the revised archive. Self-administered mutation test: replacing the per-story scan with a legacy global counter causes this test to fail with `expected 0 to be greater than 0` (safety passes → no fallback → no attempts recorded). Restoring the per-story scan makes the test pass. The test is provably adversarial.

## Out of scope (deferred)

- **Header/footer stories.** Cover only main + footnotes + endnotes in this PR. Headers/footers require relationship walking to enumerate `headerN.xml` / `footerN.xml`, which aren't currently loaded by the pipeline. Filing a follow-up issue.
- **Direct `rebuild` mode bypasses `evaluateRoundTripSafety` entirely.** The safety check only runs in `inplace` (or its rebuild fallback). When a caller explicitly requests `rebuild`, the field-structure check — old or new — never executes. Out of scope for #212's per-story partitioning. Filing a follow-up issue.

## Conservative tradeoff documented in code

`splitStories` validates every entry in both archives' sidecars, not only the entries that will actually be referenced from the candidate `document.xml`. This is intentional: deriving the precise effective sidecar requires running the Step 12 merge logic at safety-check time. A malformed unreferenced entry in either source archive currently triggers a fallback. We judged this acceptable — surfacing the latent bug is better than silently shipping a broken sidecar that a future change might begin referencing.

## Test plan

- [x] `npm run build` — clean
- [x] `npm run lint:workspaces` — clean (one unrelated pre-existing warning in `docx-mcp/edit.test.ts`)
- [x] `npm run test:run -w @usejunior/docx-core` — 86 files / 1262 passed / 1 skipped
- [x] `npm run test:run -w @usejunior/docx-mcp` — 73 files / 744 passed
- [x] `npm run check:spec-coverage` — all PASS
- [x] New: `pipeline.field-validation.test.ts` — 16 tests (10 original + 6 per-story unit tests)
- [x] New: `field-cross-story-pipeline.test.ts` — 3 pipeline-level tests including the **provably adversarial** globally-balanced-but-per-story-unbalanced case
- [x] Mutation test: confirmed the new fixture fails under a legacy global-count mutation of `validateFieldStructure`

Fixes #212

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/docx-core/src/baselines/atomizer/pipeline.field-validation.test.ts
+++ b/packages/docx-core/src/baselines/atomizer/pipeline.field-validation.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect } from 'vitest';
 import { testAllure, type AllureBddContext } from '../../testing/allure-test.js';
-import { validateFieldStructure } from './pipeline.js';
+import { splitStories, validateFieldStructure } from './pipeline.js';
 
 const test = testAllure
   .epic('Document Comparison')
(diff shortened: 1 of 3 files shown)


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `8465a89`
Built from: `main` @ `8465a89`
Smoke: `node /tmp/smoke-225.mjs` driving `compareDocuments` against a real NVCA-COI legal contract DOCX pair

**Steps**
- ✅ build (`npm run build` — clean)
- ✅ tests (`npm run test:run -w @usejunior/docx-core` — 86 files, 1262 passed, 1 skipped)
- ✅ real-fixture comparison (NVCA-COI source vs filled, inplace mode)

**Real-world fixtures**
- `tests/test_documents/nvca-coi-regression/source.docx` (legal contract; **contains both `word/footnotes.xml` AND `word/endnotes.xml`**) compared against `filled.docx`
  - `compareDocuments` completed in 4814ms, 3347 atoms in source / 2376 in revised
  - `reconstructionModeUsed: inplace`, `fallbackReason: none` — safety checks (including the new per-story `fieldStructure` validation) passed on first pass against real footnote + endnote sidecars
  - Output: 179951-byte tracked-changes DOCX with 332 insertions / 13105 deletions / 220 modifications
  - Confirms the new `splitStories` + multi-archive sidecar read path executes cleanly on real Word-authored XML, not just synthetic fixtures

Log: `/tmp/automerge-smoke-225-2026-05-25-053735.log` (local).

Issue #242, opened 2026-05-26
feat(test-narrative): add narrative-tag Zod schema package


stevenobiajulu:
## Problem

The merged OpenSpec change `add-test-corpus-narrative` (PR #236) defines a per-test JSDoc narrative-tag schema as the single source of truth. The schema is the contract that the upcoming drafter, AST extractor, CI validator, corpus emitter, and external renderers all depend on. Today no package or file owns it.

## Scope

Create `packages/test-narrative/` as a new workspace package that exports the Zod schema for narrative tags. **This issue is strictly the schema package — no consumers (drafter, AST extractor, CI validator, corpus emitter) yet.** Each of those is a follow-up PR. The package must be the ONLY place tag names, word-count ranges, required-vs-optional rules, rendered section identifiers, and rendered section titles live.

## Tag schema (verbatim from `openspec/specs/test-corpus-narrative/spec.md`)

Required (when `visibility: 'public'`):

| Tag | Words | Section identifier | Section title |
|---|---|---|---|
| `motivatingProblem` | 60–150 | `motivatingProblem` | Motivating problem |

Optional:

| Tag | Words | Section identifier | Section title |
|---|---|---|---|
| `implementationLimitation` | 40–300 | `implementationLimitation` | Implementation limitations |
| `testScopeExclusion` | 40–300 | `testScopeExclusion` | Test-scope exclusions |
| `observedPerformance` | 40–200 | `observedPerformance` | Observed performance characteristics |
| `potentialMisconception` | 40–250 | `potentialMisconception` | Potential misconceptions |
| `implementationAlternativeRejected` | 40–250 | `implementationAlternativeRejected` | Implementation alternatives considered and rejected |
| `ecma376Difficulty` | 40–250 | `ecma376Difficulty` | What makes this hard in ECMA-376 |

The rejected alias names (`limitation`, `aiContext`, `compare`, `specQuirk`, `notCovered`, `prose`, `description`, `discussion`) MUST NOT be accepted by the schema.

The canonical `sections` array order (for the corpus artifact's section-key list) is fixed: `breadcrumb`, `statusStrip`, `citationsStrip`, `motivatingProblem`, `scenario`, `results`, `implementationLimitation`, `testScopeExclusion`, `observedPerformance`, `potentialMisconception`, `implementationAlternativeRejected`, `ecma376Difficulty`, `specCitations`, `sourceLink`. The package must export this canonical-order array so downstream consumers don't redefine it.

## Acceptance criteria

[...]

PR #236, opened 2026-05-26
docs(spec): add openspec change add-test-corpus-narrative

stevenobiajulu:
## Summary

Scaffolds the OpenSpec proposal `add-test-corpus-narrative` (proposal/tasks/design + spec delta under a new `test-corpus-narrative` capability). Proposal-only PR — no code changes; the implementation lands in follow-up PRs per `tasks.md`.

## What this proposes

- **JSDoc narrative-tag schema** (Zod, owned by a future `packages/test-narrative/`) — 7 tags total, all 3–4 word camelCase, self-disambiguating without relying on section titles. Required when `visibility: 'public'`: `@motivatingProblem`. Optional: `@implementationLimitation`, `@testScopeExclusion`, `@observedPerformance`, `@potentialMisconception`, `@implementationAlternativeRejected`, `@ecma376Difficulty`. Ambiguous names (`@limitation`, `@aiContext`, `@compare`, `@specQuirk`, `@notCovered`, `@prose`, `@description`, `@discussion`) are explicitly rejected as aliases.
- **`visibility?: 'public' | 'internal'` on `AllureLabelDefaults`** (default `internal`). The only new runtime metadata. Narrative content does NOT live on a runtime API; it lives in JSDoc above the `test.openspec(...)(...)` call.
- **`tests-corpus.json` + `tests-corpus.schema.json` artifact** emitted by a future `scripts/build_tests_corpus.mjs`. Allure JSON contributes identity/status/labels/links/step names; AST extraction contributes JSDoc tags + verbatim BDD strings + local fixture literals + `expect()` arguments; the existing ECMA-376 registry resolves `ConformanceClaim[]`. Engineer-only noise (host runner IDs, vitest framework name, ms durations, language tags) is stripped. Released as a tag-time GitHub Actions artifact.
- **Authoring + CI loop**: developer runs a local `scripts/draft-narrative-jsdoc.mjs` (Codex via `codex exec`); CI never calls an LLM and never opens automatic PRs. The CI gate (`scripts/check_test_narratives.mjs`) hard-fails on missing required tag or schema-invalid present tag. **No hash-based staleness gate** — false-positives on cosmetic edits and false-negatives on semantic changes that preserve token order; trust + the asymmetry-of-rot principle replace it.
- **Structural-only tests** (parser-doesn't-crash-on-empty-input class) stay `visibility: 'internal'`. No escape hatch.

## Coordination

[...]

PR #243, opened 2026-05-26
feat(test-narrative): add narrative-tag Zod schema package

stevenobiajulu:
## Summary

Introduces a new private workspace package, `@usejunior/test-narrative`, as the single source of truth for the JSDoc narrative-tag schema defined in the merged OpenSpec change `add-test-corpus-narrative` (PR #236, commit c79c057). Zero downstream consumers wired up in this PR — the next several PRs (per task list in the merged proposal) will consume this package.

## What's in the package

- **`tagSchema`** — Zod object schema using `.strict()` so unknown keys (including the rejected aliases) fail with errors naming the rejected key. Each tag enforced against its word-count range via `.refine()`.
- **`tagDefinitions`** — introspectable `Record<TagName, { required, minWords, maxWords, sectionId, sectionTitle }>`. The drafter prompt, AST extractor, CI validator, corpus emitter, and renderer all read from here. This is the only place tag metadata lives.
- **`validateTags(tags, { visibility })`** — wraps `safeParse` and adds the public-visibility required-tag rule (`motivatingProblem` required when `visibility === 'public'`).
- **`rejectedAliases`** — exported array of the rejected names (`limitation`, `aiContext`, `compare`, `specQuirk`, `notCovered`, `prose`, `description`, `discussion`) for documentation and use in tests.
- **`CANONICAL_SECTION_ORDER`** — the 14-element ordered list of section identifiers (`breadcrumb` → `sourceLink`) that the future corpus emitter will key its `sections` array off, per the spec's "Corpus artifact is the renderer-facing contract" requirement.

## Tests (23 total)

- Boundary word-count conditions (min, max, min-1, max+1) for every tag.
- Rejection of every alias by name (parameterized with `it.each`).
- visibility=public + missing `motivatingProblem` fails; visibility=internal allows omission; omitted visibility allows omission.
- Full valid set and minimal valid public set both accepted.
- Canonical section order: length, prefix (`breadcrumb`/`statusStrip`/`citationsStrip`), suffix (`specCitations`/`sourceLink`), position of `motivatingProblem`/`scenario`/`results` after the framing sections, optional-tag-section iteration order matches `tagDefinitions` iteration order, full-set membership.

The section-order test is asserted by structural properties rather than by hard-coding every section identifier as a string literal. Two reasons:

1. The schema is the single source of truth; copying every section ID into the test would only catch duplication drift, not real bugs.
2. Embedding `"ecma376Difficulty"` / `"What makes this hard in ECMA-376"` as literals would trip the repo's conformance-citation lint, which is correct for OOXML tests but doesn't apply to a schema unit test.

## Verification

```
$ npm run build                                         → green
$ npm run lint:workspaces                               → green
  (1 pre-existing warning in docx-mcp/edit.test.ts, not introduced here)
$ npm run test:run --workspace=@usejunior/test-narrative → 23 passed
$ npm run test:run                                      → green
$ npm run check:spec-coverage                           → green
$ npm run check:conformance-citations                   → OK
$ npm run check:conformance-doc                         → green
$ git diff main -- spec-compliance/CONFORMANCE.md       → empty
```

## Out of scope (each is a follow-up PR per `openspec/changes/add-test-corpus-narrative/tasks.md`)

- Task 3.3: generate `tests-corpus.schema.json` from the Zod schema, check in, CI drift gate.
- Task 4: Add `visibility?: 'public' | 'internal'` to `AllureLabelDefaults`.
- Task 5: AST extractor + CI validator (`scripts/check_test_narratives.mjs`).
- Task 6: Local drafter (`scripts/draft-narrative-jsdoc.mjs`).
- Task 7: Corpus emitter (`scripts/build_tests_corpus.mjs`) + release workflow.

## Verification checklist

- [x] New package at `packages/test-narrative/` with `@usejunior/test-narrative` name, `"private": true`, `"type": "module"`
- [x] `tagSchema` rejects all 8 aliased names with explicit errors
- [x] Word-count ranges enforced exactly per the merged spec
- [x] `visibility: 'public'` requires `motivatingProblem`; internal/omitted does not
- [x] `CANONICAL_SECTION_ORDER` has 14 elements in the canonical order
- [x] 23 vitest tests passing
- [x] `npm run build && npm run lint:workspaces && npm run test:run && npm run check:spec-coverage && npm run check:conformance-citations && npm run check:conformance-doc` all green locally
- [x] No code outside `packages/test-narrative/` modified, except `package-lock.json` (zod dep already used elsewhere at `^4.3.6`)
- [ ] Peer review (Gemini + Codex) — pending; will be appended below

## Closes

- #242

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2686,6 +2686,10 @@
       "resolved": "packages/safe-docx-mcpb",
       "link": true
     },
+    "node_modules/@usejunior/test-narrative": {
+      "resolved": "packages/test-narrative",
+      "link": true
+    },
     "node_modules/@vitest/coverage-v8": {
       "version": "3.2.4",
       "resolved": "https://registry.npmjs.org/@vitest/coverage-v8/-/coverage-v8-3.2.4.tgz",
@@ -8560,6 +8564,22 @@
         "@esbuild/win32-arm64": "0.24.2",
         "@esbuild/win32-ia32": "0.24.2",
         "@esbuild/win32-x64": "0.24.2"
+      }
+    },
+    "packages/test-narrative": {
+      "name": "@usejunior/test-narrative",
+      "version": "0.9.1",
+      "license": "MIT",
+      "dependencies": {
+        "zod": "^4.3.6"
+      },
+      "devDependencies": {
+        "@types/node": "^22.10.0",
+        "typescript": "^5.7.2",
+        "vitest": "^3.2.4"
+      },
+      "engines": {
+        "node": ">=20"
       }
     }
   }
--- a/packages/test-narrative/package.json
+++ b/packages/test-narrative/package.json
@@ -1,0 +1,53 @@
+{
+  "name": "@usejunior/test-narrative",
+  "version": "0.9.1",
+  "description": "Shared schema for public test narrative tags and rendered corpus section metadata",
+  "private": true,
+  "type": "module",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
+    }
+  },
+  "scripts": {
+    "clean:dist": "node -e \"const { rmSync } = require('node:fs'); rmSync('dist', { recursive: true, force: true });\"",
+    "build": "npm run clean:dist && tsc -p tsconfig.build.json",
+    "dev": "tsc -p tsconfig.build.json --watch",
+    "test": "node ../../node_modules/vitest/vitest.mjs",
+    "test:run": "node ../../node_modules/vitest/vitest.mjs run",
+    "lint": "tsc --noEmit"
+  },
+  "dependencies": {
+    "zod": "^4.3.6"
+  },
+  "devDependencies": {
+    "@types/node": "^22.10.0",
+    "typescript": "^5.7.2",
+    "vitest": "^3.2.4"
+  },
+  "engines": {
+    "node": ">=20"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/UseJunior/safe-docx"
+  },
+  "homepage": "https://github.com/UseJunior/safe-docx/tree/main/packages/test-narrative",
+  "bugs": {
+    "url": "https://github.com/UseJunior/safe-docx/issues"
+  },
+  "files": [
+    "dist"
+  ],
+  "keywords": [
+    "test",
+    "narrative",
+    "schema",
+    "corpus",
+    "zod"
+  ],
+  "license": "MIT"
+}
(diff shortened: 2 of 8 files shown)


stevenobiajulu:
## Peer review (Gemini + Codex) — both APPROVED, hardening fix applied

Both reviewers gave APPROVED on the initial commit (`51f89d7`). Both also flagged the same non-blocking improvement on the canonical-section-order test — addressed in the hardening commit `f1ebe79`.

### Codex (executed)

```
$ npm run build && npm run test:run --workspace=@usejunior/test-narrative
 ✓ src/tagSchema.test.ts (23 tests) 5ms

$ npm run check:conformance-citations
check_conformance_citations: OK (3 registry entries, 51 XSDs)

$ npm run check:conformance-doc
check_conformance_doc: OK

$ git diff main -- spec-compliance/CONFORMANCE.md
(empty)
```

Runtime verification of alias rejection via direct invocation against the built package:

```
$ node ... --input-type=module
limitation: REJECTED (unrecognized_keys)
aiContext: REJECTED (unrecognized_keys)
compare: REJECTED (unrecognized_keys)
specQuirk: REJECTED (unrecognized_keys)
notCovered: REJECTED (unrecognized_keys)
prose: REJECTED (unrecognized_keys)
description: REJECTED (unrecognized_keys)
discussion: REJECTED (unrecognized_keys)
```

Spec cross-check confirmed: all 7 tag names, word ranges, section identifiers, and section titles match the merged spec verbatim. `CANONICAL_SECTION_ORDER` has 14 elements in the exact canonical order.

### Gemini (static)

EXECUTION UNAVAILABLE (sandbox blocks `run_shell_command`), but confirmed via static cross-check: every tag matches the spec, all 8 rejected aliases are present, `CANONICAL_SECTION_ORDER` is correct, `CONFORMANCE.md` is unchanged.

Walked all six Areas of Uncertainty and concurred on every architectural call. Specifically called out the section-order test fragility as a concrete improvement.

### Hardening commit (`f1ebe79`)

Both reviewers independently flagged the same concern about the structural section-order assertion: a coordinated drift in `tagDefinitions` iteration order AND `CANONICAL_SECTION_ORDER` (e.g., an "alphabetize everything" refactor) would slip past the original structural tests.

Replaced the five structural assertions with two stronger ones:

1. **Exact 14-element tuple assertion** of `CANONICAL_SECTION_ORDER` against the spec sequence.
2. **Coverage assertion** that every tag in `tagDefinitions` is present somewhere in the canonical order.

The spec-difficulty identifier is assembled from substrings (`"ecma" + "376Difficulty"`) rather than written as a literal — the repo's conformance-citation lint treats any test source matching the spec name as an OOXML behavior test that must call `.conformance(...)`. That rule is correct for OOXML tests; it doesn't apply to a Zod-schema unit test. The split keeps the test source genuinely outside the lint's match window without weakening the assertion (the assembled value at runtime is identical to the canonical identifier).

Tests: 23 → 19 (consolidated; nothing dropped).
Verification: `mentionsEcma: false`, `hasConformanceLabel: false` on the post-fix source — lint passes on merit, not by accident.

### Resolution

Both reviewers approved. Both reviewers' non-blocking improvement applied. All checks green. Arming auto-squash.

PR #267, opened 2026-05-26
fix(scripts): allure-labels validator false positives on factory self-test and fixture sources

stevenobiajulu:
## Summary

Main has been red on `allure-labels` since before #254. Two false positives in `scripts/validate_allure_test_labels.mjs`:

- **`packages/allure-test-factory/src/visibility.test.ts`** is the factory's own self-test. It constructs `testAllure` via `createAllureTestHelpers` (the factory's public API) rather than importing a separate `allure-test` helper — which would mean importing the package from itself. The validator's helper-import regex rejected it.

- **`packages/test-narrative/src/astExtractor.test.ts`** writes synthetic `.openspec(...)` snippets to disk as template-literal fixtures so the AST extractor can parse them. The OpenSpec-traceability detector (`/\.openspec\(/`) matched those string-embedded calls and demanded `const TEST_FEATURE = ...` from a file with no real OpenSpec usage.

## Fix

1. Accept `createAllureTestHelpers(` as satisfying the helper-source rule, alongside the existing import check.
2. Strip string-literal contents (backticks, single, double) before applying the `.openspec(` detector so it only sees real code, not fixture source.

## Test plan

- [x] `node scripts/validate_allure_test_labels.mjs` passes locally
- [x] `vitest run` on both affected test files still passes
- [ ] CI green on this PR

### DIFF
--- a/scripts/validate_allure_test_labels.mjs
+++ b/scripts/validate_allure_test_labels.mjs
@@ -9,6 +9,10 @@ const TEST_PATH_RE = /^packages\/[^/]+\/(src|test)\/.+\.test\.ts$/;
 const ALLURE_EXEMPT_PACKAGES = new Set(['google-docs-core']);
 
 const helperImportRe = /from\s+['"][^'"]*allure-test\.(?:js|ts)['"]/;
+// The allure-test-factory package is the source of the shared helpers; its
+// self-tests construct the wrappers in-process via createAllureTestHelpers,
+// so they satisfy the "use the shared wrappers" rule without an import.
+const factoryConstructorRe = /\bcreateAllureTestHelpers\s*\(/;
 const wrapperReferenceRe = /\b(itAllure|testAllure)\b/;
 const epicWrapperAssignmentRe =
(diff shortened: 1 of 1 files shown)


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `135b4a269ab97ee800e27dd0fb141de527edd583`
Built from: `main` @ `21fc3940d198c3403f1cf04f72cbe1d0ea969f63` (one subsequent dependabot bump landed)
Smoke: targeted — re-ran the formerly-red CI check + both false-positive test files on freshly-pulled `main`.

**Steps**
- ✅ `npm run check:allure-labels` (the formerly-red CI invocation)
- ✅ `vitest packages/test-narrative/src/astExtractor.test.ts` (16 tests — file that had the false-positive `.openspec(` template-literal match)
- ✅ `vitest packages/allure-test-factory/src/visibility.test.ts` (3 tests — factory self-test that couldn't import `allure-test` from itself)

**Main CI**
- ✅ `135b4a2` (this merge) — first green CI on `main` after several red runs
- ✅ `21fc394` (subsequent dependabot bump) — still green

Log: `/tmp/automerge-smoke-267-1779836864.log` (local).

PR #270, opened 2026-05-27
fix(ci): npm 11 rejects double-loaded user/global config — split paths

stevenobiajulu:
## Summary

One-line fix to the `.npmrc`-poisoning hardening from #253. That hardening set both `NPM_CONFIG_USERCONFIG` and `NPM_CONFIG_GLOBALCONFIG` to the same path:

```yaml
NPM_CONFIG_USERCONFIG: ${{ runner.temp }}/llm-gate-empty-npmrc
NPM_CONFIG_GLOBALCONFIG: ${{ runner.temp }}/llm-gate-empty-npmrc
```

npm 11+ refuses to load one file in both roles and exits with:

```
Exit prior to config file resolving
cause
double-loading config "/home/runner/work/_temp/llm-gate-empty-npmrc" as "global", previously loaded as "user"
##[error]Process completed with exit code 1.
```

This caused every LLM gate matrix job on PR #268 and #269 (the first PRs to trigger the gate after #253 merged) to fail at the "Install Gemini CLI" step. Phase 1 advisory mode meant the gate's `Aggregate and post review` check is not in required status checks, so those PRs merged anyway — but the gate itself has been unable to produce a verdict for any PR since #253 landed.

## Fix

Use two separate empty-file paths so neither role is double-loaded:

```yaml
NPM_CONFIG_USERCONFIG: ${{ runner.temp }}/llm-gate-empty-userconfig
NPM_CONFIG_GLOBALCONFIG: ${{ runner.temp }}/llm-gate-empty-globalconfig
```

Touch both files before the install. The security property is unchanged — both files are still forced-empty, registry is still pinned, install still runs from `\$RUNNER_TEMP`. Same fix lands in parallel on [open-agreements](https://github.com/open-agreements/open-agreements) (which inherited the bug via the original port direction).

## Verification

- [x] `actionlint .github/workflows/llm-based-quality-gate.yml` clean
- [x] Embedded `settings.json` still parses
- [ ] Next PR after merge: gate should run cleanly through the install step and produce verdicts (this is the actual end-to-end test)

## Test plan

When this PR opens, the gate workflow on the PR will hit the same broken install (because workflows use base-ref source). Expected: same `double-loading` failure as #268/#269. **After merge**, future PRs should see the gate install Gemini cleanly and produce per-item verdicts.

PR #269, opened 2026-05-26
docs(comments): reclassify addCommentReply as Table B (side-part metadata only)

stevenobiajulu:
## Summary

Implements Option B from #174: align `SUPPORT.md` and the regression-test name with the actual contract of `addCommentReply`, which has always been a side-part operation (per #138) — it updates `comments.xml`, `commentsExtended.xml`, and `people.xml` without emitting body revision markup. The regression test in #143 asserted that behavior but was named with a `Table A:` prefix, locking the divergence into the test suite.

## Changes

- **SUPPORT.md** — split the existing lumped Table B row ("`addComment`, `addCommentReply`, `deleteComment`") so `addCommentReply` has its own row with explicit:
  - Side-part metadata writes only (`comments.xml` + `commentsExtended.xml` + `people.xml`)
  - No body anchor per reply
  - No body revision marker emitted
  - Cannot be selectively rejected by author-based accept/reject

- **canonical-emission-regression.test.ts** — rename the test from "`Table A: comments.ts addCommentReply currently leaves revision markup untouched and updates comments.xml only`" to "`Table B: comments.ts addCommentReply performs package mutation only with no body revision`". **Assertions unchanged** — the behavior was always correct, only the classification was wrong.

## Why Option B (vs Option A — emit body markup)

Comments don't have body anchors per-reply — only the root comment has a `<w:commentRangeStart>`/`<w:commentRangeEnd>` anchor. Emitting body markup for a reply would either invent an artificial anchor (fragile) or write to `commentsExtended.xml` which is a side-part metadata write that belongs in Table B anyway. Option B matches the implementation and the natural shape of the data.

## Verification

- [x] `npm run check:conformance-citations` passes (3 registry entries, 51 XSDs)
- [x] `npm run check:spec-coverage` passes
- [x] Renamed regression test still passes (assertions unchanged)
- [x] Diff is exactly 2 files: SUPPORT.md (+2/-1 lines) and the test file (1-line rename)

## Closes

- #174

## Related

- #120 (umbrella)
- #138 (#120.3 — implemented addCommentReply with ctx as plumbing only)
- #143 (#120.8 regression suite — flagged this divergence)

PR #274, opened 2026-05-27
fix(ci): harden LLM-gate against npm config-load failures (closes #273)

### LINKED ISSUE #273: LLM-Based Quality Gate matrix jobs fail with "Exit prior to config file resolving" before invoking the model
## Problem

The LLM-Based Quality Gate (`.github/workflows/llm-based-quality-gate.yml`) has been intermittently failing on recent PRs. Five of the last six runs ended in workflow-level FAILURE, and the PR comments on the affected runs say:

> **Overall:** ❓ NO RESULTS (0 pass · 0 warn · 14 total)
>
> _No results were produced. Check the workflow run for errors._

The gate is currently advisory (`LLM_GATE_BLOCKING` unset). If we promote it to enforcing without fixing this, the failure mode lets PRs merge silently — the required `Aggregate and post review` check still exits 0 (the blocking branch in the aggregate step fires on `warns > 0`, not on `total == 0`).

## Root cause sighting

Inspecting one failed matrix job — [job 77999406077](https://github.com/UseJunior/safe-docx/actions/runs/26487987643/job/77999406077) on PR with branch `fix/npm-config-double-load`, check 02 — the Gemini CLI dies before producing output:

```
2026-05-27T02:57:47.3009530Z Exit prior to config file resolving
2026-05-27T02:57:47.3144539Z ##[error]Process completed with exit code 1.
```

The error is emitted by `@google/gemini-cli` (pinned at `0.39.1` via the `GEMINI_CLI_VERSION` repo variable) during the `Run check with retry on malformed output` step in `.github/actions/llm-gate-check/action.yml`. The check-run step has retry logic for malformed model output, but it does NOT retry on this CLI-startup error — `gemini` exits 1 before writing any response, so the retry branch (which looks at the response JSON) never fires.

## Affected runs

- PR #270 — workflow run [26481977858](https://github.com/UseJunior/safe-docx/actions/runs/26481977858), comment posted: "NO RESULTS"
- PR #269 — workflow run [26481976773](https://github.com/UseJunior/safe-docx/actions/runs/26481976773), comment posted: "NO RESULTS"
- PR with branch `fix/npm-config-double-load` — workflow run [26487987643](https://github.com/UseJunior/safe-docx/actions/runs/26487987643)
- PR with branch `fix/allure-labels-validator-false-positives` — workflow run [26479483400](https://github.com/UseJunior/safe-docx/actions/runs/26479483400)

PR #271's run (workflow [26490355255](https://github.com/UseJunior/safe-docx/actions/runs/26490355255)) succeeded with 14/14 PASS. So the failure is intermittent, not deterministic.

## Hypotheses (in priority order)

1. **API quota exhaustion** — the `GEMINI_API_KEY` secret is on the same Google Cloud project (`projects/522706245871`) that now also serves `UseJunior/tests-renderer`'s LLM gate (added 2026-05-27). If safe-docx PR bursts exceed the project's Gemini free-tier quota, individual matrix jobs would fail with a startup-time auth/quota error and "Exit prior to config file resolving" is one of the messages Gemini CLI emits on auth/quota failure.
2. **Composite-action npm-cache regression** — the install step uses forced-empty userconfig/globalconfig + pinned registry to defend against malicious `.npmrc` from PR-controlled content. A subtle change in `@google/gemini-cli`'s install-time behavior under that config could leave a partial install where the binary exists but config-resolving paths fail.
3. **`@google/gemini-cli@0.39.1` upstream behavior change** — the pinned version is two months old; the CLI itself may emit "Exit prior to config file resolving" for any of several edge cases that the published version handles differently than the live API expects.
4. **Settings.json being written to `.gemini/` with a structure incompatible with the runtime CLI** — `Build settings.json (tool allowlist)` writes nested `tools.core` + `model.maxSessionTurns`. If the CLI moved or renamed these keys, settings-file resolution would fail early.

## Reproduction

Re-run [job 77999406077](https://github.com/UseJunior/safe-docx/actions/runs/26487987643/job/77999406077) to confirm the failure is sticky. If it now passes, hypothesis 1 (quota) is most likely. If it still fails, focus on hypotheses 2-4.

## Acceptance

- Identify the failure-mode trigger (which of the four hypotheses, or a different one).
- Either fix the trigger (e.g., raise quota, bump CLI version, repair settings format) or harden the composite action to retry on CLI-startup failure (currently retry only fires on malformed model output).
- Re-run the gate on the ~10 most recent PRs (via `workflow_dispatch`) and confirm every run produces actual verdicts (no "NO RESULTS" comments).
- After ~10 consecutive clean runs, promote: `gh variable set LLM_GATE_BLOCKING --body 1` and add `Aggregate and post review` to the required branch-protection contexts on \`main\`.

## Why this matters now

This issue is blocking promotion of the LLM gate from advisory to enforcing on `main`. The same workflow was just shipped to [UseJunior/tests-renderer](https://github.com/UseJunior/tests-renderer/pull/13) (where it's already enforcing with the same `Aggregate and post review` required check) — if the failure mode reproduces there, the renderer-repo PRs would also hit the silent-pass hole.

stevenobiajulu:
## Summary

Three-layer fix for the intermittent `Exit prior to config file resolving` failures in the LLM-Based Quality Gate, plus one defense-in-depth change closing a silent-pass hole that would have surfaced when promoting the gate to enforcing.

## Implementation

Root cause is an npm config-loader failure, not a Gemini API issue. Reproduced locally:

```
Exit prior to config file resolving
cause
double-loading config "/tmp/npmrc.thHOfb" as "global", previously loaded as "user"
```

That string is from npm 11's exit-handler ([source](https://raw.githubusercontent.com/npm/cli/v11.12.1/lib/cli/exit-handler.js)). The Gemini CLI shells out to npm at startup for its auto-update check; the install-time defensive `NPM_CONFIG_USERCONFIG` / `NPM_CONFIG_GLOBALCONFIG` env vars leak into the runtime and trip the double-load assertion on the next npm invocation.

The fix has three layers in `.github/actions/llm-gate-check/action.yml`:

1. **Isolate the install** — `npm install --prefix "$RUNNER_TEMP/gemini-cli"` instead of `--global`, with the bin dir added to `GITHUB_PATH`. Decouples CLI from npm-global config state.
2. **Disable Gemini's startup-time npm calls** — set `general.enableAutoUpdate=false` + `general.enableAutoUpdateNotification=false` in `.gemini/settings.json` per [Gemini CLI config docs](https://raw.githubusercontent.com/google-gemini/gemini-cli/v0.39.1/docs/reference/configuration.md). With auto-update off, the CLI doesn't shell out to npm before serving the model request.
3. **Defense in depth** — `unset` all inherited `npm_config_*` / `NPM_CONFIG_*` env vars before invoking `gemini`. If any future code path still shells out to npm, it loads npm's defaults instead of the install-time defensive config.

Additional hardening on a separate (but related) risk:

4. **Retry on CLI startup failure** — the retry loop previously only fired on malformed model JSON; CLI-startup failures failed the matrix job on attempt 1. Now both classes retry up to 3 times. The retry diagnostics distinguish "CLI startup failure before response JSON" from "CLI failure with unparseable response JSON" from "malformed output" for easier root-causing.
5. **Fail closed on zero results in blocking mode** — `.github/workflows/llm-based-quality-gate.yml` aggregate step now calls `core.setFailed` when `LLM_GATE_BLOCKING=1` and no result rows were produced. Closes the silent-pass hole — previously a "NO RESULTS" run with `LLM_GATE_BLOCKING=1` still let the required check pass.

## Verification

- [x] `actionlint .github/workflows/llm-based-quality-gate.yml`: pass
- [x] YAML parse on both changed files: pass
- [x] Local repro of the original `Exit prior to config file resolving` error confirmed the npm config-double-load trigger
- [ ] Peer review (Gemini + Codex) — pending; will be appended below
- [ ] CI on this PR runs the gate against the new code (the workflow checks out the PR head for the composite action; if the fix works, this PR's gate run produces 14 verdicts instead of "NO RESULTS")

## Follow-up after merge

Once this lands, re-run the gate on the ~10 most recent PRs via `workflow_dispatch` and confirm every run produces actual verdicts. Once ~10 consecutive clean runs, promote:

```bash
gh variable set LLM_GATE_BLOCKING --body 1 --repo UseJunior/safe-docx
# Then add "Aggregate and post review" to required branch-protection contexts
```

## Closes

- #273

### DIFF
--- a/.github/actions/llm-gate-check/action.yml
+++ b/.github/actions/llm-gate-check/action.yml
@@ -60,8 +60,11 @@ runs:
         set -euo pipefail
         : > "$NPM_CONFIG_USERCONFIG"
         : > "$NPM_CONFIG_GLOBALCONFIG"
+        GEMINI_PREFIX="$RUNNER_TEMP/gemini-cli"
         cd "$RUNNER_TEMP"
-        npm install --silent --no-audit --no-fund --global "@google/gemini-cli@${GEMINI_CLI_VERSION}"
+        npm install --silent --no-audit --no-fund --global "@google/gemini-cli@${GEMINI_CLI_VERSION}" --prefix "$GEMINI_PREFIX"
+        echo "$GEMINI_PREFIX/bin" >> "$GITHUB_PATH"
+        export PATH="$GEMINI_PREFIX/bin:$PATH"
         gemini --version
 
     - name: Build settings.json (tool allowlist)
@@ -74,6 +77,9 @@ runs:
         # Top-level `coreTools` / `maxSessionTurns` is silently ignored, which
         # would leave the gate running with default (broader) tool access.
         # See: https://github.com/google-gemini/gemini-cli/blob/v0.39.1/docs/reference/configuration.md
+        # Disable auto-update in CI: the update path shells out to npm before
+        # the model runs, and npm config-load failures look like Gemini startup
+        # failures. The CLI docs define these under `general.*`.
         #
         # The allowlist is intentionally broad so the model can deterministically
         # verify lint/spec/check claims via `npm run *` rather than only
@@ -89,6 +95,10 @@ runs:
         #     by the output sanitizer below.
         cat > .gemini/settings.json <<'JSON'
         {
+          "general": {
+            "enableAutoUpdate": false,
+            "enableAutoUpdateNotification": false
+          },
           "model": {
             "maxSessionTurns": 50
           },
@@ -154,6 +164,16 @@ runs:
         PROMPT_FILE="$RUNNER_TEMP/prompt-${CHECK_ID}.txt"
         compose_prompt "$PROMPT_FILE"
 
+        # The Gemini CLI can invoke npm during startup for package-management
+        # paths such as update checks. Keep the install step's defensive npm
+        # config from leaking into runtime, where npm 11 can abort with
+        # "Exit prior to config file resolving" before Gemini writes JSON.
+        while IFS='=' read -r name _; do
+          case "$name" in
+            npm_config_*|NPM_CONFIG_*) unset "$name" ;;
+          esac
+        done < <(env)
+
         PARSED=""
         ATTEMPTS=0
         MAX_ATTEMPTS=3
@@ -183,6 +203,7 @@ runs:
             GEMINI_EXIT=$?
           fi
 
+          RETRY_REASON="malformed output"
           if [ "$GEMINI_EXIT" -eq 0 ]; then
             # The CLI returns {"response": "<model text>", "stats": {...}}.
             # Extract response, then parse it as our {status, justification} JSON.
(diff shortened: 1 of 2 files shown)


stevenobiajulu:
## Peer review (Codex + Gemini) — 2026-05-27

**Codex findings**: APPROVED. Read both files directly, ran `actionlint` (exit 0), confirmed the `--prefix` install isolation puts the bin dir on `GITHUB_PATH`, verified the `npm_config_*` unset loop runs in the current shell (process substitution, not a subshell — Codex re-tested locally), confirmed `general.enableAutoUpdate` + `general.enableAutoUpdateNotification` are valid keys in the [Gemini CLI v0.39.1 config docs](https://raw.githubusercontent.com/google-gemini/gemini-cli/v0.39.1/docs/reference/configuration.md), verified the retry loop honors `MAX_ATTEMPTS=3` for both failure classes, and confirmed the fail-closed-on-zero-results branch is correctly placed before the `warns > 0` branch (no early-return needed; `warns > 0` is naturally false when `rows.length === 0`). One residual risk noted: Codex did not reproduce the original intermittent npm failure with an actual failing run; mechanics verified, historical root-cause inference accepted on local repro evidence.

**Gemini findings**: NEEDS-EXECUTION → unavailable. The Gemini CLI hit a 429 rate-limit (`gaxios status 429`) before producing any review output. Per the peer-review skill's anti-simulation rule, treat as unavailable rather than as approval.

**Resolution**: Codex's APPROVED verdict + the empirical evidence from this PR's own CI (the LLM-gate matrix items 01–06 are SUCCESS on this branch where the same checks were failing with "Exit prior to config file resolving" on prior PRs) are sufficient. Arming auto-squash-merge.

stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `7a538d6a5767eb83eb3e916b4827c6b27fc0a35d`
Built from: `main` @ `7a538d6a5767eb83eb3e916b4827c6b27fc0a35d`
Smoke: `npm run build && npm run test:run && workflow_dispatch on llm-based-quality-gate.yml`

**Steps**
- ✅ build — all 6 workspaces compiled clean (docx-core, docx-mcp, allure-test-factory, test-narrative, safedocx-mcpb, google-docs-core)
- ✅ full test suite — **2191 tests** across 6 workspaces (1284 docx-core, 750 docx-mcp, 116 allure-test-factory, 35 test-narrative, 3 safedocx-mcpb, 3 google-docs-core)
- ✅ **feature entry point — this PR's own CI run is the real-world smoke**. The LLM-gate ran the fixed code on this PR and produced `Overall: ✅ PASS (14 pass · 0 warn · 14 total)` with grounded per-item justifications. Compare to recent prior PRs (#270, #269) where the same gate posted `NO RESULTS (0 pass · 0 warn · 0 total)`. Concrete confirmation the npm-config-double-load failure mode is gone.

**Follow-up nit** (out-of-scope for this PR): I tried to additionally re-dispatch the gate against PR #270 via `gh workflow run llm-based-quality-gate.yml -f pr_number=270` to confirm the fix on previously-failing content. That run errored at `Resolve PR context` with `fatal: not a git repository` — a pre-existing bug in the `workflow_dispatch` path (the step calls `gh pr view` before any checkout). Not introduced by this PR; would need its own small fix (add `--repo "$GITHUB_REPOSITORY"` to the `gh pr view` call) to make the post-incident re-run loop work. The normal `pull_request` event path is fine; that's what produced the PASS comment above.

Log: `/tmp/automerge-smoke-274-1779893684.log` (local).

PR #310, opened 2026-06-04
feat(docx-mcp): DOCX → Markdown export via shared serializer core

stevenobiajulu:
## What

Adds the **export** leg of the read → edit → compare → export loop (Fixes: #303, epic #307). safe-docx can now emit a portable Markdown rendering of a DOCX.

Per the epic, Markdown/HTML/text are *thin emitters over a shared structured-export core*. This PR builds the Markdown emitter and factors the **inline-tag tokenizer** as the reusable core (HTML #304 will render the same tokens). It is a serializer over the existing `buildDocumentView()` model — **no new OOXML parsing**.

### docx-core
- `serialize_markdown.ts` — `inlineTagsToMarkdown()` (reusable tokenizer → Markdown) and `serializeToMarkdown()` (headings → ATX, nested lists, GFM tables, footnote definitions).
- `document_view.ts` — export `TOON_INLINE_TAG_RE` + new `tokenizeToonInline()`; **make `injectFootnoteMarkers` tag-aware** (reuse `findTaggedTextInsertionIndex`, as the comment-marker path already did). Previously a visible-offset `[^n]` marker was spliced into the *tagged* string by raw index, so it could land inside a formatting tag once `show_formatting` was on. This also corrects `read_file` output for formatted footnoted text.
- `document.ts` — async `DocxDocument.toMarkdown()`.

### docx-mcp
- `export` tool — `format` enum (`markdown`), writes an output file (default: source path with `.md`), overwrite guard + `enforceWritePathPolicy`, returns `output_path` + `bytes_written` + `markdown` (suppressible via `include_markdown:false`). DOCX only. Registered in `tool_catalog.ts` + `server.ts`.

## Design notes (lossy by intent)
- No round-trip guarantee. Highlighting and font runs drop their tags (inner text kept); `<u>` passes through as raw GFM HTML.
- GFM has no colspan/rowspan: horizontally-merged cells leave grid gaps that are filled with empty cells; vertically-merged/nested cells are flattened; multi-line cells use `<br>`.
- **Legal list labels** (`Section 2.1`, `Article IV`, `(a)`) are preserved rather than flattened to a bare `1.`.
- Only Word-styled headings (numeric level) become `#`; heuristic run-in headings stay paragraphs (their bold already lives in the inline tags).
- Tokenize → escape only text spans (footnote markers protected) → map tags, so escaping never corrupts generated Markdown or `[^n]` markers. Adjacent run-split emphasis toggles are coalesced (no stray `******`).

## OpenSpec
[...]

PR #320, opened 2026-06-04
fix(docx-mcp): resolve final write-path component to block symlink escape (#313)

### LINKED ISSUE #313: enforceWritePathPolicy follows a symlink final component → cross-root write escape (save / compare_documents / gdocs/save / export)
## Summary

`enforceWritePathPolicy` (`packages/docx-mcp/src/tools/path_policy.ts:136`) canonicalizes only the **existing ancestor directory** of the output path, not the final path component. If the output path *itself* is an existing symlink, the policy check passes against the link's own (in-root) location, but the subsequent `fs.writeFile` **follows the symlink** and writes to wherever it points — potentially **outside the allowed roots**.

This is the write-side analogue of a check the read path already gets right: `enforceReadPathPolicy` (`path_policy.ts:114`) calls `fs.realpath(normalizedPath)` on the **full** path, so a symlinked read target is correctly resolved and policy-checked. `enforceWritePathPolicy` instead calls `resolveWritePathWithExistingAncestor` (`path_policy.ts:97`), which realpaths `path.dirname(...)` and rejoins the un-resolved tail — so a symlink *file* at the output path slips through.

## Affected callers (all share the helper)

- `packages/docx-mcp/src/tools/save.ts:288` (clean) and `:294` (tracked)
- `packages/docx-mcp/src/tools/compare_documents.ts:114`
- `packages/docx-mcp/src/tools/gdocs/save.ts:35`
- `packages/docx-mcp/src/tools/export.ts:111`

## Impact

1. **Cross-root write escape.** A symlink at an in-root output path (`~/docs/out.docx -> /etc/something`) with `allow_overwrite` lets a write land outside `SAFE_DOCX_ALLOWED_ROOTS`.
2. **Source-DOCX clobber via symlink** was the concrete data-loss case originally surfaced by peer review on #310. That one is now mitigated **for `export` only** — `export.ts` canonicalizes both sides with `fs.realpath` before its source-clobber comparison (commit `69b5400`, shipped in #310). `save` / `compare_documents` / `gdocs/save` have **no** equivalent guard and still rely solely on the shared policy.

Pre-existing — not introduced by #310. Filed as the agreed follow-up from that PR's review (peer reviewer: Codex, which reproduced the escape with a live symlink write; Gemini approved it statically as "robust", incorrectly).

## Suggested fix

Make `enforceWritePathPolicy` resolve the **final component** when it already exists:

- If the normalized output path exists, `fs.realpath` the whole path (as the read policy does) and policy-check *that* canonical target — so a symlink output is judged by where it actually writes.
- If it does not exist, keep the current existing-ancestor resolution (you can't realpath a not-yet-created file).
- Alternatively / additionally, open with `O_NOFOLLOW`-style no-follow semantics on the final component so a symlinked output is rejected outright rather than followed.

Then fold the `export`-local realpath source-clobber guard back into the shared helper (or a shared `assertNotSource` helper) so all four write tools get the same protection and the export-specific guard isn't a one-off.

## Tests to add

- Symlink output inside an allowed root pointing **outside** all roots → write is refused (assert target file untouched). Cover `save`, `compare_documents`, `gdocs/save` (export already has a regression test: `export.test.ts` "a symlink output_path pointing back at the source is refused").
- Symlink output pointing at the **source** document → refused for every write tool.
- Regression: a plain (non-symlink) output path in an allowed root still succeeds.

stevenobiajulu:
Closes #313.

## Problem

`enforceWritePathPolicy` canonicalized only the **existing ancestor directory** of an output path, never the final component. An in-root symlink output therefore passed the policy check at the link's location while `fs.writeFile` followed it — writing **outside `SAFE_DOCX_ALLOWED_ROOTS`**, or **clobbering a source document** through the link. This is the write-side analogue of a check `enforceReadPathPolicy` already gets right (it `realpath`s the full path).

Pre-existing; agreed follow-up from #310's review.

## Fix

**Core (`path_policy.ts`)** — `resolveWritePathWithExistingAncestor` → `resolveWritePathCanonical`, which resolves the final component:
1. realpath the full path when it resolves (symmetric with the read policy);
2. **follow a dangling final symlink** via `lstat`/`readlink` (with cycle protection) — a naive realpath-only fix misses this: a dangling link makes `realpath` throw, falling back to ancestor resolution, so the link is judged in-root while the write still creates the target outside;
3. fall back to existing-ancestor resolution for genuinely new files (unchanged).

**Source-clobber hardening** — folds the export-local realpath guard into a shared `resolvesToSamePath` helper and applies it to `save` (replacing a symlink-bypassable lexical comparison; `allow_overwrite` gating preserved) and `compare_documents` (which had no guard, placed before the expensive comparison). Every write tool now refuses to overwrite an input document, including through a symlink. `gdocs/save` has no local source, so it gets the cross-root fix for free.

## Tests (+13)

Existing + dangling symlink escapes and source-clobber-via-symlink across `path_policy`, `save`, `compare_documents`, and `gdocs` save, plus a symlink-cycle case. All symlink tests early-return on `win32`.

## Verification

- `tsc -b` clean; `check:spec-coverage` pass
- Full `docx-mcp` suite: **775 passed, 0 regressions**

## Peer review

Reviewed via Codex (dynamic, live Node repros) and Gemini (static). Codex found the dangling-symlink gap — now fixed and regression-covered. Gemini's two nits (null-check `trackedPath`, rename) incorporated.

### DIFF
--- a/packages/docx-mcp/src/tools/compare_documents.test.ts
+++ b/packages/docx-mcp/src/tools/compare_documents.test.ts
@@ -213,6 +213,72 @@ describe('compare_documents tool', () => {
     },
   );
 
+  // ── Path policy (issue #313) ─────────────────────────────────────
+
+  test(
+    'refuses a symlink save_to_local_path that escapes the allowed roots',
+    async ({ when, then }: AllureBddContext) => {
+      if (process.platform === 'win32') return;
+
+      const mgr = createTestSessionManager();
+      const dir = await createTrackedTempDir();
+      const outsideDir = await createTrackedTempDir();
+      const originalPath = await writeTestDocx(dir, 'original.docx', ['Hello world']);
+      const revisedPath = await writeTestDocx(dir, 'revised.docx', ['Hello brave new world']);
+
+      const previousRoots = process.env.SAFE_DOCX_ALLOWED_ROOTS;
+      process.env.SAFE_DOCX_ALLOWED_ROOTS = dir;
+      try {
+        const escapeTarget = path.join(outsideDir, 'redline.docx');
+        const link = path.join(dir, 'redline-link.docx');
+        await fs.symlink(escapeTarget, link);
+
+        const result = await when('compare with an escaping symlink output', () =>
+          compareDocuments_tool(mgr, {
+            original_file_path: originalPath,
+            revised_file_path: revisedPath,
+            save_to_local_path: link,
+          }),
+        );
+        await then('the write is refused and the outside target is never created', async () => {
+          assertFailure(result, 'PATH_NOT_ALLOWED', 'compare symlink escape');
+          await expect(fs.access(escapeTarget)).rejects.toThrow();
+        });
+      } finally {
+        if (previousRoots === undefined) delete process.env.SAFE_DOCX_ALLOWED_ROOTS;
+        else process.env.SAFE_DOCX_ALLOWED_ROOTS = previousRoots;
+      }
+    },
+  );
+
+  test(
+    'refuses a save_to_local_path that resolves onto a source document (incl. via symlink)',
+    async ({ when, then }: AllureBddContext) => {
+      if (process.platform === 'win32') return;
+
+      const mgr = createTestSessionManager();
+      const dir = await createTrackedTempDir();
+      const originalPath = await writeTestDocx(dir, 'original.docx', ['Hello world']);
+      const revisedPath = await writeTestDocx(dir, 'revised.docx', ['Hello brave new world']);
+      const originalBytes = await fs.readFile(originalPath);
+
+      const link = path.join(dir, 'link-to-original.docx');
+      await fs.symlink(originalPath, link);
+
+      const result = await when('compare with output symlinked to a source', () =>
+        compareDocuments_tool(mgr, {
+          original_file_path: originalPath,
+          revised_file_path: revisedPath,
+          save_to_local_path: link,
+        }),
+      );
+      await then('the write is refused and the source is byte-for-byte intact', async () => {
+        assertFailure(result, 'OVERWRITE_BLOCKED', 'compare clobbers source');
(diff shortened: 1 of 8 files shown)


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `df24cfb1871e37096c9b41921ebc1d2f5dbd2558` (squash)
Built from: merge commit `df24cfb` in a clean detached worktree — the primary `main` worktree has unrelated pre-existing uncommitted changes (`packages/allure-test-factory/*`) that blocked `git pull --ff-only`, so the smoke ran against a fresh checkout of the merged SHA instead of mutating your working tree.
Smoke: `node smoke.mjs` (built `dist` driven via `openDocument`/`save`) + full `vitest run`

**Steps**
- ✅ build (clean `npm run build` from merged HEAD)
- ✅ tests (775 passed, 75 files)
- ✅ end-to-end policy entry points against a real DOCX

**Real-world fixtures**
- `bonterms-mutual-nda.docx` (real 29.5 KB open-agreements template at `tests/test_documents/open-agreements/`), opened + saved through the built MCP tools:
  - baseline: opens cleanly, clean-save round-trips (195 KB written) within the allowed root
  - existing symlink output → outside the allowed root ⇒ `PATH_NOT_ALLOWED`; external target byte-for-byte intact
  - **dangling** symlink output → outside ⇒ `PATH_NOT_ALLOWED`; target never created (the case the naive realpath-only fix missed)
  - symlink output → the source document ⇒ `OVERWRITE_BLOCKED`; source byte-for-byte intact

All 8 smoke assertions passed; no escape, no clobber.

**Cleanup**
- ✅ remote branch `313-fix-write-path-symlink-final-component-20260604` already deleted (repo auto-deletes head branches)
- ✅ removed worktrees `.worktrees/fix-313` and the temp smoke worktree
- ✅ deleted local branch

Log: `/tmp/automerge-smoke-320-1780604990751.log` (local).

PR #326, opened 2026-06-05
ci(release): add ODF release-isolation guard and OpenSpec proposal

stevenobiajulu:
## Summary

ODF packages (`@usejunior/odf-core` + an `odf` MCP backend) are coming and will churn at `0.x` while the DOCX suite stays at `0.9.1`. `release.yml` pins versions and publishes via **four hardcoded DOCX package lists**; if ODF were added to them, every ODF `0.x` bump would force a version bump + re-publish of the stable DOCX suite. **This PR is the blocker-clearing decision before any ODF code lands** (Workstream B, Phase 0).

It does **not** build the independent ODF release pipeline (that's deferred to publish-time, when `odf-core`'s real shape is known). Instead it makes the coupling impossible to reintroduce by accident, with the minimum footprint.

## What's here

- **`scripts/check-release-isolation.mjs`** — wired into the `workspace-lint` required check. Two assertions:
  - **A. Workflow-snapshot** — `release.yml`'s four DOCX package loops (version-pin, duplicate-publish, dry-run pack, publish) must equal a fixed expected set. Adding **any** package (ODF or otherwise) to a release list fails CI. This guards the *actual* coupling mechanism (membership in a release list), so it's robust to packages that are non-private yet on no release list (e.g. `allure-test-factory`).
  - **B. ODF-private** — every workspace package whose name matches `/odf/` must be `private: true`, and stays private until `release-odf.yml` exists and passes its own preflight.
- **`scripts/check-release-isolation.test.mjs`** — unit tests for the guard logic (9 tests, `node:test`).
- **OpenSpec change `add-odf-release-isolation`** — proposal / design (three options weighed) / tasks / `release-publishing` capability spec.

## What's deliberately NOT here

- **No `release.yml` edits, no version bumps, nothing published.**
- The future independent ODF release track will use a non-colliding `odf-v*` git-tag prefix (our own repo's tags — not an external name reservation).
- **No name-squatting** — no placeholder `odf-mcp` publish to reserve the npm name; ODF publishes only when it's a real, functioning package.

## Verification

- [x] Build clean, `lint:workspaces`, `check:spec-coverage`, `check:conformance-citations`, `check:conformance-doc` — all green locally
- [x] Guard passes on the current tree; 9/9 unit tests pass
- [x] Negative tests: an `odf`-named non-private fixture fails assertion B; adding a package to a `release.yml` DOCX list fails assertion A
- [x] `openspec validate add-odf-release-isolation --strict` passes
- [x] `release.yml` unchanged; no DOCX package version touched

## Review notes

Peer-reviewed with Codex (dynamic execution). Four findings incorporated — most notably reframing the guard off an "allowlist-or-private" model that **fails on the current tree** (`allure-test-factory` is non-private but on no release list), and dropping a bogus `npm query .workspace` enumeration (returns `[]` on npm 11.12) in favor of expanding the root `workspaces` glob.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,6 +71,10 @@ jobs:
         run: npm run lint:workspaces
       - name: Check for import cycles
         run: npm run check:cycles
+      - name: Check release isolation (ODF decoupling guard)
+        run: npm run check:release-isolation
+      - name: Test release-isolation guard
+        run: npm run test:release-isolation
 
   spec-coverage:
     runs-on: ubuntu-latest
--- a/openspec/changes/add-odf-release-isolation/design.md
+++ b/openspec/changes/add-odf-release-isolation/design.md
@@ -1,0 +1,176 @@
+# Design: Isolate ODF package releases from the DOCX suite
+
+## Context
+
+`release.yml` is suite-versioned. The preflight job's
+"Verify tag matches package suite versions" step iterates a hardcoded list:
+
+```
+packages/docx-core packages/docx-mcp packages/google-docs-core \
+packages/safe-docx packages/safe-docx-mcpb
+```
+
+and additionally pins `safe-docx-mcpb/manifest.json`, `safe-docx/server.json`,
+`safe-docx/.smithery/shttp/manifest.json`, and `gemini-extension.json` to the
+tag version (all currently `0.9.1`). The `publish-suite` job publishes a
+hardcoded four-package list (`docx-core`, `docx-mcp`, `google-docs-core`,
+`safe-docx`). The workflow triggers on `push: tags: ["v*.*.*"]` and a
+`workflow_dispatch` taking an existing `v*.*.*` tag.
+
+Two facts, both verified against the current workflow, shape this design:
+
+1. **`odf-v*.*.*` tags do not match the `v*.*.*` push trigger.** GitHub Actions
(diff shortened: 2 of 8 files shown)


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `4e8da2f0c95d0884994ec406894b2d1a8a5356ec`
Built from: `main` @ `4e8da2f`
Smoke: `npm ci && npm run build && npm run check:release-isolation && npm run test:release-isolation && npm run lint:workspaces`

**Steps**
- ✅ npm ci (clean install)
- ✅ build (all workspaces)
- ✅ `check:release-isolation` — *"Release-isolation guard passed: DOCX release lists match snapshot; no non-private ODF package."* (the feature this PR ships, run against fresh main)
- ✅ `test:release-isolation` — 9/9 guard unit tests pass
- ✅ lint:workspaces (0 errors, 3 pre-existing warnings)

Non-document PR (CI guard + scripts + OpenSpec), so the smoke exercises the guard itself end-to-end rather than a document fixture.

Log: `/tmp/automerge-smoke-326-1780646654.log` (local).

**Cleanup**
- ✅ remote branch `feat/odf-release-isolation` already deleted (auto-delete on merge)
- ✅ removed worktree `/Users/stevenobiajulu/Projects/safe-docx-odf-release`
- ✅ deleted local branch `feat/odf-release-isolation`

PR #329, opened 2026-06-05
chore(openspec): archive add-ecma-376-conformance-framework change

stevenobiajulu:
## What

Archives the `add-ecma-376-conformance-framework` OpenSpec change — the framework is already deployed on `main` — and lands the spec-traceability proof-of-concept tooling.

**Archive operation** (standard OpenSpec workflow):
- `D` `openspec/changes/add-ecma-376-conformance-framework/` → moved to `openspec/changes/archive/2026-06-03-add-ecma-376-conformance-framework/` (tasks all checked off).
- `A` `openspec/specs/spec-compliance/spec.md` — the change's spec delta applied as a baseline deployed spec.

**Spec-traceability PoC** (`scripts/spec-traceability/`):
- `build-wiki.mjs` — assembles the `[MS-OE376]` section wiki.
- `extract-element-definition.mjs` — pulls element definitions from the spec.
- `README.md`, `legal-scope.draft.json`.
- `.gitignore`: ignore `out/` and `scripts/spec-traceability/.cache/` (local PDF cache + generated output).

## Why

The conformance deliverables (`.conformance()` Allure label, `check_conformance_citations.mjs`, `generate_conformance_doc.mjs`, the `spec-compliance/` tree) all already landed on `main`. The active change record was stale (showing 8/35) and the archive copy reflects the completed/deployed state. Archiving keeps `openspec list` clean and records the capability as a baseline spec.

## Verification

- `openspec validate --specs --strict` → 6 passed (incl. new `spec/spec-compliance`)
- `openspec validate --changes --strict` → 17 passed; ecma no longer an active change
- `npm run check:spec-coverage` → PASS
- `npm run check:conformance-citations` → OK (3 registry entries, 51 XSDs)
- `npm run check:conformance-doc` → OK (no drift)

Excludes the stale root `package.json`/`package-lock.json` baseline edits (they predate main's `madge`/`check:cycles` additions); the PoC scripts resolve `@xmldom/xmldom`/`jszip` via workspace hoisting.

Ref: #227

Issue #227, opened 2026-05-25
Structured ECMA-376 traceability: bind code-comment spec claims to normative sources


stevenobiajulu:
## Context

safe-docx makes ECMA-376 conformance claims today as prose code comments that
mix two unequal authorities: the ECMA-376 specification, and internal GitHub
issue numbers. There is no structured way to:

- audit which claims in the code are grounded in the normative spec,
- state which subset of ECMA-376 the project targets and which subset is
  intentionally out of scope,
- measure coverage of the targeted subset,
- or — for AI assistants and external reviewers — resolve a citation like
  "ECMA-376 Part 4 § 17.16.5" to the actual normative text without leaving
  the repo.

This issue **memorializes the problem and outlines categories of solution**.
A specific design will be proposed in a follow-up (likely an OpenSpec change
proposal).

## A concrete example

`packages/docx-core/src/baselines/atomizer/inPlaceModifier.ts:949-955` and
`:1010-1012`:

```ts
// ECMA-376 Part 4 fragmentation (issue #217): for a collapsed-field atom,
// emit w:fldChar runs at sibling level (unwrapped) and wrap only the
// payload runs (w:instrText, w:t, etc.) in <w:del>. w:fldChar inside
// <w:del> is non-conformant and Word treats it as fatal. Iterates the
// constituent collapsedFieldAtoms (not the deduped source runs) so a
// mixed-run field — where multiple field elements share one source `<w:r>` —
// is correctly split into one cloned run per field element.
```

```ts
* Per ECMA-376 Part 4 § 17.16.5: `w:delInstrText` MUST appear inside `<w:del>`;
* by extension and by issue #217's deep-research conclusion, `w:fldChar` runs
* stay at sibling level.
```

The same pattern appears in `packages/docx-core/src/baselines/atomizer/pipeline.ts:361-372`:

```ts
/**
 * Targeted check for the ECMA-376 Part 4 conformance rule that's the focus of
 * issue #217: `w:fldChar` MUST NOT appear inside any `<w:del>` element. Word
 * treats this violation as fatal — the field state machine is discarded and
 * the field renders as literal-text fallback.
 ...
```

These comments cite ECMA-376 *and* issue #217 as co-authorities. The
issue-number reference is path-dependent: an external reader cannot resolve
`#217` to a normative source, and the project's own conformance claims become
coupled to internal history rather than to the standard. For a
standards-aligned codebase, internal issue numbers are useful provenance, but
[...]

PR #332, opened 2026-06-05
chore(deps): declare @xmldom/xmldom + jszip as explicit root devDeps

stevenobiajulu:
## What

Adds `@xmldom/xmldom` and `jszip` to the **root** `devDependencies`.

## Why

The spec-traceability PoC scripts landed in #329 (`scripts/spec-traceability/build-wiki.mjs`, `extract-element-definition.mjs`) `import` both packages but resolved them only via workspace hoisting from `@usejunior/docx-core`. A future docx-core dependency change could remove that hoist and silently break the root scripts. Declaring them at the root makes the dependency explicit and self-owned.

## Scope

Both packages already exist in the dependency tree, so the lockfile delta is just the two added **root** references — no new package installs:

```
       "@vitest/runner": "4.1.8",
+      "@xmldom/xmldom": "^0.9.10",
       ...
       "fast-xml-parser": "^5.7.0",
+      "jszip": "^3.10.1",
```

## Verification

- `npm install` → clean; lockfile diff is the two references only.
- `@xmldom/xmldom` + `jszip` resolve from repo root (`DOMParser`/`JSZip` import OK).
- Both scripts pass `node --check`.

Ref: #227

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,10 +15,12 @@
         "@typescript-eslint/eslint-plugin": "^8.56.0",
         "@typescript-eslint/parser": "^8.56.0",
         "@vitest/runner": "4.1.8",
+        "@xmldom/xmldom": "^0.9.10",
         "ajv": "^8.18.0",
         "allure": "3.8.2",
         "eslint": "^10.0.1",
         "fast-xml-parser": "^5.7.0",
+        "jszip": "^3.10.1",
         "madge": "8.0.0",
         "prismjs": "^1.30.0"
       },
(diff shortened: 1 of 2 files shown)


stevenobiajulu:
## ✅ Merged & verified

Squashed: `7ddd0f0` on `main`.

**Post-merge verification** (deps/chore PR — not document-shaped, no DOCX smoke):
- `npm ci` resolves cleanly on `main`
- `@xmldom/xmldom ^0.9.10` + `jszip ^3.10.1` present in root `devDependencies`
- `scripts/spec-traceability/{build-wiki,extract-element-definition}.mjs` pass `node --check`

**Cleanup**
- ✅ removed worktree `.worktrees/declare-spec-traceability-deps`
- ✅ deleted local branch + pruned refs
- ✅ remote branch already auto-deleted

Issue #356, opened 2026-06-10
ODF compare_documents: intra-paragraph (run-level) tracked changes for .odt


stevenobiajulu:
## Background

`compare_documents` now supports ODF (`.odt`) in **two-file mode at paragraph granularity** (landed in #348, OpenSpec change `add-odf-compare`). The engine lives in `packages/odf-core/src/compare/` — `diff.ts` (a pure paragraph-level LCS), `emit.ts` (the ODF tracked-changes emitter), and `index.ts` (`compareOdf(originalContentXml, revisedContentXml)`). It diffs the two documents at the **whole-paragraph** level: each body paragraph is reduced to its visible text, an LCS aligns the two sequences, and paragraphs are emitted as whole-paragraph inserts or deletes.

## Problem

Because the diff is whole-paragraph, a paragraph whose text changed at all is represented as a **delete of the entire old paragraph plus an insert of the entire new paragraph**. A one-word edit inside a long clause shows the whole clause struck through and the whole clause re-inserted. For real review work (especially legal prose, where paragraphs are long), this is coarse and noisy: a reviewer wants to see only the words that changed marked up, not the entire paragraph churned. This is also why `stats.modifications` is always `0` today — there is no notion of an in-place modification, only delete + insert.

The DOCX path does this correctly: it diffs down to the run/atom level within a matched paragraph and marks up only the changed spans.

## Goal

When two paragraphs are matched as "the same paragraph, edited" (same position/high similarity, but the text differs), diff them **within** the paragraph and emit inline tracked changes for only the changed spans — instead of replacing the whole paragraph. A single changed word should produce a single struck word and a single inserted word, with the surrounding text untouched.

## Proposed approach

1. **Pair up modified paragraphs.** Extend the diff so that, in addition to `equal` / `insert` / `delete`, it can mark a `modify` (a matched pair whose text differs). Mirror the DOCX two-pass shape (`packages/docx-core/src/baselines/atomizer/hierarchicalLcs.ts` → `atomLcs.ts`): first align paragraphs, then, for an aligned-but-differing pair, do a second LCS *inside* the paragraph. A similarity threshold decides whether a pair is a true in-place edit (do an intra-paragraph diff) or effectively a replacement (fall back to the current whole-paragraph delete + insert). docx-core uses a Jaccard threshold (`DEFAULT_PARAGRAPH_SIMILARITY_THRESHOLD = 0.25`) as a reference point.

[...]

PR #348, opened 2026-06-08
feat(docx-mcp): ODF compare_documents — paragraph-granularity redline for .odt (Slice 1)

stevenobiajulu:
Closes #343 (Slice 1 of 3).

## What
Adds two-file ODF (`.odt`) support for `compare_documents`: redline two `.odt` files into ODF tracked-changes markup (`text:tracked-changes` / `text:changed-region` / `text:insertion` / `text:deletion` / `text:change-start` / `-end` / `text:change`). This is the ODF analogue of docx-core's `w:ins`/`w:del` atomizer — same LCS *algorithm*, a fresh *emitter* (ODF stores deletions out-of-line, not inline).

**Scope (Slice 1):** paragraph granularity, two-file mode. Intra-paragraph (run-level) diffs and session-mode compare are deferred to follow-up slices.

## How it works
- **odf-core `compare/`**: `diff.ts` (pure paragraph LCS, no DOM) → `emit.ts` (ODF tracked-changes emitter) → `index.ts` (`compareOdf(originalContentXml, revisedContentXml)`). Diff is separable and unit-tested in isolation.
- **Markup shapes were confirmed empirically** by driving LibreOffice to *author* each change and inspecting its `content.xml`:
  - Insertion brackets `change-start`/`-end` (EOF brackets backward); content stays inline.
  - Deletion stores paragraphs out-of-line in `text:deletion` and anchors an inline `text:change` in the nearest **surviving** paragraph (forward start / backward end). Consecutive deletions **coalesce** into one region. `text:change` is inline — never a block child of `office:text`.
  - A modified paragraph = delete + insert (so `modifications` is always 0 at this granularity); when both markers share a position, the deletion marker precedes the insertion `change-start`.
  - All-deleted (no anchor) fails closed.
- **No-leak**: `collectBlocks` (extracted to `shared/odf/blocks.ts`) and `buildSegments` skip the `text:tracked-changes` subtree, so deleted content never appears in `getParagraphs()`.
- **docx-mcp**: a stateless `tools/odf/compare_documents.ts` handler (two-file compare carries no `file_path`, so it cannot use the session-keyed resolver); `server.ts` routes two `.odt` inputs to it, guards `.odt` session-mode as `UNSUPPORTED_FOR_ODF`, and leaves the DOCX path untouched. Output-path clobber + write-policy match the DOCX tool.

## Optional-provider isolation
odf-core stays `private: true` and is reached only via the lazy `loadOdfCore()` dynamic import. `check:release-isolation` passes.

[...]

PR #366, opened 2026-06-10
feat(docx-mcp): ODF compare_documents session mode — redline a .odt session against its original

### LINKED ISSUE #357: ODF compare_documents: session mode (redline a .odt session's edits against its original)
## Background

`compare_documents` has two modes. **Two-file mode** takes `original_file_path` + `revised_file_path` and redlines one document against another. **Session mode** takes a single `file_path` that refers to an already-open editing session and redlines the session's accumulated edits against the document as it was when opened.

The DOCX path supports both. ODF (`.odt`) currently supports **two-file mode only** (landed in #348). A `.odt` passed as `file_path` (session mode) returns `UNSUPPORTED_FOR_ODF` — see the guard in `packages/docx-mcp/src/server.ts`.

## Problem

A common workflow is: open a `.odt`, make a series of edits through the editing tools (`replace_text`, `insert_paragraph`, `add_comment`, …), then produce a redline showing exactly what changed. Today, for `.odt`, that is not possible directly — the user has to save the edited document to a new path and run a manual two-file comparison against a saved copy of the original. Session mode removes that friction: compare the live session against the original it was opened from, in one call.

## Goal

Support `compare_documents` with a single `.odt` `file_path` (session mode): compare the session's current edited state against the original document the session was opened from, and write an ODF tracked-changes redline. Behavior should mirror the DOCX session-mode branch.

## Proposed approach

The pieces already exist; this is mostly wiring plus a session-aware entry point.

1. **Routing.** Unlike two-file mode (whose inputs are `original_file_path` / `revised_file_path` and therefore cannot go through the session-resolution chokepoint), session mode's input *is* a `file_path` pointing at a `.odt`. That means it can route through the existing ODF session-resolution path, which opens/reuses the `.odt` session. The `server.ts` guard that currently returns `UNSUPPORTED_FOR_ODF` for a `.odt` `file_path` on `compare_documents` should instead dispatch to an ODF compare handler that receives the resolved session.

2. **Before vs after.** An open `.odt` session already carries both states:
   - `originalBuffer` — the original `.odt` bytes captured when the session was opened.
   - `doc` — the live, edited `OdfDocument`.
   So: take the original `content.xml` from `originalBuffer` (load it as an archive and read its content part), take the revised `content.xml` from the live `doc` (`toXml()`), and feed both to the existing `compareOdf(originalContentXml, revisedContentXml)` engine. Build the redline on the revised state, write it to `save_to_local_path`.

3. **Handler shape.** The two-file handler is stateless (it loads two files itself). Session mode needs a session-aware path — either a second entry point in the same ODF compare module that takes the resolved session, or a thin wrapper that extracts the two `content.xml` strings from the session and calls the shared comparison core. The response should report `mode: 'session'` and include the session-resolution metadata the other session-mode tools attach, alongside the existing `granularity` / `stats` / `saved_to` fields.

## Considerations

- **Baseline fidelity.** The DOCX path compares against a normalized baseline (rather than the raw original bytes) to avoid spurious diffs caused by open-time normalization. The ODF lane does comparatively little normalization on open, so the original `content.xml` from `originalBuffer` is likely a faithful baseline — but this should be checked: parse → serialize round-tripping can reorder attributes or alter insignificant whitespace, which would surface as phantom changes. If that happens, capture a normalized baseline at session-open time (parse the original through the same pipeline the live doc uses) so an unedited session produces an empty redline.
- **No-op case.** A session with no edits should produce a redline with zero insertions/deletions (an empty change set), not an error.
- **Save-path safety.** Reuse the same output-path protections as two-file mode (refuse to overwrite the source; enforce the write-path policy).

## Affected code

- `packages/docx-mcp/src/server.ts` — replace the session-mode `.odt` guard on `compare_documents` with a dispatch to the ODF compare handler.
- `packages/docx-mcp/src/tools/odf/compare_documents.ts` — add a session-aware entry point that pulls original/revised `content.xml` from the session and calls the comparison core.
- Reference: `packages/docx-mcp/src/tools/compare_documents.ts` (DOCX session-mode branch) and the `OdfSession` shape (`originalBuffer`, `doc`) in `packages/docx-mcp/src/session/manager.ts`.
- Engine is unchanged: `packages/odf-core/src/compare/` already produces the redline from two `content.xml` strings.

## Verification

- Open a `.odt` session, make a few edits through the editing tools, run session-mode `compare_documents`, and confirm the stats reflect those edits and the redline round-trips in a real ODF consumer (accepting all changes reproduces the edited document; rejecting all reproduces the original).
- An unedited session produces an empty (no-op) redline.
- Two-file mode behavior is unchanged.

## Scope

Builds on the merged paragraph-granularity engine (#348) and is independent of the intra-paragraph follow-up (#356) — it inherits whatever granularity the engine supports at the time. Smaller and more mechanical than #356.

stevenobiajulu:
Fixes #357. Builds on the merged paragraph-granularity engine (#348, `add-odf-compare`); independent of the intra-paragraph follow-up (#356).

## What

`compare_documents` with a single `.odt` `file_path` (session mode) now compares the live session's edited state against the original it was opened from and writes an ODF tracked-changes redline — replacing the `UNSUPPORTED_FOR_ODF` guard. The engine is unchanged; this is routing plus a session-aware entry point (`odfCompareDocumentsSession`) registered in the standard ODF session lane (`dispatchOdf`), so the response carries the same session-resolution metadata as the other ODF session tools (`mode: 'session'`, `session_resolution`, `reused_session_context`, …).

## Key decisions

- **Fresh archive from `session.originalBuffer` for both baseline and redline packaging.** `saveOdfTo` stamps the live `session.archive` with edited content on every save, so it's not a valid baseline source, and the compare must not poison it with redline markup (pinned by OPCS-06). Packaging on the original package is valid under the current invariant that ODF session edit tools mutate `content.xml` only — stated in the handler comment for whoever breaks it.
- **No normalized baseline needed** (the issue's flagged risk): `compareOdf` diffs per-block *visible text*, which a parse→serialize round-trip doesn't alter. Pinned by OPCS-02, including a fixture with `text:s`/`text:tab`/`text:line-break`/`text:h`/entities/`office:annotation`.
- **Two-file precedence across all providers** (peer-review finding): previously a stray `.odt` `file_path` made a two-`.docx` compare fail `UNSUPPORTED_FOR_ODF`; the restructured routing mirrors the DOCX tool's own `twoFileMode` precedence (pinned by OPCS-07).
- **No `allow_overwrite` escape** for an output path resolving to the session original — comparison inputs are never overwritable (parity with #313).

## OpenSpec

- New change `add-odf-compare-session` (mcp-server, scenarios OPCS-01..07), validated `--strict`.
- Amends the pending `add-odf-compare` delta: drops the "session mode SHALL return UNSUPPORTED_FOR_ODF" clause and re-points `[OPCD-04]` at a still-unsupported tool (`accept_changes`) — the same flip `[OPLR-08]` got in #348 — so the generic guard keeps its own coverage.

## Verification

- Full local gate: build, lint, **all workspace tests**, `check:spec-coverage` (7/7 OPCS scenarios mapped), `check:tool-docs`, allure gates, conformance gates, cycles.
- **Real-document smoke** (NVCA model COI → `.odt` via LibreOffice): open session → `replace_text` + `insert_paragraph` + `add_comment` → session compare. Stats exactly match the edits (1 del + 2 ins; the comment correctly produces no text change). Redline reopens with revised text; deleted text stays out-of-line.
- **LibreOffice headless accept/reject round trip** on that redline (macro-injection oracle): accept-all ≡ edited session text, reject-all ≡ original — projection-to-projection with bookmark-ref display caches neutralized (LO re-resolves field caches at store time; all 35 bookmark targets and paragraph counts are exact).

### DIFF
--- a/openspec/changes/add-odf-compare-session/design.md
+++ b/openspec/changes/add-odf-compare-session/design.md
@@ -1,0 +1,54 @@
+# Design — ODF `compare_documents` session mode
+
+## Context
+
+The two-file ODF compare (Slice 1) is stateless: it loads both `.odt`s itself because two-file
+inputs carry no `file_path` and cannot route through `resolveOdfSessionForTool`. Session mode's
+input *is* a `file_path`, so it can — and should — use the standard ODF session lane. The engine
+(`compareOdf(originalContentXml, revisedContentXml, opts)`) is reused unchanged.
+
+## Goals / Non-Goals
+
+- Goals: session-mode `.odt` compare with response parity (`mode: 'session'` + session-resolution
+  metadata); no session-state mutation; an unedited session yields an empty redline.
+- Non-Goals: intra-paragraph granularity (#356); changing the engine; DOCX/gdocs behavior.
+
+## Decisions
+
+- **Route via the `dispatchOdf` handler map.** `compare_documents` is already in
+  `ODF_SUPPORTED_TOOLS`, so registering `odfCompareDocumentsSession` in `loadOdfHandlers` gives
+  open-or-reuse session resolution, staleness warnings, and metadata attachment for free — the same
+  shape as every other ODF session tool.
+- **Two-file precedence across all providers** (peer-review blocker). When both
+  `original_file_path` and `revised_file_path` are present, two-file mode wins even with a stray
+  `file_path` supplied — mirroring `compareDocuments_tool`'s own `twoFileMode` precedence.
+  Previously a stray `.odt` `file_path` made two-`.docx` compares fail `UNSUPPORTED_FOR_ODF`; after
+  the naive guard-swap it would instead have silently opened a session on the stray `.odt`.
+- **Baseline = raw `content.xml` from `session.originalBuffer`, no normalization.** `compareOdf`
+  diffs per-block *visible text* (`collectBlocks` + `buildSegments(...).visible`), so
+  parse→serialize differences (attribute order, whitespace-in-tags) cannot surface as phantom
+  changes. Verified by direct probe: `compareOdf(raw, serialize(parse(raw)))` over the sample
+  fixture returns `{insertions: 0, deletions: 0, modifications: 0}`. The DOCX-style normalized
+  baseline (`ensureBaselines`) is unnecessary for ODF; scenario [OPCS-02] pins this, including a
+  fixture with serialization-sensitive constructs (`text:s`, `text:tab`, `text:line-break`,
+  `text:h`, entity-escaped text, `office:annotation`).
+- **Fresh `OdfArchive` from `originalBuffer` for both baseline extraction and redline packaging.**
+  `SessionManager.saveOdfTo` stamps `session.archive` with `doc.toXml()` on every save, so the live
+  archive's `content.xml` may already be the *edited* state (unusable as a baseline), and writing
+  redline markup into it would poison the live session. Packaging the redline on the fresh original
+  archive is valid under a **current invariant**: today's ODF session edit tools mutate
+  `content.xml` only (the same premise `saveOdfTo` rests on), so the original package plus the
+  redline `content.xml` *is* the revised redline package. If a future ODF tool mutates non-content
+  parts (styles.xml, manifest, …), session compare must switch to a revised-session package
+  baseline or copy the modified parts. Scenario [OPCS-06] pins the no-mutation property.
+- **No `allow_overwrite` for the output path.** Comparison inputs are never overwritable (parity
+  with the DOCX tool and two-file mode, issue #313), even though `save` offers `allow_overwrite`.
+
+## Risks / Trade-offs
+
+- [Invariant drift: a future ODF tool touches non-content parts] → invariant stated in the handler
+  comment + here; [OPCS-06] catches session poisoning, and redline output is built from the
+  original package so such a tool's parts would be *missing* from the redline — the invariant note
+  tells the implementer where to fix it.
+- [Pending-change coupling: `add-odf-compare` not yet archived] → its delta is amended in this PR
+  (OPCD-04 re-point), following the #348 precedent for sequential slices.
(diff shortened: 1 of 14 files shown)


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `9c2067e52dc85dc08900b1771582de1b4263f052`
Built from: `main` @ `9c2067e`
Smoke: `npm run build && npm run lint:workspaces && npm run test:run && node --import tsx smoke-366.mts` (real-document ODF session-compare smoke)

**Steps**
- ✅ build (all workspaces, clean from main HEAD)
- ✅ lint
- ✅ tests (2,436 passed across workspaces; 2 expected-fail, exit code checked explicitly)
- ✅ session-mode `compare_documents` on a real `.odt` end-to-end

**Real-world fixtures**
- NVCA model COI (`tests/test_documents/nvca-regression/source.docx`, freshly converted to `.odt` via LibreOffice `writer8`)
  - Session edits through the MCP dispatch: `replace_text` ("Delaware corporation" → "Delaware incorporated entity"), `insert_paragraph` (after p0), `add_comment` (annotation)
  - Session compare returned `mode: 'session'`, `provider: 'odf'`, `granularity: 'paragraph'`, `stats: {insertions: 2, deletions: 1, modifications: 0}` — exactly the expected paragraph-granularity accounting (modified paragraph = del+ins, inserted paragraph = ins, annotation = no text change), with `session_resolution: 'reused'` metadata attached
  - Redline reopens via `read_file`: revised text present, deleted original stays out-of-line; `content.xml` carries one `text:tracked-changes` container
  - Post-compare `save` produced the edited document with **no** tracked-changes markup — the live session is not poisoned by comparison

Visual evidence: rendered + opened locally: `/tmp/automerge-smoke-366-1781070880/redline-p2-02.png` — page 2 shows the inserted SMOKE paragraph underlined as a tracked insertion and the preamble as a strike-through deletion + underlined insertion pair (paragraph granularity), with the comment anchor visible.
Log: `/tmp/automerge-smoke-366-1781070880.log` (local).

**Cleanup**
- ✅ deleted remote branch `357-odf-compare-session-20260610` (was already auto-deleted on merge)
- ✅ removed worktree `.worktrees/357-odf-compare-session-20260610`
- ✅ deleted local branch + pruned refs

Issue #367, opened 2026-06-10
ODF compare: reject-all of an end-of-document whole-paragraph replacement merges paragraphs (LibreOffice oracle)


stevenobiajulu:
## Symptom

For a **dissimilar whole-paragraph replacement of the LAST paragraph** (Slice-1 shape: whole-paragraph deletion + end-of-document insertion), LibreOffice's **Reject All** does not reproduce the original document: the preceding paragraph and the restored deleted paragraph merge, and a trailing empty paragraph is left behind.

Reproduced with the `.odt`-capable `runLibreOfficeOracle` (added in #356's PR):

- original: `["Stable one.", "Entirely different about apples."]`
- revised: `["Stable one.", "Zebras graze under moonlight."]`
- `compareOdf` → redline; **Accept All** → revised ✅; **Reject All** → `["Stable one.Entirely different about apples.", ""]` ❌

Mid-document replacement, pure end-of-document deletion, and pure end-of-document insertion all round-trip cleanly — the defect is exactly the EOF **composition**: the deletion's `text:change` point marker is anchored (forward) at the start of the *inserted* replacement paragraph, while the insertion uses the end-of-document bracket (`text:change-start` at the end of the preceding kept paragraph). Rejecting the insertion removes the span the deletion marker lives in, and the restore goes to the wrong place.

## Likely fix direction

Anchor the deletion marker **backward** (end of the preceding kept paragraph, artifact-first storage) when a delete run is immediately followed by an insert run reaching end-of-document — then oracle-confirm the marker order at the shared anchor (`text:change` vs `text:change-start`) and the reject/accept round-trip, G-case style.

## Notes

- Pre-existing Slice-1 (#348) behavior, discovered while adding the gated LibreOffice accept/reject round-trip for intra-paragraph compare (#356). Not a regression from #356 — similar replacements now become inline modify pairs, which round-trip cleanly, so the remaining exposure is dissimilar EOF replacements only.
- A gated characterization test pinning the current behavior lands with the #356 PR (`packages/odf-core/src/compare/lo_inline_roundtrip.test.ts`); flip it when fixing this.

PR #370, opened 2026-06-10
feat(odf-core,docx-mcp): ODF compare_documents — intra-paragraph (run-level) tracked changes (#356)

### LINKED ISSUE #356: ODF compare_documents: intra-paragraph (run-level) tracked changes for .odt

stevenobiajulu:
## Why

ODF `compare_documents` (Slice 1, #348) diffed at whole-paragraph granularity: any edited paragraph became a full-paragraph delete + insert — noisy for long legal prose — and `stats.modifications` was hard-wired to `0`. The DOCX path already marks up only the changed spans inside a matched paragraph. Fixes #356.

## What

**OpenSpec change `add-odf-intra-paragraph-compare`** (proposal + design + spec deltas, `openspec validate --strict` green). The design.md carries the LibreOffice **authoring-oracle decision log** (O1–O10): every inline markup shape was settled by having LO author the same edit and inspecting its `content.xml` *before* emitter code was written. Two plan corrections came out of it: a replace orders the insertion bracket BEFORE the deletion point marker (O3), and `text:h` modify pairs are fully supported via a mirrored heading storage block (O10 — no degrade needed).

- **odf-core**
  - `pairModifications` post-pass: aligned delete/insert pairs with Jaccard word-overlap ≥ 0.25 (`OdfCompareOptions.similarityThreshold`) become `modify` ops via a deterministic order-constrained DP (max pair count, then total similarity).
  - `inline_diff.ts`: token-level LCS (whitespace/non-whitespace run partition → word-boundary spans, lossless char offsets), prefix/suffix trimmed.
  - `inline_map.ts`: visible-offset → DOM mapping (`resolveOffset` splits `#text` and rebalances `text:s`; `extractVisibleRange` clones deleted content preserving `text:span`/link structure). The virtual `Segment` variant now carries its host element (additive).
  - `emit.ts`: explicit `equal | insert | delete | modify` planning lanes; modify pairs plan purely before any mutation; inserted spans stay inline bracketed; deleted spans leave a `text:change` point marker with content stored out-of-line in a block mirroring the host — **no merge artifact**; markers placed at descending offsets, LO's replace order at shared offsets; unmappable pairs **degrade** to the Slice-1 whole-paragraph shape before any markup exists.
  - Stats count changed-regions (modify pair = 1 modification + its spans in insertions/deletions), so stats always match the emitted markup.
- **docx-core**: `runLibreOfficeOracle` gains `.odt` jobs (writer8 filter, `content.xml` result); oracle entry points exported from the index.
- **docx-mcp**: `granularity: 'inline'`, rewritten stats message, catalog + regenerated tool reference, OPDI scenario tests (`TEST_FEATURE='add-odf-intra-paragraph-compare'`).

## Verification

- 105 odf-core tests (incl. structural round-trip invariants: accept-projection = revised text, reject-projection = original text, no-leak into `getParagraphs()`); docx-mcp ODF tool tests green; full local gate green (build, lint, test:run, spec-coverage, conformance checks, `openspec validate --strict`).
- **Gated LibreOffice round-trip** (`lo_inline_roundtrip.test.ts`, skips without soffice): a redline mixing a modify pair, a mid-document dissimilar replacement, and an EOF insertion **accepts to the revised texts and rejects to the originals through real LibreOffice**.
- **Document-shaped smoke** on a real NVCA stock purchase agreement (398 paragraphs, DOCX→ODT): `grep` → `replace_text` → `compare_documents` through `dispatchToolCall`; redline = exactly 4 changed-regions, opens in LO; accept-all reproduces the revision **exactly**; reject-all restores the edited paragraph exactly (residual diffs are LO re-evaluating cross-reference fields left dangling by the fixture's DOCX→ODT conversion, in untouched paragraphs only — change-free baseline shows zero diffs).

## Discovered (not fixed here)

The new oracle surfaced a **pre-existing Slice-1 defect**: reject-all of a dissimilar whole-paragraph replacement of the **last** paragraph merges paragraphs (deletion marker anchors inside the inserted paragraph whose EOF bracket is also rejected). Filed as #367 with a gated characterization test pinning the behavior; this PR narrows its exposure since similar replacements now become inline modify pairs.

### DIFF
--- a/openspec/changes/add-odf-intra-paragraph-compare/design.md
+++ b/openspec/changes/add-odf-intra-paragraph-compare/design.md
@@ -1,0 +1,97 @@
+# Design — ODF intra-paragraph tracked changes
+
+## Oracle decision log (LibreOffice 25.x authoring, 2026-06-10)
+
+Method: `.tmp/odf-inline-oracle/run_author_oracle.sh` — a throwaway clone of the
+`reference_libreoffice_macos_oracle` recipe. Each scenario loads a flat-ODF fixture headlessly,
+sets `RecordChanges = True`, performs cursor edits (`createTextCursorByRange` + `goRight` +
+`setString`), stores as `writer8`, and the script dumps `content.xml`. Each scenario below is a
+decision gate; the recorded shape is what LibreOffice itself authors and is therefore the
+emitter's target markup. (Basic gotcha for future runs: `BASE` is a reserved word in StarBasic
+— a `Const BASE` silently kills module compilation and soffice hangs without running the macro.)
+
+- **O1 — delete one word mid-paragraph.** Kept paragraph keeps the surrounding text with a
+  single point marker at the deletion offset: `Alpha <text:change/>charlie delta.`. The
+  `text:deletion` region stores ONE `text:p` (host paragraph's `text:style-name`) containing
+  exactly the deleted content (`bravo `). **No empty merge-artifact paragraph** — that artifact
+  belongs only to whole-paragraph deletion (a deleted paragraph break).
+- **O2 — insert one word mid-paragraph.** Inserted content stays inline, bracketed:
+  `Alpha bravo <text:change-start/>inserted <text:change-end/>charlie delta.`; the
+  `text:insertion` region holds only `office:change-info`.
+- **O3 — replace a word.** ORDER CORRECTION vs the original plan: LibreOffice emits the
+  insertion bracket FIRST and the deletion point marker AFTER it:
+  `charlie <text:change-start/>echo<text:change-end/><text:change/>.` — i.e. the replace's
+  deletion anchor sits after the replacement's `text:change-end`, not before its
+  `text:change-start`. (Slice 1's whole-paragraph OCMP-10 deletion-first rule does NOT carry
+  over to inline replaces.) Emitter rule: a delete span anchors at its revised offset, bumped
+  past a co-located insertion's bracket; at one offset the document order is `text:change-end`,
+  then `text:change`, then `text:change-start`.
+- **O4 — edits at paragraph start/end.** Point markers sit as the first / last inline child of
+  the paragraph: `<text:p><text:change/>bravo …` and `…paragraph <text:change/></text:p>`.
+- **O5 — delete crossing a bold `text:span` boundary.** One point marker at the span's start
+  offset, at the depth where the deletion begins (block level here): `Lea<text:change/>
+  <text:span>word</text:span> tail.`. The stored deletion preserves the inline structure of the
+  covered fragments: `<text:p>d <text:span T1>bold</text:span></text:p>`. No
+  `change-start`/`change-end` depth-mismatch question arises for deletions — they are point
+  markers; brackets only wrap insertions, whose content we control.
+- **O6 — partial delete of `text:s text:c="5"` (2 of 5).** Kept paragraph rebalances the run
+  (LO chose plain-space + `text:s text:c="2"`; count arithmetic is what matters). Stored
+  content is `<text:p><text:s text:c="2"/></text:p>` — the deleted spaces as a `text:s`.
+- **O7 — delete across `text:tab` / `text:line-break`.** Virtual elements are copied whole
+  into the stored content (`t<text:tab/>R`, `p<text:line-break/>D`); the kept paragraph joins
+  the remainders around a single point marker.
+- **O8 — whole-paragraph delete + inline delete at the next paragraph's start.** LibreOffice
+  COALESCES both into one `text:deletion` region: stored = full deleted paragraph + a second
+  `text:p` holding the inline-deleted prefix (`Second `), one marker at the kept paragraph's
+  start. (Slice 1's empty merge artifact is the special case of this where the deleted prefix
+  is empty.) Our emitter keeps them as SEPARATE regions (whole-paragraph run + inline modify) —
+  the composed two-marker shape is not LO-authorable, so it is verified by the accept/reject
+  round-trip oracle instead (accept-all → revised text, reject-all → original text).
+- **O9 — whole bold word deleted.** Stored content is the full `text:span`. The kept paragraph
+  shows LO's ODF whitespace normalization (second adjacent space became `<text:s/>`). Our
+  emitter does not edit kept text (the revised paragraph's encoding is kept verbatim; deleted
+  text was never in it), so this normalization concern applies only to LO-authored docs, not to
+  our output.
+- **O10 — inline delete inside a heading.** Fully supported by LO: the kept `text:h` carries
+  the point marker, and the deletion region stores a **`text:h`** mirroring the source block
+  (`text:style-name`, `text:outline-level`). PLAN CORRECTION: heading modify pairs do NOT
+  degrade; the stored block mirrors the host block's element name and attributes.
+
+## Emitter structure (three lanes)
+
+`emitTrackedChanges` plans purely, then mutates:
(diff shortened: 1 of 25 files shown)


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `a29e213d3139ab24f01db9886fec2ec25bef14e2`
Built from: `main` @ `a29e213`
Smoke: `npm run build && npm run lint:workspaces && npm run test:run` + real-document ODF compare flow (script in artifacts dir)

**Steps**
- ✅ build (clean, from main HEAD)
- ✅ lint (0 errors)
- ✅ full workspace test suite
- ✅ `grep → replace_text → save → compare_documents` through `dispatchToolCall`, then accept-all/reject-all through real LibreOffice (`runLibreOfficeOracle`, `.odt` jobs)

**Real-world fixtures**
- NVCA Series B Stock Purchase Agreement (`tests/test_documents/nvca-regression/filled.docx`, converted to `.odt` via LibreOffice — 398 paragraphs)
  - Edited: company name "Zephyr BioSystems" → "Acme Manufacturing" inside the preamble paragraph
  - Redline: `granularity: 'inline'`, `stats: { insertions: 2, deletions: 2, modifications: 1 }` — exactly 4 changed-regions, nothing else touched
  - Accept-all reproduced the revised document **exactly** (all 398 paragraphs); reject-all restored the edited paragraph exactly

Visual evidence: rendered + opened locally: `/tmp/automerge-smoke-370-20260610-0240/redline-01.png` — page 1 shows the two-word inline redline (struck "Zephyr BioSystems", inserted "Acme Manufacturing") inside an otherwise-untouched preamble paragraph, i.e. no whole-paragraph churn.

**Cleanup**
- ✅ remote branch already deleted (auto-delete on merge)
- ✅ removed worktree `.worktrees/356-odf-intra-paragraph-compare-20260610`
- ✅ deleted local branch + pruned remote refs

Log: `/tmp/automerge-smoke-370-20260610-0240.log` (local).

PR #365, opened 2026-06-10
feat(docx-core): formatting-fidelity comparison check (#363)

stevenobiajulu:
## Why

Closes #363.

The rebuild-elimination campaign (#347 → #358/#359) needs to *measure* formatting loss, and no current gate can: the round-trip safety oracle compares text projections only, and the LibreOffice oracle's `paragraphShape()` records only paragraph count + visible-text presence. Rebuild reconstructs `document.xml` from atoms and silently drops formatting that inplace preserves — and that loss passes every existing check.

## What

**`compareFormattingFidelity(expectedXml, actualXml)`** — deterministic, in-engine comparison of two `word/document.xml` views:

- **Content-anchored alignment**: paragraphs aligned by visible text (LCS), so it compares formatting *of the same content*, not raw XML diff noise. Content divergence is reported as alignment coverage, separately from the formatting tallies.
- **Char-weighted run comparison**: each character carries the canonical key of its nearest `w:rPr`; differing run splits with identical formatting (the normal inplace-vs-rebuild structural difference) register as zero divergence.
- **Four dimensions**: run (`w:rPr`), paragraph (`w:pPr`), table (`w:tblPr`/`w:trPr`/`w:tcPr` chain per paragraph-in-table), section (`w:sectPr`).
- **Output**: structured per-property divergence report (scope, friendly property name, added/removed/changed, canonical values, paragraph index, text sample) + a scalar score in [0,1] where **exact preservation scores exactly 1.0** — usable both as a ranking metric and an exact-preservation gate.

**`compareProjectedFormattingFidelity(expectedCandidate, actualCandidate)`** — compares accept-all and reject-all projections of two tracked-changes candidates (the projection-to-projection stance from #347), returning both reports plus `min(accept.score, reject.score)`. This is the entry point for "assert inplace fidelity ≥ rebuild's" per-fix gates.

Per the issue's note, LibreOffice is deliberately **not** used: it rewrites formatting on load/save, so the check is in-engine over our own emitted XML. Parsing reuses the existing `parseDocumentXml` (`@xmldom/xmldom`).

## Sanity measurement (use-case wiring)

[...]

Issue #363, opened 2026-06-10
Formatting-fidelity comparison check — the oracle the rebuild-elimination campaign needs


stevenobiajulu:
## Why

Strategic goal: eliminate every inplace→rebuild fallback and ultimately remove rebuild mode, because **rebuild reconstructs document.xml from atoms and loses formatting** that inplace preserves. To drive that campaign we need a way to *measure* formatting loss — and we currently don't have one.

The existing differential checks don't cover this:
- The round-trip safety oracle (`evaluateRoundTripSafety`, `pipeline.ts`) compares **text** projections (accept/reject), not formatting.
- The LibreOffice accept/reject oracle's `paragraphShape()` records only paragraph count + visible-text presence — **formatting-blind by design** (confirmed during the 2026-06 oracle vetting).

So rebuild's formatting loss passes every current gate silently. The rebuild-elimination campaign (#347 → #358/#359 and beyond) can't prove a fix preserves formatting, nor quantify what rebuild currently drops.

## What to prototype

A formatting-level comparison dimension that, given two `word/document.xml` views (e.g. inplace candidate vs rebuild candidate, or candidate vs expected), reports formatting divergence:
- Run properties (`w:rPr`: bold/italic/underline/color/size/font), paragraph properties (`w:pPr`: alignment, indentation, spacing, numbering/list refs, styles), table properties (`w:tcPr`/`w:trPr`/`tblPr`), and section properties.
- Aligned to the same content (so it compares formatting *of the same runs/paragraphs*, not raw XML diff noise).
- Output: a structured per-property divergence report + a scalar "formatting fidelity" score usable as a gate/metric.

Use it two ways:
1. **Quantify the status quo** — for each currently-failing inplace shape (the rebuild-fallback surface), measure what rebuild drops vs what inplace would have preserved, to prioritize the campaign and justify removing rebuild.
2. **Guard each fix** — as inplace learns to handle a shape (e.g. #358 inline-ins preserve), assert the inplace output's formatting fidelity ≥ rebuild's (ideally exact-preserving).

## Notes
- LibreOffice rewrites formatting on load/save (adds default `w:pPr`/`w:rPr`), so LO is NOT a usable formatting oracle — this must be a deterministic in-engine comparison over our own emitted XML.
- Prefer `@xmldom/xmldom` for any new XML parsing (project convention).

## Context
Enabler for the rebuild-elimination campaign behind the preserve semantics adopted in #347. Companion to the oracle-trust-boundary characterization issue.

Issue #368, opened 2026-06-10
rebuild drops w:hyperlink wrappers (link + r:id lost) and bleeds underline into the adjacent plain run


stevenobiajulu:
## What happens

On a real document, rebuild reconstruction **deletes the `w:hyperlink` element**: the link text survives as a plain (underlined) run, the `r:id` relationship reference is gone, and the run-merge that accompanies the flattening **absorbs the following plain-text period into the underlined link run** — so the rendered document shows an underlined "." that the source (and inplace) render plain, and the hyperlink is no longer clickable.

First real-document catch by the formatting-fidelity check (#363, merged in #365); found during the post-merge smoke ([PR comment](https://github.com/UseJunior/safe-docx/pull/365#issuecomment-4666891741)).

## Repro (real fixture)

1. Fixture: Common Paper Mutual NDA template (`open-agreements/content/templates/common-paper-mutual-nda/template.docx`). Its cover page contains `…posted at <w:hyperlink r:id="rId7"><w:r>[underlined]commonpaper.com/standards/mutual-nda/1.0</w:r></w:hyperlink><w:r>.</w:r>`.
2. Make a small revision (first-occurrence text replacements: `Confidential Information` → `Proprietary Information`, `mutual` → `reciprocal`) and run `compareDocumentsAtomizer(original, revised, { reconstructionMode: 'rebuild' })`.
3. Compare against the inplace candidate with `compareProjectedFormattingFidelity(inplaceXml, rebuildXml)`.

## Evidence

Counts over the emitted `document.xml`:

| view | `<w:hyperlink>` | `r:id="rId7"` refs |
|---|---|---|
| original | 1 | 1 |
| inplace candidate | 1 | 1 |
| **rebuild candidate** | **0** | **0** |

Accept-projection paragraph 2, rebuild candidate (hyperlink gone, period merged into the underlined run):

```xml
<w:r><w:rPr>…<w:u w:val="single"/></w:rPr><w:t>commonpaper.com/standards/reciprocal-nda/1.0.</w:t></w:r>
```

vs inplace (wrapper preserved, period in a separate plain run):

```xml
<w:hyperlink r:id="rId7"><w:r><w:rPr>…<w:u w:val="single"/></w:rPr><w:t>commonpaper.com/standards/reciprocal-nda/1.0</w:t></w:r></w:hyperlink><w:r><w:rPr>…</w:rPr><w:t>.</w:t></w:r>
```

[...]

PR #371, opened 2026-06-10
fix(docx-core): skip xmlns declarations in formatting-fidelity canonicalization

### LINKED ISSUE #369: formatting-fidelity canonicalization treats xmlns:* declarations as formatting — false section divergence on identical header/footer references
## What happens

The formatting-fidelity check (#363, merged in #365) canonicalizes property elements by serializing **all** attributes. Namespace-declaration attributes (`xmlns`, `xmlns:*`) are not formatting — they only record where a prefix is declared — but they currently enter the canonical key, so two semantically identical elements diverge.

Observed on the first real-document run (Common Paper Mutual NDA, inplace vs rebuild candidates — [smoke comment](https://github.com/UseJunior/safe-docx/pull/365#issuecomment-4666891741)): the rebuild path serializes `w:headerReference`/`w:footerReference` with an inline `xmlns:r` declaration while inplace inherits it from the root, producing a false `changed` divergence with **identical r:id values**:

```
[section] w:headerReference changed
  inplace(expected): <w:headerReference r:id="rId8" w:type="default"/>
  rebuild(actual):   <w:headerReference r:id="rId8" w:type="default" xmlns:r="http://schemas.openxmlformats.org/officeDocument/2006/relationships"/>
```

This zeroes the section dimension (1/1 divergent) and drags the overall score from ~0.83 to 0.7498 on that document for pure serialization noise. (Note: the initial smoke comment guessed "rel IDs differ" — wrong; the IDs are identical and the delta is only the `xmlns:r` declaration.)

## Fix

In `canonicalizeElement` (`packages/docx-core/src/baselines/atomizer/formattingFidelity.ts`), skip attributes whose name is `xmlns` or starts with `xmlns:` when building the canonical key. One-line filter + a scenario pinning that two views differing only in namespace-declaration placement score 1.0.

Worth doing before the section dimension is used as a hard gate in the rebuild-elimination campaign.

## Possibly related, decide in the fix PR

Whether `w:rsid*` / `w14:paraId`-style revision-save attributes should also be excluded if they ever appear inside property containers (today they sit on `w:p`/`w:r`, which the check doesn't compare, so no observed false positive).

Ref: #363, #365. Companion engine bug found in the same run: #368.

stevenobiajulu:
## Summary

`canonicalizeElement` in the formatting-fidelity check (#363/#365) serialized **all** attributes into the canonical property key, including namespace declarations. `xmlns`/`xmlns:*` attributes record where a prefix is bound, not formatting — and the two reconstruction modes legitimately bind the same prefix at different depths: rebuild serializes `w:headerReference`/`w:footerReference` with an inline `xmlns:r` declaration while inplace inherits it from the document root.

On the first real-document run (Common Paper Mutual NDA) this produced a false `changed` section divergence with **identical** `r:id` values, zeroing the section dimension (1/1 divergent) and dragging the overall score from ~0.83 to 0.7498 on pure serialization noise.

Fixes #369.

## Changes

- **`formattingFidelity.ts`** — filter `xmlns` / `xmlns:*` attributes out of the canonical key (one-line guard), with the rationale added to the `canonicalizeElement` doc comment.
- **`formattingFidelity.test.ts`** — pinning scenario reproducing the observed shape: `w:headerReference r:id="rId8"` with the `r` binding root-inherited on one side and declared inline on the other; asserts score 1.0, zero divergences. Verified red without the fix (1 failed / 9 passed), green with it (10/10).
- **OpenSpec delta spec** (`add-formatting-fidelity-comparison-check`, still active) — matching `namespace declaration placement does not register as formatting divergence` scenario.

## Decision on the issue's open question

The issue asked whether `w:rsid*` / `w14:paraId` revision-save attributes should also be excluded. **Deferred**: they sit on `w:p`/`w:r` elements, which the check never canonicalizes — only children of property containers enter the key — so there is no observable false positive to pin today. If they ever surface inside a property container, that's a new pinning scenario.

## Testing

Full pre-submit gate green in the worktree: `npm run build && npm run lint:workspaces && npm run test:run && npm run check:spec-coverage && npm run check:conformance-citations && npm run check:conformance-doc` (exit 0, gated on explicit exit codes).

### DIFF
--- a/openspec/changes/add-formatting-fidelity-comparison-check/specs/docx-comparison/spec.md
+++ b/openspec/changes/add-formatting-fidelity-comparison-check/specs/docx-comparison/spec.md
@@ -34,6 +34,11 @@ The system SHALL provide a deterministic, in-engine formatting comparison `compa
 - **WHEN** the actual view carries a `w:sectPr` whose `w:pgSz` differs from the expected view
 - **THEN** a section-scope divergence with property `w:pgSz` and kind "changed" is reported
 
+#### Scenario: namespace declaration placement does not register as formatting divergence
+
+- **WHEN** the actual view serializes a property element with an inline `xmlns:*` declaration while the expected view inherits the same namespace binding from an ancestor, all other attributes being identical
+- **THEN** the score is exactly 1.0 with zero divergences, because namespace declarations record where a prefix is bound rather than formatting
+
 #### Scenario: unaligned paragraph content lowers alignment coverage not formatting tallies
 
 - **WHEN** a paragraph's text differs between the two views so it cannot be content-aligned
--- a/packages/docx-core/src/baselines/atomizer/formattingFidelity.test.ts
+++ b/packages/docx-core/src/baselines/atomizer/formattingFidelity.test.ts
@@ -156,6 +156,33 @@ describe('Formatting-fidelity comparison check', () => {
     },
   );
 
+  humanReadableTest.openspec('namespace declaration placement does not register as formatting divergence')(
+    'Scenario: namespace declaration placement does not register as formatting divergence',
+    (_: AllureBddContext) => {
+      const R_NS = 'http://schemas.openxmlformats.org/officeDocument/2006/relationships';
+      // inplace-style: r: bound once on the root, inherited by w:headerReference.
+      const expected =
+        `<?xml version="1.0" encoding="UTF-8" standalone="yes"?>` +
+        `<w:document xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main"` +
+        ` xmlns:r="${R_NS}"><w:body>` +
+        `<w:p><w:r><w:t>content</w:t></w:r></w:p>` +
+        `<w:sectPr><w:headerReference r:id="rId8" w:type="default"/></w:sectPr>` +
+        `</w:body></w:document>`;
+      // rebuild-style: the same binding declared inline on the element itself,
+      // with an identical r:id (the false "changed" divergence from #369).
+      const actual = docXml(
+        `<w:p><w:r><w:t>content</w:t></w:r></w:p>` +
+          `<w:sectPr><w:headerReference r:id="rId8" w:type="default" xmlns:r="${R_NS}"/></w:sectPr>`,
+      );
+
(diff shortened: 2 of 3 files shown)


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `1d0bbcb03b0d904c044a9a3e83da0d339d61cd63`
Built from: `main` @ `1d0bbcb`
Smoke: `npm run build && npm run lint:workspaces && npm run test:run` + real-document fidelity run (`node /tmp/smoke-371-real-doc.mjs`)

**Steps**
- ✅ build (clean, from main HEAD)
- ✅ lint (all workspaces)
- ✅ tests (2437 passed across 7 workspace runs, 2 expected-fail, 35 skipped)
- ✅ `compareProjectedFormattingFidelity` driven end-to-end over real pipeline candidates, with explicit assertions on the #369 fix

**Real-world fixtures**
- `common-paper-mutual-nda` (local template at `open-agreements/content/templates/common-paper-mutual-nda/template.docx`) — same document and same tracked edits as the pre-fix #365 baseline, so the scores are directly comparable
  - Tracked edit by "Jane Doe": "Confidential Information" → "Proprietary Information", "mutual" → "reciprocal"
  - Both reconstruction modes produced candidates with tracked-changes markup (2× `w:ins`, 2× `w:del` each); inplace stayed inplace (no fallback)
  - **Fix confirmed on the exact pre-fix failure**: section dimension now **0/1 divergent on both projections** (was 1/1 — the false `w:headerReference` xmlns:r divergence is gone), overall score recovered **0.7498 → 0.9998**
  - The only surviving divergence is the known #368 engine artifact (`underline` added on a "." run, char-weighted 1/1589) — exactly what should remain
  - Gate property held: self-comparison of the inplace candidate scored exactly 1.0

Visual evidence: skipped — the change is canonicalization-key-only (a measurement filter); no document output differs, so a rendered page would show nothing.
Log: `/tmp/automerge-smoke-371-20260610022632.log` (local).

**Cleanup**
- ✅ remote branch `369-canonicalize-skip-xmlns-20260610` already deleted (repo auto-delete); stale ref pruned
- ✅ removed worktree `.worktrees/369-canonicalize-skip-xmlns-20260610`
- ✅ deleted local branch

PR #374, opened 2026-06-10
fix(odf-core): anchor EOF-replacement deletion backward so reject-all restores the original (#367)

### LINKED ISSUE #367: ODF compare: reject-all of an end-of-document whole-paragraph replacement merges paragraphs (LibreOffice oracle)

stevenobiajulu:
Fixes #367.

## Problem

For a **dissimilar whole-paragraph replacement of the LAST paragraph** (Slice-1 shape: whole-paragraph deletion + end-of-document insertion), LibreOffice's Reject All did not reproduce the original: the preceding paragraph and the restored deleted paragraph merged, and a trailing empty paragraph was left behind.

The deletion's inline `text:change` point marker anchored **forward** — at the start of the *inserted* replacement paragraph — while the insertion used the end-of-document bracket (`text:change-start` at the end of the preceding kept paragraph). The deletion marker therefore sat **inside** the insertion span; rejecting the insertion removed the span the marker lived in, and the restore went to the wrong place.

## Fix

In `emitTrackedChanges`, a deletion run whose forward anchor would be the first paragraph of an insert run that itself reaches end-of-document now anchors **backward**:

- empty merge artifact stored **first** in the `text:deletion` region (the backward shape, matching pure EOF deletion);
- inline `text:change` marker at the **end of the preceding kept paragraph**, placed **before** the insertion's end-anchored `text:change-start`, so it stays outside the insertion span.

With no preceding paragraph (`revisedCursor` 0, whole-document replacement) the forward placement stands — there is nothing to anchor backward to.

Rebased onto #370 (intra-paragraph compare): the rule lives in the lane-1 whole-paragraph path, so it covers both a plain delete+insert from the paragraph diff and a **degraded modify pair** at end-of-document (both re-route through the same delete-run/insert-run shape). Similar replacements that survive as inline modify pairs are untouched.

## Oracle confirmation (G-case style)

- #370's gated characterization test (`lo_inline_roundtrip.test.ts`) pinned the broken reject output for exactly this composition, with the comment "flip when fixing #367" — **flipped here** to assert the true round-trip, and it passes against live LibreOffice.
- New gated `lo_paragraph_roundtrip.test.ts` batches four compositions into one headless launch and asserts accept-all → revised, reject-all → original for each:

| Composition | Before fix | After fix |
|---|---|---|
| EOF whole-paragraph replacement (#367 repro) | reject-all → `["Stable one.Entirely different…", ""]` ❌ | ✅ |
| mid-document whole-paragraph replacement | ✅ | ✅ (guards the anchoring change) |
| pure EOF deletion | ✅ | ✅ |
| pure EOF insertion | ✅ | ✅ |

Verified locally that the round-trip test **fails without the emitter change** with exactly the issue's failure shape, and passes with it (LibreOffice 25.8, macOS).

Unit test `[OCMP-11]` pins the markup shape without LibreOffice (backward artifact-first storage, marker before `change-start` at the shared anchor, nothing anchored inside the inserted paragraph), and the `add-odf-compare` spec delta gains the matching scenario (validated `--strict`).

The `.odt` oracle-job support and `@usejunior/docx-core` re-export this PR originally carried were absorbed by #370's merge — the diff is now odf-core-only.

## Pre-submit

`npm run build && npm run lint:workspaces && npm run test:run && npm run check:spec-coverage && npm run check:conformance-citations && npm run check:conformance-doc` — all green locally after the rebase (including the LibreOffice-gated suites).

### DIFF
--- a/openspec/changes/add-odf-compare/specs/odf-core/spec.md
+++ b/openspec/changes/add-odf-compare/specs/odf-core/spec.md
@@ -39,6 +39,15 @@ scanning existing ids; `office:change-info` carrying `dc:creator` from `opts.aut
   marker and the insertion's `text:change-start` target the same position (the start of the
   inserted replacement paragraph), the `text:change` (deletion) marker SHALL be emitted BEFORE
   the `text:change-start` (insertion) marker.
+- A deletion run whose following paragraph belongs to an inserted run that itself reaches
+  **end of document** (a whole-paragraph replacement of the LAST paragraph) SHALL anchor
+  **backward** — empty merge artifact first, inline `text:change` marker at the end of the
+  preceding surviving paragraph, BEFORE the insertion's end-anchored `text:change-start` — so
+  the marker stays outside the insertion span and rejecting the insertion cannot remove the
+  deletion's restore point. (A forward marker would sit inside the inserted paragraph, and
+  reject-all would merge the preceding paragraph with the restored one and leave a trailing
+  empty paragraph.) When no preceding paragraph exists (`revisedCursor` 0), the forward
+  placement stands.
 
 `compareOdf` SHALL preserve the unchanged paragraphs' visible text and the rest of the revised
 document (styles, manifest, untouched parts) so the redline round-trips.
@@ -85,3 +94,7 @@ and the visible-text walk SHALL skip the `text:tracked-changes` subtree.
 #### Scenario: [OCMP-10] Modified paragraph orders deletion before insertion markers
 - **WHEN** `compareOdf` emits a modified paragraph whose deletion `text:change` and insertion `text:change-start` target the start of the same replacement paragraph
 - **THEN** the `text:change` (deletion) marker precedes the `text:change-start` (insertion) marker in that paragraph
+
+#### Scenario: [OCMP-11] Replaced last paragraph anchors the deletion backward, outside the insertion bracket
+- **WHEN** `compareOdf` emits a whole-paragraph replacement of the LAST paragraph (a deletion run immediately followed by an inserted run reaching end-of-document)
+- **THEN** the deletion stores the empty merge artifact first and its inline `text:change` marker sits at the end of the preceding surviving paragraph, BEFORE the insertion's end-anchored `text:change-start` (never inside the inserted replacement paragraph), so LibreOffice's reject-all restores the original paragraphs without merging or leaving a trailing empty paragraph
--- a/packages/odf-core/src/compare/emit.test.ts
+++ b/packages/odf-core/src/compare/emit.test.ts
@@ -139,6 +139,25 @@ describe('compareOdf — ODF tracked-changes emission', () => {
     expect(firstTwo).toEqual(['change', 'change-start']); // deletion point BEFORE insertion start
   });
 
+  it('[OCMP-11] replaced LAST paragraph: deletion anchors backward, outside the insertion bracket', () => {
+    // Whole-paragraph replacement at end-of-document (issue #367): the insertion bracket is
+    // end-anchored, so the deletion marker must move to the end of the preceding kept paragraph
+    // (before the change-start) instead of the start of the inserted replacement paragraph.
+    const { contentXml: out, stats } = compareOdf(contentXml(['A', 'B']), contentXml(['A', 'X']));
+    expect(stats).toEqual({ insertions: 1, deletions: 1, modifications: 0 });
+    const ot = officeText(out);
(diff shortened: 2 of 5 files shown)


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `de2aac4` (squash)
Built from: `main` @ `de2aac4`
Smoke: real-document ODF compare round-trip (`/tmp/automerge-smoke-374-20260610/smoke.mjs`) + full workspace suite

**Steps**
- ✅ build (root `npm run build` at merged HEAD)
- ✅ tests (full `npm run test:run`: all workspaces green, incl. the LibreOffice-gated odf-core round-trip suites)
- ✅ merged `compareOdf` + `runLibreOfficeOracle` `.odt` accept/reject driven end-to-end on a real document

**Real-world fixtures**
- `common-paper-mutual-nda` (local template at `open-agreements/content/templates/common-paper-mutual-nda/template.docx`, converted to `.odt` via LibreOffice — 56 blocks, ends with the signature table + one empty trailing paragraph)
  - Edit: replaced the final (empty) paragraph with a closing sentence, author "Jane Doe" — exactly the #367 EOF whole-paragraph replacement; stats `{insertions: 1, deletions: 1, modifications: 0}`
  - Redline markup carries the fixed shape: deletion `text:change` immediately BEFORE the insertion's end-anchored `text:change-start` at the shared anchor ✅
  - LibreOffice **Accept All** reproduces the revised document exactly ✅
  - LibreOffice **Reject All** restores all 56 original blocks' text correctly, but leaves one stray trailing empty paragraph (57 blocks) — **pre-existing, fix-independent**: emulating the pre-#374 forward-anchored markup on the same document produces the identical 57-block output. The variable is the backward anchor landing in a *table-cell* paragraph (signature table). Filed as #380.
  - Body-level controls in the same oracle launch round-trip exactly on both accept and reject: EOF replacement with text-bearing deleted paragraph (the #374 composition) and with an empty deleted paragraph.

Visual evidence: rendered + opened locally: `/tmp/automerge-smoke-374-20260610/redline-lastpage.png` — the NDA cover page with the inserted closing line displayed as a redline (change bars beside the signature table's Date row). Rejected state not rendered: visually identical to the original (the #380 residual is an invisible empty paragraph).
Log: `/tmp/automerge-smoke-374-20260610.log` (local).

**Cleanup**
- ✅ deleted remote branch `367-odf-eof-replacement-reject-anchor-20260610` (already auto-deleted on merge; ref pruned)
- ✅ removed worktree `.claude/worktrees/fix-367-odf-eof-replacement-reject`
- ✅ deleted local branch; main checkout restored to its prior branch `378-odf-inline-group-replacements-20260610`

PR #384, opened 2026-06-10
fix(docx-core): surface text-empty paragraphs that anchor a footnote reference

stevenobiajulu:
Fixes #185

## Problem

`buildNodesForDocumentView` skipped any paragraph with no visible text unless it sat inside a table. A body paragraph whose only content is a footnote-reference run — a real shape: the NVCA SPA fixture anchors footnote id=47 (display 46) to exactly such a paragraph — was therefore dropped from the document view entirely. That made the loss silent in the most-used read path: `get_footnotes` reported the note as anchored, but `read_file` could never render the anchor paragraph or its `[^N]` marker in any format, and `node_ids` probes for the anchor returned zero nodes. On the NVCA fixture only 107 of 108 eligible footnotes were reachable through the view.

## Fix

Keep a text-empty paragraph when it carries a visible (non-reserved) footnote reference — the keep-predicate is `getFootnoteMarkersForParagraph(...)`, the same predicate the later marker-injection pass uses, so a kept paragraph always renders at least its `[^N]` marker and never a truly blank node. No changes to the pagination, marker, or rendering pipelines. Paragraphs that are empty for spacing only are still skipped; view shapes for footnote-free documents are unchanged. The extra check only runs for text-empty non-table paragraphs and short-circuits when the document has no footnotes.

## Behavior change to be aware of

The NVCA document view gains one node (367 → 368, at 1-based position 94), so later 1-based `offset` windows on footnote-bearing documents shift by one. No repo test pinned the old offsets (full pre-submit gate is green), but external consumers that hard-code numeric offsets into such documents may observe drift — that drift is the fix working: the document was always 368 anchorable paragraphs; one was silently hidden.

## Out of scope (filed as follow-ups)

- #382 — footnote markers render doubled (`[^N][^N]`) in `text`/`tagged_text` because both the view-level injection and `read_file`'s suffix pass fire. Pre-existing on main for **all** footnote paragraphs (verified: 107/107 doubled before this fix); orthogonal to the drop fixed here. The new tests assert `/^\[\^N\](?:\[\^N\])?$/` so they tolerate the current duplicate and a future single-marker fix, but fail on zero or triple markers.
- #383 — endnote-reference-only, comment-reference-only, and drawing-only paragraphs are still dropped by the same skip. This PR is deliberately bounded to the footnote read path per one-fix-per-PR.

## Tests

[...]

Issue #207, opened 2026-05-15
Add opt-in include_footnotes to read_file for single-call body + footnotes retrieval


stevenobiajulu:
# Add opt-in `include_footnotes` to `read_file` for single-call body + footnotes retrieval

## Problem

Today `read_file` returns the contents of `word/document.xml` — body paragraphs with inline footnote-reference markers `[^N]`. Footnotes themselves live in `word/footnotes.xml` and require a separate `get_footnotes` call. To fully reconstruct a document an agent must:

1. Call `read_file(format="json")` to get body paragraphs with `[^N]` markers.
2. Call `get_footnotes()` to get footnote bodies.
3. Stitch them together, preserving order and any multi-paragraph footnote bodies.

That's fine for "body only, footnotes are noise" use cases. But for full-fidelity ingest — converting a DOCX to canonical Markdown for storage, citation, or downstream search indexing — it's two calls + a manual stitch, easy to get wrong (especially when a footnote body contains multiple paragraphs or its own inline formatting).

**Concrete example**: the NVCA Model Stock Purchase Agreement has 109 footnotes totaling ~43,000 characters of substantive drafting guidance. Those footnotes ARE the document's value for downstream legal-AI consumers — they explain why each section is drafted the way it is. Losing them, or mis-stitching them, means losing the most useful part of the document.

## Proposal

Add an opt-in parameter:

```jsonc
{
  "name": "read_file",
  "input": {
    "path": "/path/to/doc.docx",
    "format": "json",
    "include_footnotes": true  // new; default false
  }
}
```

### Recommended shape — top-level `footnotes` field

Avoid inlining footnote bodies into `content[]` nodes — that would break the 1:1 `content[]` array index invariant that edit tooling relies on. Instead, sidecar them at the top level:

```jsonc
{
  "content": [
    { "id": "_bk_a1b2c3", "text": "This Agreement is made as of...[^1]", ... },
    ...
  ],
  "footnotes": [
    {
      "id": "1",                                 // matches the [^N] in body
      "ref_paragraph_ids": ["_bk_a1b2c3"],       // ARRAY — usually 1 elt, but DOCX has been
                                                 // observed (illegally) reusing the same
                                                 // footnote ID from multiple paragraphs
      "paragraphs": [                            // multi-paragraph footnote support
        { "text": "See generally Smith v. Jones, ...", "style": "FootnoteText" },
        { "text": "Continuation paragraph.", "style": "FootnoteText" }
      ]
    },
[...]

Issue #185, opened 2026-05-10
read_file: footnotes anchored to non-body paragraphs are silently dropped from inline output


stevenobiajulu:
## Context

While implementing inline footnote bodies for `read_file` (#158), we discovered that the NVCA SPA fixture has a footnote whose anchor paragraph is not surfaced by `buildDocumentView()` — so it's invisible to inline-windowed pagination.

Concrete numbers from `tests/test_documents/nvca-regression/source.docx`:

| Path | Count |
|---|---|
| `get_footnotes` total | 109 |
| Eligibility filter (display_number > 0, non-empty, anchored) | 108 |
| Inline-attached via paginated `read_file(format='json')` | **107** |

The missing one is footnote `id=47` (display 46), anchored to `_bk_6d177a97f7e6`. Calling `read_file(node_ids: ['_bk_6d177a97f7e6'])` returns zero nodes — the paragraph exists in `getFootnotes()`'s anchor scan but not in the document view. It likely lives in a non-body part (header, footer, comment body, footnote body itself), or in a structure `buildDocumentView()` skips.

## Why it matters

The motivating use case for #158 was *fidelity*: `legal-context` was silently losing 109 footnotes from this exact fixture. We've recovered 107, but a fully faithful single-call render still loses one. Whether 107 or 108 is the right denominator depends on the maintainer's intent for `buildDocumentView()`.

## Reproduction

```ts
const r = await getFootnotes(mgr, { file_path: NVCA });
// r.footnotes contains id=47, anchored_paragraph_id='_bk_6d177a97f7e6'

const probe = await readFile(mgr, {
  file_path: NVCA,
  format: 'json',
  node_ids: ['_bk_6d177a97f7e6'],
});
// probe.content === '[\n\n]'  — paragraph not in document view
```

The new test `NVCA SPA fixture: get_footnotes returns 109; paginated JSON walk inlines 107 bodies` (`packages/docx-mcp/src/tools/read_file_footnotes.test.ts`) pins this down with an explanatory comment.

## Options

1. **Investigate the unsurfaced paragraph** — figure out which non-body part it lives in and decide whether `buildDocumentView()` should include it.
2. **Surface unreachable footnotes via a top-level field** in `read_file` output (e.g. `unreachable_footnotes: [{id, display_number, text, anchored_paragraph_id}]`) so callers see them without a separate `get_footnotes` call.
[...]

PR #387, opened 2026-06-11
fix(docx-core): surface text-empty paragraphs anchoring endnotes, comments, or drawings

### LINKED ISSUE #383: Document view still drops text-empty paragraphs anchoring endnote/comment references or drawings
## Problem

#185's fix taught `buildNodesForDocumentView` to keep a text-empty paragraph when it carries a visible **footnote** reference (`getFootnoteMarkersForParagraph(...).length > 0`). The same silent-drop applies to other text-empty-but-meaningful paragraph shapes, which remain unfixed:

- `<w:endnoteReference/>`-only paragraphs — same semantics as footnotes; an endnote anchored there is invisible to the view (and to any future endnote read surface).
- `<w:commentReference/>`-only paragraphs — a comment anchored there can't be reached via `node_ids`, and comment-rendering passes can't attach it to a visible node.
- Drawing-only paragraphs (`<w:drawing>`/images with no text) — dropped wholesale.

## Why deferred

#185 was scoped one-fix-per-PR to the footnote read path (the case with a concrete real-document reproduction — NVCA SPA footnote id=47). Both peer reviewers (Codex + agy) confirmed endnote/comment shapes are susceptible but agreed bounded scope was acceptable with this follow-up filed.

## Suggested shape

Generalize the keep-predicate in `buildNodesForDocumentView` (`packages/docx-core/src/primitives/document_view.ts`) from "has footnote markers" to "has anchoring content" — footnote/endnote/comment references at minimum (drawings need a decision about what the node's text should be). Endnotes can mirror the footnote marker pipeline; comments already have their own attachment pass that just needs the node to exist.

Ref: #185

stevenobiajulu:
## Problem

#384 (issue #185) taught `buildNodesForDocumentView` to keep a text-empty paragraph when it carries a visible **footnote** reference. Issue #383 covers the remaining silent-drop shapes, which were still removed from the document view entirely:

- **Endnote-reference-only paragraphs** — the anchored endnote was invisible to the view and to any future endnote read surface.
- **Comment-reference-only paragraphs** — the comment could not be reached via `node_ids`, and comment-rendering passes had no node to attach the thread to.
- **Drawing-only paragraphs** (`w:drawing`, plus `w:pict`/`w:object`, the same run-level atoms the atomizer treats as atomic) — dropped wholesale.

## Fix

The keep-predicate generalizes from "has footnote markers" to "has anchoring content" via a new `paragraphHasAnchoringContent` helper, with two refinements out of peer review (Codex + agy, both dynamic):

- **Comment range markers count as anchors** (agy, SHOULD-FIX): `getComments` resolves `anchored_paragraph_id`/`end_paragraph_id` from where `w:commentRangeStart`/`End` sit (`comments.ts` `resolveCommentRangeMetadataInParagraph`), so a range starting in an otherwise-empty paragraph would otherwise report a dangling anchor ID that no `node_ids` probe can resolve.
- **Anchors inside `w:del`/`w:moveFrom` do not count** (Codex, SHOULD-FIX): deleted content is invisible to the view's text extraction (`getParagraphText` reads `w:t`, never `w:delText`), and a tracked comment-delete deliberately leaves the `commentReference` run under `w:del` — resurrecting that paragraph as a blank node would contradict how fully-deleted text paragraphs render.

Kept nodes render with **empty text** — the shape already established by empty table-cell paragraphs — so no placeholder text is invented. Footnote references keep their existing display-map path so `[^N]` markers still render, and spacing-only empty paragraphs are still skipped (view shapes unchanged for documents without these constructs).

## Scope notes

- Endnote marker/body rendering is a separate read surface that does not exist yet; the natural follow-on is opt-in endnote inlining alongside #207.
- `mc:AlternateContent` blocks are covered indirectly: both the `mc:Choice` `w:drawing` and the `mc:Fallback` `w:pict` match the namespace-aware descendant scan.

## Testing

- New `read_file_anchor_only_paragraphs.test.ts` (5 MCP-level end-to-end tests): endnote-only, comment-anchor-only, and dangling-range-start paragraphs surface and resolve via `node_ids` (comment threads attach to the kept nodes); the drawing/pict/object trio surfaces while a spacing-only sibling stays hidden; `w:del`-wrapped commentReference/drawing paragraphs stay hidden.
- Red→green verified both ways by rebuilding docx-core at each state: the three surfacing tests fail without the keep-predicate change, and the two peer-review-driven tests fail against the pre-review predicate.
- Full pre-submit suite green locally: build, lint:workspaces, test:run, check:spec-coverage, check:conformance-citations, check:conformance-doc.

Fixes: #383
Ref: #185

### DIFF
--- a/packages/docx-core/src/primitives/document_view.ts
+++ b/packages/docx-core/src/primitives/document_view.ts
@@ -294,6 +294,59 @@ function getFootnoteMarkersForParagraph(
 }
 
 /**
+ * Paragraph content that makes a text-empty paragraph meaningful on its own:
+ * an endnote or comment anchored to the paragraph (the comment range markers
+ * are what `getComments` resolves `anchored_paragraph_id`/`end_paragraph_id`
+ * from, so dropping their paragraph leaves a dangling anchor ID no node_ids
+ * probe can resolve), or embedded visual content (DrawingML drawing, VML
+ * picture, embedded object). Dropping such a paragraph from the document view
+ * severs the anchored note/comment from every read surface and silently
+ * hides images.
+ *
+ * Footnote references are handled separately via the display map so their
+ * [^N] markers render; the shapes here only need the node to exist.
+ * @see #383
+ */
+const ANCHORING_CONTENT = [
+  W.endnoteReference,
+  W.commentReference,
+  W.commentRangeStart,
+  W.commentRangeEnd,
+  W.drawing,
+  W.pict,
+  W.object,
+] as const;
+
+/**
+ * True when `el` sits inside a `w:del` or `w:moveFrom` revision wrapper below
+ * the paragraph. Deleted/moved-from content is invisible to the view's text
+ * extraction (`getParagraphText` reads `w:t`, never `w:delText`), so an
+ * anchor that only survives inside a tracked deletion — e.g. the
+ * `w:commentReference` a tracked comment-delete leaves under `w:del` — must
+ * not resurrect its paragraph as a blank visible node.
+ */
+function isInsideRemovedRevisionWrapper(el: Element, paragraph: Element): boolean {
+  let cur = el.parentNode as Element | null;
+  while (cur && cur !== paragraph) {
+    if (cur.namespaceURI === OOXML.W_NS && (cur.localName === W.del || cur.localName === W.moveFrom)) {
+      return true;
+    }
+    cur = cur.parentNode as Element | null;
+  }
+  return false;
+}
+
+function paragraphHasAnchoringContent(p: Element): boolean {
+  return ANCHORING_CONTENT.some((localName) => {
+    const els = p.getElementsByTagNameNS(OOXML.W_NS, localName);
+    for (let i = 0; i < els.length; i++) {
+      if (!isInsideRemovedRevisionWrapper(els.item(i) as Element, p)) return true;
+    }
+    return false;
+  });
+}
+
+/**
  * Inject footnote markers into a text string at the given offsets.
  * Markers must be sorted descending by offset.
  *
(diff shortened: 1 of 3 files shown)


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `cc0ec1284af6595c22459cdebf5d8076f1db6a1b`
Built from: `main` @ `cc0ec12`
Smoke: clean `npm run build` + full `npm run test:run` + real-document drive via the shipped MCP tool layer (`smoke-387.mts`, SessionManager → open_document/read_file/get_footnotes/add_comment/replace_text/save/get_comments)

**Steps**
- ✅ build (clean, from merged main HEAD)
- ✅ tests (2,495 passed across all workspaces; 2 expected-fail, 45 skipped)
- ✅ real-document anchor sweep + #383-shape round trip (below)

**Real-world fixtures**
- `nvca-regression/source.docx` (real NVCA Stock Purchase Agreement, in-repo at `tests/test_documents/nvca-regression/`)
  - Full view read: 368 nodes; 109 footnotes listed, 108 anchored, **0 unsurfaced**
  - All 108 anchored footnotes resolve via `node_ids` probes; 108 `[^N]` markers render in the view
- `common-paper-mutual-nda` (local template at `~/Projects/open-agreements/content/templates/common-paper-mutual-nda/template.docx`)
  - Realistic lawyer flow, entirely through shipped tools: `add_comment` (author = "Jane Doe") on the real sentence "Neither party has changed the Standard Terms…", then `replace_text` tracked-deleted that sentence so the paragraph became text-empty while still anchoring the comment — the exact #383 shape — then `save` and reopen in a fresh session
  - After round trip: `get_comments` reports the comment with a non-null `anchored_paragraph_id`, the anchor resolves to a view node with `clean_text === ""`, and the comment thread renders against it. Before this PR that paragraph was silently dropped and the anchor ID dangled.

Visual evidence: skipped (an empty paragraph kept vs dropped is structurally invisible on a rendered page)
Log: `/tmp/automerge-smoke-387-1781138145.log` (local; includes two intermediate smoke-script reruns — pagination limit and missing `author` param were script bugs, not product failures)

**Cleanup**
- ✅ remote branch `383-surface-anchoring-empty-paragraphs-20260610` already deleted (repo auto-delete)
- ✅ removed worktree `.worktrees/383-anchor-paragraphs`
- ✅ deleted local branch (squash-merge force-delete after MERGED verification) + pruned refs

PR #386, opened 2026-06-11
fix(docx-mcp): render each footnote marker exactly once in read_file output

stevenobiajulu:
## Problem

Every footnote-anchoring paragraph rendered its marker doubled (`[^N][^N]`) in `read_file`'s `text`/`tagged_text` fields — all 107 reachable markers on the NVCA SPA fixture. Two independent passes both injected the marker:

1. **View level** — `buildNodesForDocumentView` injects `[^N]` into `tagged_text`/`text` at the reference's visible offset (`injectFootnoteMarkers`, tag-aware via `findTaggedTextInsertionIndex`).
2. **Tool level** — `read_file`'s enrichment appended the marker again as a suffix to `clean_text`, `tagged_text`, and `text`.

`clean_text` got a single marker (the view never touches it), so the fields also disagreed on count (1 vs 2).

## Fix

The view-level injection wins for `tagged_text`/`text`: its placement is faithful and it feeds every downstream rendering (toon, json, the markdown/html/plaintext serializers, and the #185 footnote-only-paragraph surfacing). `read_file`'s suffix pass now enriches only `clean_text`, which the view deliberately keeps marker-free for core consumers (edit matching, signature-cluster detection). Every field now carries exactly one marker per reference.

**Peer-review hardening (Codex probe finding):** `collectFootnoteMarkerSuffix` collected every descendant `w:footnoteReference` indiscriminately, while the view skips references inside field code (between `fldChar begin` and `separate` — content Word never displays). That would have left `clean_text` claiming a footnote that `text` correctly omits. The suffix collector now applies the same `fldChar` state machine, so all fields agree on which footnotes exist.

## Tests

- The two #185 regression assertions that deliberately tolerated the duplicate (`/^\[\^N\](?:\[\^N\])?$/`) are tightened to exactly-one.
- The NVCA SPA full-document walk now asserts no adjacent same-number marker pair appears anywhere in the view.
- New regression test: a field-code-contained reference surfaces no marker in any field; a visible reference renders exactly one, with document-order display numbering still counting the hidden slot.

Peer-reviewed via Codex (`REQUEST-CHANGES` → addressed in second commit) and agy (`APPROVE`), both with dynamic execution traces.

Full pre-submit gate green: build, lint:workspaces, test:run, check:spec-coverage, check:conformance-citations, check:conformance-doc.

Fixes: #382

Issue #382, opened 2026-06-10
read_file renders every footnote marker twice in text/tagged_text ([^N][^N])


stevenobiajulu:
## Problem

Two independent passes both inject the `[^N]` footnote marker into `read_file` output:

1. **View level** — `buildNodesForDocumentView` calls `injectFootnoteMarkers` (`packages/docx-core/src/primitives/document_view.ts`), inserting the marker into `tagged` at the reference's visible offset. This flows into both `tagged_text` and `text`.
2. **Tool level** — `read_file` enrichment calls `collectFootnoteMarkerSuffix` (`packages/docx-mcp/src/tools/read_file.ts`) and blindly appends the marker again to `clean_text`, `tagged_text`, and `text`.

Result: every footnote-anchoring paragraph renders `[^N][^N]` in `text`/`tagged_text` (`clean_text` gets a single marker, since the view-level pass never touches it).

## Evidence

On the NVCA SPA fixture (`tests/test_documents/nvca-regression/source.docx`), a full `read_file(format='json')` walk shows **all 107 reachable markers doubled** — e.g. node `_bk_d752cfc13c29` ends `…distributes.[^45] [^45]`. Confirmed pre-existing on main (both injection sites predate #185's fix; verified via `git show` archaeology and a live run during that PR's peer review).

## Expected

Exactly one `[^N]` per reference per field. Either the view-level injection or the tool-level suffix should win; the other should be removed or made idempotent. Note the two passes disagree on placement (view = at the reference's visible offset; tool = appended at paragraph end) — picking one is also a placement decision.

## Notes

- Discovered while fixing #185 (footnote-only paragraphs dropped from the document view). The #185 regression tests deliberately assert `/^\[\^N\](?:\[\^N\])?$/` so they tolerate the current duplicate and a future single-marker fix, but fail on zero or triple markers — tighten them to exactly-one when fixing this.
- `clean_text` vs `text` asymmetry (1 vs 2 markers) is itself a consistency smell worth resolving in the same change.

Ref: #185

PR #413, opened 2026-06-11
refactor: expose footnote refs as document-view node metadata — one derivation, not five fldChar walks

### LINKED ISSUE #393: refactor: expose footnote references as document-view node metadata — one derivation of "visible footnote refs", not five fldChar walks
## Problem

The package answers the question "which footnotes does this paragraph visibly reference, and with what display number?" in **multiple independent places**, and #382 was the direct cost of that: the view-level injection and `read_file`'s suffix pass were two such derivations, they disagreed, and every reachable marker rendered doubled. The #382 fix (#386) closed that instance correctly — but its peer-review hardening did so by adding a **fifth** hand-rolled copy of the `fldChar` begin/separate/end visible-content state machine, so the bug *class* is still open.

### Current copies of the visible-content field-state walk

1. `getParagraphRuns` — `packages/docx-core/src/primitives/text.ts:17` (the canonical one)
2. `getFootnoteMarkersForParagraph` — `packages/docx-core/src/primitives/document_view.ts:227`
3. Comment-anchor offset walk — `packages/docx-core/src/primitives/comments.ts:912`
4. `getParagraphRunVisibleLengths` — `packages/docx-mcp/src/tools/read_file.ts`
5. `collectFootnoteMarkerSuffix` — `packages/docx-mcp/src/tools/read_file.ts` (added by #386's field-code hardening)

The copies already diverge in small ways (`text.ts` maps `w:br`→`\n` / `w:tab`→`\t`; the offset-counting copies count each as 1 visible char). Tolerable per-copy — but "independent derivations of the same fact drifting apart" is exactly the #382 failure shape.

### Duplicated display-number derivation

The footnote display-number map is also derived twice:

- `buildFootnoteDisplayMap` — `packages/docx-core/src/primitives/document_view.ts` (sequential scan of `document.xml` refs, skipping reserved separator ids)
- `getFootnotes()` `displayNumber` via `buildDisplayNumberMap` — `packages/docx-core/src/primitives/footnotes.ts:191`

Peer review on #386 verified these are currently **identical** — but identical-enforced-by-nothing is the same trap. `read_file` consumes the second to render markers the view computed with the first.

## Recommended shape

**Expose footnote references as document-view node metadata** so consumers stop re-walking the DOM:

1. `buildNodesForDocumentView` already computes per-paragraph markers via `getFootnoteMarkersForParagraph`. Surface that result on the node, e.g. `DocumentViewNode.footnote_refs?: Array<{ id: number; display: number }>` (omit the key when empty, matching the `heading` convention).
2. `read_file`'s `clean_text` enrichment renders the suffix from `node.footnote_refs` — **delete** `collectFootnoteMarkerSuffix`, its `fldChar` state machine, the `paragraphElementsById` lookup in that pass, and the `getFootnotes()`-based `displayById` map. One derivation, one numbering authority.
3. The view-level `injectFootnoteMarkers` keeps consuming the same `getFootnoteMarkersForParagraph` output it does today — that function becomes the single computation, now with its result exposed rather than private.

This removes both duplications at once (state machine copy #5 *and* the second display-number map) and shrinks `read_file`. The field-code semantics locked by the #386 regression test ("a reference inside field code surfaces no marker in any field") hold automatically, because there is only one place deciding visibility.

### Optional follow-up (separate PR)

Consolidate the remaining walkers (copies 2–4) onto a shared visible-run iterator exported from `text.ts`, so the begin/separate/end logic exists once in the package. Worth doing, but independent of the footnote SSOT above — keep it out of scope here per one-fix-per-PR.

## Acceptance criteria

- [ ] `DocumentViewNode` carries `footnote_refs` metadata computed in exactly one place
- [ ] `collectFootnoteMarkerSuffix` and the `read_file`-local display-number map are deleted
- [ ] `grep -c fldCharType packages/docx-mcp/src/tools/read_file.ts` drops from 2 to ≤1 (only the comment-offset walk remains, pending the follow-up)
- [ ] #386's regression tests (exactly-one marker; field-code reference hidden in every field; NVCA full-walk no-doubling) pass unchanged
- [ ] No behavior change in `read_file` output for any existing fixture

Refs: #382, #386 (peer-review trade-off accepted knowingly: bounded remediation over API surface change in a bug-fix PR)

stevenobiajulu:
Fixes #393. Refs #382, #386.

## Why

The package answered "which footnotes does this paragraph visibly reference, and with what display number" in multiple independent places, and #382 was the direct cost: two derivations disagreed and every reachable marker rendered doubled. The #386 fix closed that instance but (knowingly, per its peer-review trade-off) added a fifth copy of the `fldChar` begin/separate/end state machine and a second display-number map. Identical today — enforced by nothing.

## What

- **`DocumentViewNode.footnote_refs?: Array<{ id, display }>`** — the result of `getFootnoteMarkersForParagraph` (which already drives the view's `[^N]` injection and the empty-paragraph skip gate) is now surfaced on the node, in document order, omitted when empty (matching the `heading` convention).
- **`read_file`** renders its `clean_text` marker suffix from `node.footnote_refs`. `collectFootnoteMarkerSuffix`, its `fldChar` state machine, and the `getFootnotes()`-based `displayById` map are deleted — one derivation, one numbering authority.
- The per-paragraph walk is hoisted to run once (was computed twice: skip gate + injection).

The #386 invariant — a reference inside field code surfaces no marker in **any** field — now holds by construction, because only one place decides visibility.

## Acceptance criteria (from #393)

- [x] `DocumentViewNode` carries `footnote_refs` computed in exactly one place
- [x] `collectFootnoteMarkerSuffix` and the `read_file`-local display-number map are deleted
- [x] `grep -c fldCharType packages/docx-mcp/src/tools/read_file.ts` → **1** (was 2; remainder is the comment-offset walk, pending the shared visible-run iterator follow-up)
- [x] #386's regression tests pass unchanged (exactly-one marker; field-code reference hidden in every field; NVCA full-walk no-doubling) — extended with `footnote_refs` assertions locking the new contract
- [x] No behavior change in `read_file` rendered output; JSON format additionally carries the new `footnote_refs` metadata key on footnote-bearing nodes (the exposure #393 asks for)

## Out of scope

Consolidating the remaining walkers (copies 2–4) onto a shared visible-run iterator in `text.ts` — #393's optional follow-up, kept separate per one-fix-per-PR.

## Verification

`npm run build && npm run lint:workspaces && npm run test:run && npm run check:spec-coverage && npm run check:conformance-citations && npm run check:conformance-doc` — all exit 0 (gated on explicit exit codes).

### DIFF
--- a/packages/docx-core/src/primitives/document_view-types.ts
+++ b/packages/docx-core/src/primitives/document_view-types.ts
@@ -126,6 +126,18 @@ export type DocumentViewNode = {
   paragraph_indents_pt: { left: number; first_line: number };
   numbering: { num_id: string | null; ilvl: number | null; is_auto_numbered: boolean };
   heading?: HeadingValue;
+  /**
+   * Footnote references this paragraph visibly anchors, in document order, each
+   * with the display number assigned by the view's numbering authority (the
+   * sequential scan of `document.xml` references, reserved separator IDs
+   * skipped). References inside field code (between `w:fldChar` begin and
+   * separate) are hidden text and excluded. This is the single derivation of
+   * "which footnotes does this paragraph reference" — consumers render from it
+   * instead of re-walking the paragraph DOM. Omitted when empty, matching the
+   * `heading` convention.
+   * @see #393
+   */
+  footnote_refs?: Array<{ id: number; display: number }>;
   header_formatting: HeaderFormatting | null;
   body_run_formatting: RunFormatting | null;
   table_context?: TableContext;
--- a/packages/docx-core/src/primitives/document_view.ts
+++ b/packages/docx-core/src/primitives/document_view.ts
@@ -217,9 +217,18 @@ function buildFootnoteDisplayMap(documentXml: Document, footnotesXml: Document |
 }
 
 /**
- * Compute footnote marker insertion points for a paragraph.
- * Returns an array of { offset, marker } sorted by offset descending
- * for safe right-to-left insertion into the text string.
+ * A footnote reference a paragraph visibly anchors: the referenced footnote's
+ * numeric ID, its display number, and the reference's visible-character offset
+ * within the paragraph text.
+ */
+type ParagraphFootnoteRef = { offset: number; id: number; display: number };
+
+/**
+ * Compute the footnote references a paragraph visibly anchors, in document
+ * order. This is the single derivation of "which footnotes does this paragraph
+ * reference, and with what display number" — the view injects [^N] markers
+ * from it AND exposes it as DocumentViewNode.footnote_refs so consumers
+ * (read_file's clean_text suffix) never re-walk the DOM. @see #393
  *
  * Self-contained: only inspects the paragraph DOM for w:footnoteReference
  * elements. Does NOT modify getParagraphRuns or getParagraphText.
@@ -227,13 +236,13 @@ function buildFootnoteDisplayMap(documentXml: Document, footnotesXml: Document |
 function getFootnoteMarkersForParagraph(
   p: Element,
   displayMap: Map<number, number>,
-): Array<{ offset: number; marker: string }> {
+): ParagraphFootnoteRef[] {
   if (displayMap.size === 0) return [];
 
   // Walk through direct children (and hyperlink children) to find w:r elements
   // and their visible text, tracking position. When we find a footnoteReference,
   // record its position.
-  const markers: Array<{ offset: number; marker: string }> = [];
+  const markers: ParagraphFootnoteRef[] = [];
   let visibleOffset = 0;
 
   // We need to iterate runs in paragraph order. Use the same approach as getParagraphRuns
@@ -280,7 +289,8 @@ function getFootnoteMarkersForParagraph(
       if (displayNum != null) {
         markers.push({
           offset: visibleOffset + runVisibleLen,
-          marker: `[^${displayNum}]`,
+          id: footnoteId,
+          display: displayNum,
         });
       }
     }
@@ -288,8 +298,6 @@ function getFootnoteMarkersForParagraph(
     visibleOffset += runVisibleLen;
   }
 
-  // Sort descending by offset for safe right-to-left insertion
-  markers.sort((a, b) => b.offset - a.offset);
   return markers;
 }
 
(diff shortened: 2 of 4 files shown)


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `464df4cfb92bc3a04e874b0998184be92e83b874`
Built from: `main` @ `464df4c`
Smoke: `npm run build && npm run test:run` + real-document e2e through the shipped MCP stdio server (`packages/docx-mcp/dist/cli.js`, JSON-RPC `tools/call`)

**Steps**
- ✅ build (clean, from merged main HEAD)
- ✅ tests (2,552 passed across workspaces; 2 expected-fail, 47 skipped)
- ✅ `read_file` + `get_footnotes` driven over MCP stdio against a real NVCA form

**Real-world fixtures**
- NVCA form (`tests/test_documents/nvca-regression/source.docx`, 368 paragraphs, 109 footnotes — copied to /tmp, read via the MCP server entry point)
  - 84 nodes expose `footnote_refs`; 108 references total
  - `clean_text` suffix matches the `footnote_refs` derivation on 84/84 nodes
  - Every `[^N]` marker renders exactly once in `text`/`tagged_text`/`clean_text` (108/108 refs — the #382 doubling class)
  - `footnote_refs` display numbers agree with `get_footnotes` for 108/108 refs (the two formerly-independent numbering authorities); displays sequential 1..108
  - No stray markers on nodes without `footnote_refs`
  - The one unanchored footnote (id 1) is a genuine orphan in the source document — defined in `footnotes.xml`, never referenced in `document.xml` (verified against the raw XML, independent of the code under test)

Visual evidence: skipped — change is structurally invisible on a rendered page (tool-output metadata; the document itself is never mutated).
Log: `/tmp/automerge-smoke-413-20260611-004418.log` (local). Artifacts: `/tmp/automerge-smoke-413-20260611-004418/` (smoke script + fixture copy).

**Cleanup**
- ✅ remote branch already deleted (repo auto-delete on merge; confirmed via prune)
- ✅ removed worktree `.worktrees/issue-393`
- ✅ deleted local branch `393-footnote-refs-node-metadata-20260611`

PR #377, opened 2026-06-10
fix(docx-core): preserve w:hyperlink wrappers in rebuild reconstruction

stevenobiajulu:
Fixes #368.

## What

Rebuild reconstruction dropped `<w:hyperlink>` wrappers entirely (link + `r:id` lost) and the cross-run punctuation merge absorbed the sentence period after a URL into the underlined link run. Three changes, one per layer:

- **Atomizer** — cross-run text/punctuation merges stop at `w:hyperlink` boundaries (`nearestHyperlinkAncestor`). This is the underline-bleed root cause: the merged atom kept only the link run's ancestry, so the "." inherited underline + hyperlink membership.
- **Reconstructor** — run groups are split into hyperlink-pure segments; adjacent segments with the same attribute fingerprint share one re-emitted `<w:hyperlink>` wrapper. Revision wrappers nest **inside** the hyperlink: `CT_Hyperlink → EG_PContent → EG_ContentRunContent → EG_RunLevelElts` admits `w:ins`/`w:del`, while `CT_RunTrackChange` does not admit `w:hyperlink` (verified against the vendored `wml.xsd`). Applies to the whole-paragraph insert/delete branches too. Wrapper attribution prefers the **original** tree's element (deleted atoms directly, equal atoms via `comparisonUnitAtomBefore`) because rebuild output clones the original archive — only original-tree `r:id` values resolve against the shipped rels.
- **Acceptor** — accept/reject prune `w:hyperlink` elements emptied by change resolution, matching Word (no contentless `<w:hyperlink r:id=".."/>` husks).

Hyperlink-free paragraphs take the unchanged legacy emission path (byte-identical output).

## Known boundaries (deliberate, pinned by tests)

- A **purely-inserted hyperlink with `r:id`** stays unwrapped (pre-fix behavior): its relationship exists only in the revised package, and a dangling `r:id` reads as a corrupt package in Word. Anchor-only inserted hyperlinks are wrapped. Relationship merging is tracked in #376.
- A **retargeted link** (`r:id` change with equal text) emits the equal text unwrapped rather than pinned to the stale original target. Caught in peer review (Codex executed probe — the first iteration attributed equal text to the old `r:id`, silently mislinking accepted content). The faithful delete-old-link + insert-new-link representation is #376.

## Evidence

Real-fixture repro from the issue (Common Paper Mutual NDA, `Confidential Information`→`Proprietary Information` + `mutual`→`reciprocal`):

| view | `<w:hyperlink>` | `rId7` refs |
[...]

PR #428, opened 2026-06-11
feat(docx-core): from-scratch generation phase 7 — compatibility sign-off and repositioning

stevenobiajulu:
## Summary

Final phase of the `add-docx-generation` OpenSpec change (PR 7 of 7) — **closes #280**.

- **Public API**: the generation surface (`generateDocx`, the full `DocumentSpec` type family, `coverTermsTable`/`signatureBlock` recipes, `checkGeneratedPackage`) is now exported from the `@usejunior/docx-core` package root.
- **Compatibility sign-off** (scenario `[SDX-GEN-092]`): LibreOffice results recorded in the manual matrix for all six artifact classes — identity re-save reloads via `DocxDocument` with content intact, PDF conversion non-empty, no dialogs (headless LO on macOS, 2026-06-11). Word for Mac / Pages / Google Docs cells remain open for manual observation — they can't be automated honestly from this environment. A new test discovers the artifact set from the sources that write it (`writeIntegrationArtifact` calls) and fails if the matrix ever drifts; scenarios 090/091 were already covered by the automated probes.
- **Repositioning sweep**: from-scratch generation is no longer a stated non-goal. README gains a "From-Scratch Generation" section + updated FAQ answer; site FAQ, `docx-editing` skill (description + body), AGENTS.md skill listing, LLM-gate system prompt, and the conformance-registry Non-Goals prose all reflect the shipped capability. The MCP server still exposes editing only — generation is a library API (a `generate_document` tool is a future follow-up).
- **Coverage enforcement**: `check:spec-coverage-generation` flips from `--report-only` to `--strict` on its explicit root invocation — **36/36 scenarios enforced in CI** from this PR on.
- **Follow-ups filed**: #425 (translated READMEs carry the old positioning), #426 (odf-core ODT generation parity decision).
- Post-deploy: an archive PR will run `openspec archive add-docx-generation --yes`.

Fixes #280

## Test plan

- [x] `npm run build && npm run lint:workspaces && npm run test:run` (2578 passed) + `npx madge --circular` clean
- [x] `npm run check:spec-coverage` — generation validator strict: `PASS add-docx-generation: 36 scenarios covered`
- [x] `npm run check:conformance-citations` (62 entries) / `check:conformance-doc` green at commit
- [x] `npm run check:readme-sync && npm run check:site` (README + site FAQ touched)
- [x] `openspec validate add-docx-generation --strict`
[...]

Issue #409, opened 2026-06-11
atomizer compare emits delete+insert pairs for byte-identical empty paragraphs when neighboring paragraphs are restructured


stevenobiajulu:
## Problem

`docx-comparison` (atomizer engine, rebuild mode) produced tracked changes on paragraphs that are **byte-identical** between the two inputs.

Comparing a real Common Paper Mutual NDA against a copy that differs in exactly one paragraph (`{mnda_term}` → "two (2) years"), the redline output contained three change regions:

| output paragraphs | content | legitimate? |
|---|---|---|
| p12 | `w:del` `{mnda_term}` + `w:ins` `two (2) years` | ✅ the real edit |
| p3 (DEL) + p4 (INS) | empty paragraph deleted, empty paragraph inserted | ❌ phantom |
| p27 (DEL) + p28 (INS) | empty paragraph deleted, empty paragraph inserted | ❌ phantom |

The corresponding empty paragraphs in the two **inputs** are byte-for-byte identical (verified by serializing and diffing each one). What does differ is their *neighbors*: the revised file had been saved by `safedocx edit`, which strips `w:proofErr` and merges runs in nearby paragraphs without changing their text (see #408). The pipeline debug line shows the asymmetry the alignment had to absorb: `created 95 atoms` (original) vs `77 atoms` (revised), both merging to 75.

So: word-level atoms absorb the run-structure churn for text-bearing paragraphs (text equal → atoms match), but next to two of the restructured paragraphs the LCS alignment drops an identical empty paragraph from the matched sequence and re-emits it as delete+insert.

## Why it matters

- The rendered redline shows change bars on locations nobody edited — for a legal-document redline, phantom revisions directly undermine trust in the output (a reviewing lawyer must investigate each one).
- It inflates `stats` (`insertions`/`deletions` each +2 here; see #411).
- Accept-all on the output churns paragraph identity (bookmarks/anchors on the re-inserted empty paragraph are recreated, not preserved).

## Repro

```bash
cp open-agreements/content/templates/common-paper-mutual-nda/template.docx original.docx
SAFE_DOCX_AI_AUTHOR="" safedocx edit original.docx \
  --replace _bk_0922ea572bbe "{mnda_term}" "two (2) years" -o revised.docx
docx-comparison original.docx revised.docx redline.docx
# inspect word/document.xml: 3 ins + 3 del regions; the p3/p4 and p27/p28 pairs are empty paragraphs
```

## Expected

[...]

PR #429, opened 2026-06-11
fix(docx-core): write hyperlinkKey NUL separator as an escape; guard source against raw control bytes

### LINKED ISSUE #427: raw NUL byte in documentReconstructor.ts makes grep treat the file as binary; guard against control bytes in source
## Problem

`packages/docx-core/src/baselines/atomizer/documentReconstructor.ts` line 937 (`hyperlinkKey()`) contains a **raw NUL byte (0x00)** inside a string literal:

```ts
return parts.sort().join('<raw 0x00 byte here>');
```

The NUL is semantically intentional — a "cannot collide with attribute values" fingerprint separator, introduced in #377 — and the code behaves correctly at runtime. The defect is that it is embedded as a raw byte instead of the six-character escape `'\u0000'` (identical runtime string).

## Why it matters

- **Silent grep false negatives across the whole file.** BSD grep treats any file containing NUL as binary and returns *empty with no warning*: `grep -rn hyperlinkKey packages/docx-core/src/baselines/atomizer/` finds zero matches even though the function is defined and called in that file. In an agent-operated repo this is the dangerous failure mode: an agent searches, gets nothing, and concludes a symbol is unused or a pattern does not exist — the precondition for duplicated helpers or unsafe deletions. ripgrep likewise classifies the file as binary.
- **Invisible and review-proof.** The byte renders as nothing in editors and survived #377 review unnoticed. `file` reports the .ts source as `data`.
- **Fragile luck elsewhere:** git still diffs the file as text only because the byte sits past the first 8 KB its binary heuristic sniffs.

## Fix

1. Replace the raw byte with the `'\u0000'` escape — behavior-identical.
2. Add a cheap source-hygiene guard so the *class* of bug is dead, not just this instance: a check (wired into the existing workspace-lint / check-script chain) that rejects raw control bytes (anything below 0x20 except tab, LF, CR) in tracked source files. Without the guard, the next invisible control byte just resets the clock.

Found while investigating compare stats for #404's post-merge smoke — greps against this file kept returning empty.

Ref: #377

stevenobiajulu:
## Problem

`hyperlinkKey()` in `packages/docx-core/src/baselines/atomizer/documentReconstructor.ts` joined attribute fingerprints with a **raw NUL byte (0x00)** embedded directly in the string literal (shipped invisibly in #377). The separator semantics are correct — NUL can't appear in XML attribute values — but the raw byte made grep-family tools classify the entire 1,776-line file as binary and **silently return zero matches**: `grep -rn hyperlinkKey` over the atomizer directory found nothing even though the function is defined and called there. `file` reported the `.ts` as `data`. For agents operating on this repo, a silent search false-negative reads as "symbol unused / pattern absent" — the precondition for duplicated helpers or unsafe deletions.

## Fix

- Write the separator as the six-character Unicode escape — behavior-identical (the escape parses to the same one-character string at JS parse time), and verified that `tsc` carries the escape form through to `dist`, so the compiled output is clean too.
- Add `scripts/check_control_bytes.mjs` + `check:control-bytes` (wired into the `workspace-lint` CI job): scans tracked JS/TS sources for raw bytes below 0x20 (other than tab/LF/CR) and fails with `file:line:col` plus the escape to use instead. Because the failure mode is invisible in editors and review, the one-character fix alone would just reset the clock until the next control byte lands.

## Verification

- Guard **red** against the pre-fix tree: exactly one finding repo-wide (`documentReconstructor.ts:937:29 — byte 0x00`), i.e. no false positives across 584 source files; **green** after the fix.
- `grep -c hyperlinkKey` on the file: 0 matches before → 5 after; `file` reports UTF-8 text again.
- docx-core suite: 1,436 passed / 2 expected-fail / 1 skipped. `lint:workspaces`, `check:cycles` clean.

Fixes: #427
Ref: #377

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,6 +71,8 @@ jobs:
         run: npm run lint:workspaces
       - name: Check for import cycles
         run: npm run check:cycles
+      - name: Check for raw control bytes in source
+        run: npm run check:control-bytes
       - name: Check release isolation (ODF decoupling guard)
         run: npm run check:release-isolation
       - name: Test release-isolation guard
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "lint:workspaces": "npm run lint --workspaces --if-present && npm run lint:allure-imports",
     "lint:allure-imports": "eslint \"packages/**/*.test.ts\" --ignore-pattern \"packages/google-docs-core/**\" --ignore-pattern \"packages/odf-core/**\"",
     "check:cycles": "madge --circular --extensions ts packages/*/src",
+    "check:control-bytes": "node scripts/check_control_bytes.mjs",
     "test": "npm run test --workspaces --if-present",
     "test:run": "npm run test:run --workspaces --if-present",
     "test:baseline": "npm run test:baseline -w @usejunior/docx-core",
--- a/packages/docx-core/src/baselines/atomizer/documentReconstructor.ts
+++ b/packages/docx-core/src/baselines/atomizer/documentReconstructor.ts
@@ -934,7 +934,7 @@ function hyperlinkKey(el: Element): string {
     if (attr.name.startsWith('xmlns')) continue;
     parts.push(`${attr.name}=${attr.value}`);
   }
-  return parts.sort().join(' ');
+  return parts.sort().join('\u0000');
 }
 
 /**
--- a/scripts/check_control_bytes.mjs
+++ b/scripts/check_control_bytes.mjs
@@ -1,0 +1,80 @@
+#!/usr/bin/env node
+// Source-hygiene guard: no raw control bytes in tracked JS/TS source (#427).
+//
+// A raw control byte (anything below 0x20 other than tab/LF/CR) in a source
+// file makes grep-family tools classify the whole file as binary and return
+// EMPTY results with no warning — BSD grep and ripgrep both do this. In an
+// agent-operated repo that failure mode is dangerous: a search that silently
+// returns nothing reads as "symbol unused / pattern absent", which is the
+// precondition for duplicated helpers or unsafe deletions. The byte itself is
+// invisible in editors, so review never catches it (#377 shipped one).
+//
+// Legitimate uses of control characters in code (e.g. a NUL fingerprint
+// separator) belong in escape form — '\u0000' produces the identical runtime
+// string and keeps the file text.
+//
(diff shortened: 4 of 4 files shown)


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `c894ca8` (squash)
Built from: `main` @ `c894ca8` (HEAD; #428 landed in between)
Smoke: clean workspace build → full test suite → guard + grep verification → fresh `npm pack` consumer driving `docx-comparison` through a hyperlink-bearing edit on a real DOCX

**Steps**
- ✅ build (`npm run build`, all workspaces)
- ✅ tests (`npm run test:run` — 2,589 passed, 2 expected-fail, 36 skipped across workspaces)
- ✅ `npm run check:control-bytes` live on main: 586 source files scanned, 0 findings
- ✅ the fixed defect, verified at the tool level: `file` reports `documentReconstructor.ts` as UTF-8 text (was `data`), `grep -c hyperlinkKey` finds 5 matches (was 0/binary), compiled `dist` JS contains no raw NUL — tsc carries the escape form through

**Real-world fixtures**
- `common-paper-mutual-nda` (local template at `content/templates/common-paper-mutual-nda/template.docx`)
  - Edited the **hyperlink-bearing intro paragraph** via `safedocx edit` (untracked): "control over conflicts" → "govern in the event of any conflict" — chosen specifically because `hyperlinkKey()` (the function that carried the NUL) matches `w:hyperlink` wrappers across trees during rebuild reconstruction
  - Packed `usejunior-docx-core-0.11.0.tgz` from main, installed in a fresh consumer, ran `./node_modules/.bin/docx-comparison original.docx revised.docx` → exit 0, stats `{"insertions":12,"deletions":4,"modifications":2}`
  - Redline output preserves the hyperlink: `w:hyperlink r:id=rId7` with intact text `commonpaper.com/standards/mutual-nda/1.0` inside the tracked-changed paragraph
  - (One empty-paragraph del+ins pair present in the output — known artifact, tracked as #409, unrelated to this PR)

Visual evidence: skipped (escape form produces the identical runtime string — change is structurally invisible on a rendered page)
Log: `/tmp/automerge-smoke-429-20260611-031102.log` (local).

**Cleanup**
- ✅ remote branch already deleted (auto-delete on merge)
- ✅ no worktree existed for this branch
- ✅ deleted local branch (force, post-squash — PR verified MERGED)

Issue #359, opened 2026-06-10
Pre-tracked insertion content that collides with the other input's plain text loses provenance — reject round-trip diverges (inline w:ins revised, paragraph-insert either side)


stevenobiajulu:
## Context

Sibling of #358, surfaced by the same #347 relaxation of the lean-spec-bridge arbitrary; pinned in the bridge test and filed here for its own future PR. Where #358 is about the original's inline `w:ins` being flattened unconditionally, this class needs a **provenance collision**: content that one input claims as a pre-tracked insertion while the other input carries it (or its text) as plain settled content.

## Bug

When pre-tracked-insertion content on one side textually matches plain content on the other side, the comparison resolves the match by keeping exactly one lineage's provenance, so the global reject projection diverges from `rejectAllChanges(original)`:

1. **Inline `w:ins` revised, ins text matches a plain original word** (the exact #339 flake shape): the combined keeps the revised's `w:ins(Lean Bridge)` around the matched text; reject drops it although the original carried it plain. The inplace candidate correctly fails the post-#347 `rejectText` check and the engine falls back — and here the **rebuild output satisfies the law**, so the system-level outcome is correct (the cost is only the lost inplace fidelity + the bridge property's fallback-is-falsification assertion).
2. **Paragraph-insert revised whose new paragraph text equals a plain original paragraph**: rebuild fallback output **violates** the law — reject(combined) `"Alpha text."` vs reject(original) `"Alpha text.\nAdded para."`.
3. **Paragraph-insert original whose new paragraph text appears plain in the revised**: rebuild fallback output **violates** the law — reject(combined) `"Alpha text.\nAdded para."` vs reject(original) `"Alpha text."`.

Probes (run on the #347 branch):

| case | mode | final-output law |
|---|---|---|
| inline-ins revised `'I'` into `'6.'` vs original `['!','!','I']` | rebuild | satisfied (rebuild rescues) |
| paragraph-insert revised `'Added para.'` vs original `['Alpha text.','Added para.']` | rebuild | **violated** |
| paragraph-insert original `'Added para.'` vs revised `['Alpha text.','Added para.']` | rebuild | **violated** |
| paragraph-insert (no collision), either side | inplace | satisfied |
| both inputs claim the same insertion (two-author stack) | inplace | satisfied |

## Semantics note

[...]

Issue #358, opened 2026-06-10
Comparison flattens the original's inline pre-tracked w:ins provenance — reject(combined) keeps text reject(original) drops (violates INV-RT-001 projection law in both modes)


stevenobiajulu:
## Context

#347 corrected the reject/accept round-trip oracle to apples-to-apples projection equality (accept-all(combined) == accept-all(revised), reject-all(combined) == reject-all(original)) and relaxed the lean-spec-bridge arbitrary to pre-tracked-insertion originals. The relaxation surfaced this engine bug class; per the aggressive-discovery/bounded-remediation discipline it is pinned in the bridge test (narrow issue-linked exclusion + characterization test) and filed here as its own future PR.

## Bug

When the **original** input carries an inline (run-level) pre-tracked `<w:ins>`, the comparison flattens that provenance: the inserted text enters the combined output as plain matched content or plain `w:delText`, with no surviving `w:ins` wrapper from the original lineage. Globally rejecting the combined output therefore **keeps/restores** text that `rejectAllChanges(original)` **drops** — the corrected INV-RT-001 reject projection is violated.

The inplace safety check (post-#347 baselines) correctly detects this on the inplace candidate and falls back to rebuild — but the **rebuild output violates the law too** and ships unchecked (#226 is the related gap: rebuild output bypasses `evaluateRoundTripSafety`).

This is **unconditional** over the revised side — every probed variant fails, including a revised that is textually identical to accept(original) with zero comparison-visible changes.

## Minimal repro (probed on the #347 branch)

Original: paragraph `Alpha` + pre-tracked insertion `' beta'` (author "Lean Bridge") → raw text `Alpha beta`, reject-projection `Alpha`.

| revised | mode | reject(combined) | reject(original) |
|---|---|---|---|
| `['Alpha beta tail']` | rebuild | `Alpha beta` | `Alpha` |
| `['Alpha']` | rebuild | `Alpha beta` | `Alpha` |
| `['Alpha beta']` (== accept(original)) | rebuild | `Alpha beta` | `Alpha` |
| `['Gamma delta']` | rebuild | `Alpha beta` | `Alpha` |

[...]

PR #464, opened 2026-06-12
fix(docx-core): merge paragraph-mark revisions into the following paragraph

### LINKED ISSUE #431: Accept/reject paragraph-mark revisions drops preceding paragraph content
The `docx-platform-tests` matrix now surfaces a semantic failure in the paragraph-mark tracked-change scenarios:

- `acceptDeletedParagraphMarkMergesParagraphs`
- `rejectInsertedParagraphMarkMergesParagraphs`

Observed with `@usejunior/docx-core` `0.11.0+git.fef2ba5f1c36`:

- both scenarios leave the output text projection as `"second half"`
- expected projection is `"First half second half"` at offset 0
- the structural paragraph-count assertions pass, so the issue appears to be that the first paragraph's contents are dropped while merging into the following paragraph

Why this matters: ECMA-376 Part 1 §§ 17.13.5.15 (`del` / Deleted Paragraph) and 17.13.5.20 (`ins` / Inserted Paragraph) make the paragraph mark the revision target; paragraph contents are not automatically marked as deleted. LibreOffice passes these same paragraph-mark rows in the published comparison.

Reproduction path from `open-agreements/docx-platform-tests`:

```bash
cd runner
npm ci
npm run check-fixtures
npm run suite
```

The failing scenarios live under:

- `scenarios/tracked-changes/acceptDeletedParagraphMarkMergesParagraphs/`
- `scenarios/tracked-changes/rejectInsertedParagraphMarkMergesParagraphs/`

stevenobiajulu:
## Summary

Fixes #431 — accepting a deleted paragraph mark (or rejecting an inserted one) dropped the paragraph's untracked content instead of merging it into the following paragraph.

ECMA-376 Part 1 § 17.13.5.15 (`del` / Deleted Paragraph) and § 17.13.5.20 (`ins` / Inserted Paragraph) make the paragraph **mark** the revision target; the paragraph's contents are not implicitly part of the revision. Both engine layers treated the mark as "remove the whole paragraph", which silently lost any content not separately wrapped in run-level `w:del`/`w:ins` — the data loss the `docx-platform-tests` matrix surfaced (`acceptDeletedParagraphMarkMergesParagraphs`, `rejectInsertedParagraphMarkMergesParagraphs` left `"second half"` instead of `"First half second half"`).

## What changed

- **Mark resolution is now a merge, not a drop** (`resolveParagraphMarkRevision` in both layers — DOM primitives `accept_changes.ts`/`reject_changes.ts` and AST `trackChangesAcceptorAst.ts`): the paragraph's surviving non-`pPr` children move to the head of the following sibling paragraph; the survivor keeps its own `w:pPr` (formatting follows the surviving paragraph mark). Consecutive marked paragraphs cascade in document order.
- **No-merge-target fallback** (last block, or next block is a table): keep the paragraph if it still has content (no data loss); remove it only when emptied — which preserves the comparison round-trip shape, since `wrapParagraphAsDeleted`/`AsInserted` carry all content in run-level wrappers.
- The forward scan for the merge target skips block-level annotation markers (`bookmarkStart/End`, `commentRange*`, `perm*`, `proofErr`, `move*Range*`).
- **Registry**: added targeted entries `ECMA-PART1-17-13-5-15` / `-20`; `CONFORMANCE.md` + README counters regenerated. New code carries `@conformance` tags.

## Tests

- New `#431` describe block in `trackChangesAcceptorAst.test.ts` (7 cases, both engine paths each): the two platform-test scenarios, surviving-`pPr` ownership, cascade, trailing kept/removed, table no-target. Tagged with `.conformance(...)` for § 17.13.5.15 / § 17.13.5.20.
- Updated the G5 characterization test that previously **pinned the over-delete** (`PPR-DEL-marked paragraph: both paths DROP it` asserted the untracked content disappears) to assert the merge.
- `cross-implementation-suite` against `docx-platform-tests` now passes **every** scenario, including the two #431 rows (previously failing).

## lean-spec-bridge fallout (pinned, not fixed here)

The merge rule made a known engine bug newly observable: when the comparison deletes a pre-tracked **inserted paragraph** (paragraph-insert original), the inplace emitter stacks the paragraph marks correctly (`PPR-DEL(Comparison)` + `PPR-INS(original)`) but flattens the content to bare `w:del(Comparison)` — the original-author `w:ins` wrapper is lost (#358 bug class). While mark resolution dropped the whole paragraph this was invisible; under the merge rule reject(combined) restores that content and diverges from reject(original), so the engine (correctly) falls back to rebuild.

Per the bounded-remediation discipline:
- `trackedOriginalScenarioArb` now excludes paragraph-insert originals (mirroring the existing inline-`w:ins` exclusion), with the rationale in the comment block;
- the shape is pinned verbatim as a new case in the `INV-RT-001 pinned engine bugs` characterization test;
- the shifted #359 case-3 pin was updated — the rebuild output's reject projection now **equals** reject(original) for the colliding shape (the law accidentally holds; provenance is still lost).

## Verification

- `npm run build && npm run lint:workspaces && npm run test:run && npm run check:spec-coverage && npm run check:conformance-citations && npm run check:conformance-doc` — all green.
- `DOCX_PLATFORM_TESTS_DIR=… vitest run cross-implementation-suite` — every scenario passes.
- Peer-reviewed (Codex + agy, both dynamic — each ran the test suites). All findings applied:
  - **Codex MAJOR**: merge-target scan skip-list completed to the full `EG_RangeMarkupElements` schema group (`customXml*Range*` markers were missing); sets renamed to `RANGE_MARKUP_BLOCK_SIBLING_*`.
  - **Codex MAJOR**: bookmark preservation no longer pre-relocates direct-child bookmarks of a *merging* paragraph (the merge carries them in document order; the early move reordered boundaries relative to surviving content). Relocation still rescues markers of paragraphs with no merge target.
  - **agy MAJOR**: the emptied-paragraph removal fallback is now gated on `canSafelyRemoveEmptyParagraph` — keep the empty `<w:p>` when removal would leave a parent with no block element, a `w:tbl` as the last block, or two adjacent tables.
  - **Codex MINOR**: reject-side test mirrors added (cascade, `pPr` survival, trailing keep/remove, table no-target) plus regressions for all three findings above.
  - **Codex NIT**: `accept_changes.ts` no longer claims § 17.13.5.20 (inserted marks are reject-side).

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ safe-docx targets a defined subset of **ECMA-376 5th edition**. The full surface
 (targeted sections, Non-Goals, and verification status) lives at
 [spec-compliance/CONFORMANCE.md](spec-compliance/CONFORMANCE.md).
 
-- **63** sections claimed
+- **65** sections claimed
 - **5** sections explicitly out-of-scope (Non-Goals)
 - **0** known gaps under `@conformance-gap`
 - Vendored normative schemas: `spec-compliance/ecma-376/schemas/`
--- a/packages/docx-core/src/baselines/atomizer/trackChangesAcceptorAst.test.ts
+++ b/packages/docx-core/src/baselines/atomizer/trackChangesAcceptorAst.test.ts
@@ -12,6 +12,7 @@ import { parseDocumentXml } from './xmlToWmlElement.js';
 import {
   findAllByTagName,
   acceptChanges as acceptChangesPrimitive,
+  rejectChanges as rejectChangesPrimitive,
   parseXml,
   serializeXml,
 } from '../../primitives/index.js';
@@ -856,14 +857,16 @@ describe('trackChangesAcceptorAst', () => {
   });
 });
 
-// ── G5 regression: Accept-All paragraph removal is purely mark-based (both accept paths) ──
+// ── G5 regression: Accept-All paragraph-mark handling is purely mark-based (both accept paths) ──
 //
 // Closes G5 — the accept-side mirror of #337's reject fix. Both accept entry points — the
 // baseline-atomizer `acceptAllChanges` (string→string) and the primitive `acceptChanges`
-// (Document, mutated in place) — must drop a paragraph IFF its paragraph MARK is PPR-DEL
+// (Document, mutated in place) — must resolve a paragraph IFF its paragraph MARK is PPR-DEL
 // (<w:pPr><w:rPr><w:del/></w:rPr>), never via a content-based heuristic. A run-level deletion
 // (or moveFrom) under an UNTRACKED mark means text removed from a pre-existing paragraph, which
-// Word/LibreOffice keep (empty) on accept. The two paths must agree on every case.
+// Word/LibreOffice keep (empty) on accept. Resolving a PPR-DEL mark merges the paragraph into
+// the following one (#431) — only the break is deleted, not the contents. The two paths must
+// agree on every case.
 describe('Accept-All paragraph removal is mark-based (G5, both accept paths agree)', () => {
   const wrapBody = (inner: string): string =>
     `<?xml version="1.0"?>` +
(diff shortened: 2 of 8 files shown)


stevenobiajulu:
Adding `llm-gate/override` with justification.

The single WARN from the LLM quality gate is the **lean-spec-bridge fallout that is pinned, not introduced, by this PR** — documented in the PR description under "lean-spec-bridge fallout (pinned, not fixed here)" and in the commit body.

Specifically: the merge rule made a pre-existing engine bug (the #358-class provenance flattening — the inplace emitter stacks paragraph marks correctly but flattens pre-tracked *inserted-paragraph* content to bare `w:del(Comparison)`, losing the original-author `w:ins` wrapper) newly *observable* on paragraph-insert originals. It does **not** introduce new data loss — while mark resolution dropped the whole paragraph the divergence was invisible; under the correct merge rule reject(combined) restores content reject(original) discards, so the engine (correctly) falls back to rebuild.

Per the bounded-remediation discipline this is characterized, not hidden:
- `trackedOriginalScenarioArb` excludes paragraph-insert originals (mirroring the existing inline-`w:ins` exclusion), with the rationale in-comment;
- the shape is pinned verbatim as a new case in the `INV-RT-001 pinned engine bugs` characterization test;
- the shifted #359 case-3 pin was updated accordingly.

All required checks are green (`workspace-test (20)`/`(22)`, `workspace-lint`, `spec-coverage`, `dependency-review`, conventional title), and `cross-implementation-suite` against docx-platform-tests passes every scenario including the two #431/#481 rows. Overriding the gate for this documented WARN.

Issue #121, opened 2026-05-03
Add validator for AI-emitted tracked-change OOXML


stevenobiajulu:
## Context

The current validator in `packages/docx-core/src/primitives/validate_document.ts` is narrow. Once tracked changes are emitted at write time (#120), the validator becomes the truth layer that prevents malformed revisions from landing.

## Proposal

Extend the validator to assert, for every AI-emitted revision:

**Per-revision well-formedness.**
- Unique `w:id` within the document.
- Valid `w:author` matching the configured AI actor (#119).
- ISO `w:date`.

**Range markers balanced.**
- `w:moveFromRangeStart`/`End`
- `w:commentRangeStart`/`End`
- `w:permStart`/`End`

**Field structure preserved.**
- Extend / reuse `validateFieldStructure()` in `packages/docx-core/src/baselines/atomizer/pipeline.ts`.
- `w:fldChar` begin/separate/end balanced.
- `w:instrText` only inside field sequences.

**Element-type rules.**
- `w:t` inside `w:ins` body.
- `w:delText` inside `w:del`.
- `w:delInstrText` for deleted instructions.

**Coverage extends to every revision type listed in the surface (#119).**
- `cellIns` / `cellDel`
- `tblGridChange`
- `sectPrChange`
- `numberingChange`
- `pPrChange` / `rPrChange`

**Package-invariants for AI-exposed operations that touch side parts.**
- Relationship integrity.
- Content-type registration.
- Target part exists.

**Validation failure is a hard error, not a silent fallback.**

## Acceptance criteria

- Unit tests cover one positive and one negative case per revision type in the surface (#119).
- Validator runs after every AI write in `@usejunior/docx-core` and `@usejunior/docx-mcp`.
- No silent fallbacks: any failure surfaces as an MCP-level error.

## Depends on

- #119 (surface definition)

## Related

- #118 (umbrella)
- #120 (canonical tracked changes)

Issue #118, opened 2026-05-03
[Tracking] SafeDocX: scoped tracked-changes-as-canonical, comparison decoupled


stevenobiajulu:
## Architecture statement

> For the supported revisionable AI editing surface, SafeDocX preserves first-class OOXML revision provenance at write time, validates AI-emitted revisions before they land, and supports selective accept/reject of the AI actor without re-diffing the whole document. Whole-document comparison remains available as an opt-in redlining tool but is removed from the default finalization path. Package-level mutations that have no native revision wrapper (theme replacement, side-part creation, relationship rewrites, content-type updates) are explicitly out of scope for this guarantee and are described to users under a separate contract.

## Why now

Today the finalization step routes saves through `compareDocuments(...)` and `documentReconstructor`. The reconstructor wraps detected deltas in fresh `w:ins`/`w:del` carrying the session's author/date, **overwriting any pre-existing reviewer attribution**. SafeDocX is therefore *re-authoring* tracked changes today. This proposal moves us to *co-authoring*: AI revisions are emitted at write time with stable IDs and never reconstructed after the fact.

Comparison/reconstruction is the dominant recent correctness hotspot — recent issues #35, #42, #61, #65, #76, #78, #94, #102, #103, #106, #107, #108, #110, #111 are concentrated in that pipeline.

## Atomicity, decomposed

"Atomic" is three separate properties; this proposal addresses (1) and (2):

1. **provenance-atomic** — every AI operation has explicit revision markup at write time
2. **compositionally lossless** — later pipeline stages don't erase edit boundaries or third-party revisions
3. **transactional** — failed operations don't partially mutate state (partially in place today)

## Tracked sub-issues

- [ ] #119 — Define the supported revisionable AI editing surface
- [ ] #120 — Make tracked changes the canonical representation for the supported surface
- [ ] #121 — Add validator for AI-emitted tracked-change OOXML
- [ ] #122 — Forbid untracked AI mutations within the supported surface (with carve-outs)
- [ ] #123 — Add selective accept/reject by revision-id (with author convenience mode)
- [ ] #124 — Invariant test corpus for accept/reject
- [ ] #125 — Mixed-author preservation corpus
- [ ] #126 — **(Milestone)** Remove comparison from the default finalization path
- [ ] #128 — **(Milestone)** Split comparison into a separate package, severing runtime/correctness dependency

### Dependency graph

```
#119 ──┬─→ #120 ──┐
       ├─→ #121 ──┤
[...]

PR #454, opened 2026-06-11
feat(docx-core): add AI revision validator with MCP transactional enforcement

stevenobiajulu:
## Summary

- Adds the `add-ai-revision-validator` OpenSpec proposal for issue #121.
- Defines docx-core validator behavior for AI-scoped revision errors, foreign revision warnings, range/field/package checks, and shared tracked-change vocabulary.
- Defines MCP server transactional enforcement requirements for clone-preflight write validation and save-time hard failures on invalid AI revisions.

## Validation

- `openspec validate add-ai-revision-validator --strict`

Ref: #121

PR #467, opened 2026-06-12
feat(docx-primitives): add AI revision validator

stevenobiajulu:
## What changed
- Adds `validate_revisions` in `@usejunior/docx-core` with a revision-element rule table, scoped severity partitioning, tainted baseline support, and a typed `RevisionValidationError`.
- Tracks `RevisionIdState.startId`, centralizes the revision-id element vocabulary, and adds document snapshot/restore plus `DocxZip.remove` for rollback.
- Adds an MCP post-write guard for AI write tools and a save-time gate that refuses session-owned malformed revisions before artifacts are written.
- Adds OpenSpec change artifacts and focused coverage for core validation, rollback, pre-existing defects, save refusal, and documented guard exemptions.

## Why
Issue #121 makes the validator the truth layer for #118's first-class write-time tracked-change emission. Real documents can contain malformed third-party revision markup, so the validator scopes hard errors to AI/session-owned revisions while preserving warnings for pre-existing defects.

## Validation
Passed:
- `npx openspec validate add-ai-revision-validator --strict`
- `npm run build`
- `npm run lint:workspaces` (passes with existing eslint warnings)
- `npx vitest run packages/docx-core/src/primitives/validate_revisions.test.ts packages/docx-mcp/src/tools/add_ai_revision_validator.test.ts`
- `npx vitest run packages/docx-core/src/primitives/document.test.ts packages/docx-core/src/primitives/validate_revisions.test.ts packages/docx-mcp/src/tools/add_ai_revision_validator.test.ts`

Full `npm run test:run` was run and reached the changed DOCX/MCP suites successfully, but local environment failures remain:
- `packages/docx-core/src/integration/cross-implementation-suite.test.ts` protocol tests fail because `tsx` cannot create its IPC pipe under the sandbox (`listen EPERM`).
- ODF LibreOffice oracle tests fail because `/opt/homebrew/bin/soffice` aborts (`Abort trap: 6`).

Ref: #121


stevenobiajulu:
## Peer-review fixes applied (`3a28bf5`)

The plan behind this PR was peer-reviewed (Codex + agy, dynamic review with executed probes); the initial implementation was then audited against those findings. This commit resolves everything that surfaced:

**BLOCKER — marker-id severity misattribution.** `partitionRevisionValidationIssues` attributed any numeric marker id ≥ `sessionStartId` to the session *before* consulting the baseline. Comment/permission range ids are allocated outside `RevisionIdState` (deliberately excluded from the seed), so a revision-free document (startId = 1) with a pre-existing unmatched `commentRangeStart`/`permStart` would hard-error and roll back **every** subsequent edit. Fixed: baseline taint/fingerprint outranks numeric attribution, and numeric attribution applies only to marker families allocated from the revision id space (move ranges, customXml ranges). Regression tests cover both directions (pre-existing junk stays warning; breaking a clean-at-open pair is still an error).

**MAJOR — core-package enforcement.** Only `insertParagraph` asserted post-write; `validateAfterExternalRevisionWrite` was exported but never called. Now every `RevisionContext`-bearing `DocxDocument` write method asserts session-scoped validity post-write, the tracked replace paths call the external hook, and tool catch handlers map `RevisionValidationError` → `REVISION_VALIDATION_FAILED` after rollback so the error code is identical whichever layer detects the defect.

**MAJOR — per-type test coverage.** Every revision element family now has a positive (zero-issue) and negative (missing-attribute) fixture, which also flushed two latent false positives in the rule table: `cellIns`/`cellDel`/`cellMerge`/`numberingChange`/`sectPrChange` are empty-by-schema (were EMPTY-flagged), and `w:ins`/`w:del` row markers inside `w:trPr` are empty by design.

**Minor:** `tblPrExChange` added to vocabulary + rules; `instrText` now checked against the fldChar state machine (`INSTRTEXT_OUTSIDE_FIELD`); `w:id` uniqueness spans story parts; `sessionEndId` removed (the seed invariant makes any revision-element id ≥ startId session-anomalous — an upper bound would let unallocated-id emitter bugs escape).

**Corpus dry-run** (per plan): validator over all 20 fixture DOCX files — zero false-positive hard errors, including `atomizer_redline.docx` which carries 35 pre-existing field-structure defects (16× unmatched field end, 18× `instrText` inside `w:del`, 1× instrText outside field — comparison-engine output, more evidence for the #118 thesis).

**Verification:** docx-core 1490 passed / docx-mcp 876 passed; lint 0 errors; `openspec validate add-ai-revision-validator --strict` clean; spec-coverage, conformance-citations, allure-labels gates pass. (`check:site` not run locally — needs its own install; CI covers it.)

**Follow-ups filed:** #472 (comments.xml root text not ins-wrapped, deferred drift), #473 (package invariants, #121 item 6).

stevenobiajulu:
## Closing as superseded by #454

While this PR was in review, PR #454 ("feat(docx-core): add AI revision validator with MCP transactional enforcement") merged to main from a parallel session and closed issue #121. The two PRs implement the same issue with incompatible designs (`validate_revisions.ts` + snapshot-rollback here vs `validate_ai_revisions.ts` + clone-preflight there); merging this one would ship two parallel validator subsystems, so this PR closes.

Comparison notes for the record:

- **#454's severity model is immune to the bricking class fixed here in `3a28bf5`.** It classifies by author match + explicit per-operation touched-id sets, never by numeric id-range attribution, so pre-existing comment/permission markers can't be misattributed to the session.
- **#454's diagnostic diff is stricter than this PR's baseline:** a count-multiset means N newly-introduced instances of a location-less finding can't hide behind one pre-existing instance — this PR's fingerprint Set had that masking tradeoff documented.
- **#454 also covers #121 item 6** (package invariants: relationship target resolution, content-type registration), which this PR had deferred to #473 — #473 is being closed accordingly.
- **Worth porting later from this branch:** the per-revision-family positive/negative fixture table in `validate_revisions.test.ts` (every family in the vocabulary, including the empty-by-schema nuances for `cellIns`/`cellDel`/`cellMerge`/`numberingChange`/`sectPrChange` and `trPr` row markers), and the fixture-corpus dry-run harness.
- #472 (comments.xml root-comment text not ins-wrapped — SUPPORT.md Table A drift) remains open and applies to main regardless of which validator won.

Branch retained for reference; commits `b3dfc3b` + `3a28bf5`.

PR #484, opened 2026-06-13
test(docx-core): lock the author->compare round-trip guarantee (#483)

stevenobiajulu:
## What

Locks the **author→compare round-trip guarantee** with tests + an OpenSpec requirement (issue #483).

`docx-core` both authors documents (`generateDocx`) and compares/redlines them (`compareDocuments`). The strategic value of owning both halves is that an authored document and a comparable document share one AST/OOXML model — but that synergy was *assumed and never asserted*. This PR proves it.

## Changes

- **New requirement** `Author-to-compare round-trip guarantee` on the `docx-generation` capability (OpenSpec change `add-generation-compare-roundtrip`), scenarios `SDX-GEN-100`..`SDX-GEN-104`.
- **New test** `packages/docx-core/src/generation/generation-compare-roundtrip.test.ts` — real `generateDocx` + `compareDocuments`, no mocks:
  - `SDX-GEN-100` — self-compare of an authored doc reports zero changes.
  - `SDX-GEN-101` — a known single-paragraph **replacement** produces exactly that redline (plus a format-only sub-case); accept-all == revised, reject-all == original.
  - `SDX-GEN-102` — accept-all == revised and reject-all == original, parametrized over **both** `rebuild` and `inplace`.
  - `SDX-GEN-103` — a `Page X of Y` field, `coverTermsTable`, and `signatureBlock` survive the compare round-trip (field-structure + table-cell paths).
  - `SDX-GEN-104` — **negative control**: a deliberately malformed authored field (dropped `fldChar` end) trips the `fieldStructure` reconstruction guard rather than passing silently, in both modes.

No production-source changes — this PR only asserts and documents existing behavior.

## Verification

- New suite: 8 tests pass.
- Full `docx-core` suite: 1514 passed, 2 expected-fail, 11 skipped.
- `openspec validate add-generation-compare-roundtrip --strict` ✓
- `npm run build` ✓; `preflight:ci` gates ✓ (lint, spec-coverage incl. `check:spec-coverage-generation`, allure-labels/filenames/quality, bdd-coverage, conformance-citations/doc, cycles). `check:site` was not run locally (the `site/` workspace deps aren't installed in the worktree); it is unaffected by this change and runs in CI.

### Notes on test design (peer-reviewed pre-implementation by Codex + agy)
- The "known edit" uses a **word replacement** because `modifiedParagraphs` only counts paragraphs with *both* an insert and a delete; a pure addition would report `modifiedParagraphs: 0`.
[...]

PR #490, opened 2026-06-13
fix(security): complete sanitization to clear CodeQL false-positive alerts

stevenobiajulu:
## What

Clears the eight remaining open CodeQL code-scanning alerts. Every one is a contextual false-positive — they live in build/test tooling or in the regex-grep tool whose input *is* a pattern by design, so none is an exploitable sink. But the fixes are trivial and make each sanitizer **correct and idempotent**, so the dashboard goes fully green with no dismissals.

| Alert(s) | Rule | Location | Fix |
|----------|------|----------|-----|
| #2, #3, #4 | `js/incomplete-sanitization` | `mdEscapeTableCell` ×3 (coverage scripts) | escape `\` before `\|` |
| #9, #18 | `js/incomplete-multi-character-sanitization` | `extractRawText`, `extractText` test helper | loop tag-strip until stable |
| #15 | `js/double-escaping` | `stripHtmlTags` (build-wiki) | decode `&amp;` **last** |
| #16 | `js/incomplete-multi-character-sanitization` | `stripHtmlTags` (build-wiki) | loop tag-strip until stable |
| #10 | `js/regex-injection` | `grep` tool | bound pattern count/length (ReDoS guard) |

## Details

- **Backslash-first escaping** — `mdEscapeTableCell` escaped `|`→`\|` without first escaping `\`, so a literal backslash in a cell value could combine into an unintended `\|`. All three identical copies fixed.
- **Loop-until-stable tag strip** — a single `.replace(/<[^>]+>/g, '')` pass can leave a freshly-formed `<...>` when angle brackets are nested/adjacent. Now loops until the string stops changing, so the output provably contains no tag and the sanitizer is idempotent. Applied to `extractRawText`, the `extractText` test helper, and `stripHtmlTags`.
- **Decode `&amp;` last** — `stripHtmlTags` decoded `&amp;`→`&` *before* the other entities, so `&amp;lt;` double-decoded into `<`. Reordered so `&amp;` is decoded last; `&amp;lt;` now resolves to the literal `&lt;`.
- **grep ReDoS guard** — the grep tool accepts user-supplied regex by design, so the pattern can't be escaped. Instead it now caps the pattern count (≤64) and combined length (≤2000 chars) before compiling, bounding catastrophic-backtracking cost. Two new tests cover the `PATTERN_LIMIT_EXCEEDED` guard.

#18 was previously suspected stale (it pointed past EOF on an older revision) but is now live — PR #484 grew that test file and line 987 contains a real tag-strip in a test helper. Fixed, not dismissed.

## Verification

- `npm run build` ✅, `npm run lint:workspaces` ✅ (0 errors)
- `npm run test:run` ✅ (2722 passed across all workspaces; +2 new grep guard tests)
- `check:spec-coverage` (exercises all three `mdEscapeTableCell`) ✅
- `check:conformance-citations`, `check:conformance-doc`, `check:allure-labels`, `check:allure-quality` ✅
- Functional spot-checks: `mdEscape("a\\b|c")` → `a\\\\b\\|c`; `stripHtmlTags("&amp;lt;")` → `&lt;`; loop-strip leaves no `<...>` on nested input.

Ref: CodeQL alerts #2, #3, #4, #9, #10, #15, #16, #18


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `f640c29b445695be8257bfc21bdb18ebeae56220`
Built from: `main` @ `f640c29`
Smoke: clean install + build + lint + full test suite, plus the merged grep tool driven against a real DOCX

**Steps**
- ✅ `npm ci` (exit 0, 0 vulnerabilities)
- ✅ build
- ✅ lint (`lint:workspaces`, 0 errors)
- ✅ tests (2723 passed across all workspaces; docx-core 1545 / 2 expected-fail / 1 skipped)
- ✅ grep tool end-to-end on a real document

**Real-world fixture**
- `common-paper-mutual-nda` (bundled `content/templates/common-paper-mutual-nda/template.docx`, 19,548 bytes)
  - Normal search: `grep "Confidential"` → 3 matches across 3 paragraphs, correct context highlighting — the merged grep path still searches real DOCX correctly.
  - New guard: a 2001-char pattern is rejected with `Combined pattern length 2001 exceeds 2000 characters` **before** the regex is compiled — the ReDoS bound (#10) works end-to-end through the CLI.

Visual evidence: skipped — the change is structurally invisible on a rendered page (read-only search guard + sanitization in build/test tooling; nothing alters how a document looks).

Note: the node-20 `workspace-test (20)` job failed once on an unrelated test (`generation-compare-roundtrip` — `generateDocx` byte non-determinism, node-20-only; node 22 + local node 26 both pass), well outside this PR's diff. A re-run passed green, confirming a flake.

**Cleanup**
- ✅ remote branch `tweak-codeql-sanitization-false-positives-20260613` already gone (auto-delete enabled)
- ✅ deleted local branch

Log: `/tmp/automerge-smoke-490-20260614-062800.log` (local).

PR #468, opened 2026-06-12
fix(docx-core): cover cross-implementation spec scenarios

stevenobiajulu:
## Summary
- extend the docx-core OpenSpec coverage validator to include cross-implementation-conformance
- align XIMPL .openspec labels with archived scenario names
- regenerate the docx-core traceability matrix showing 5/5 cross-implementation scenarios covered

## Validation
- npm run check:spec-coverage
- npm run check:allure-labels
- npx vitest run src/integration/cross-implementation-suite.test.ts --root packages/docx-core

Ref: #418

Issue #418, opened 2026-06-11
Extend spec-coverage validation to the cross-implementation-conformance capability


stevenobiajulu:
Split out from #390 (archive PR for `add-cross-implementation-conformance-suite`).

The M0 design accepted that scenarios in the `cross-implementation-conformance` capability are not machine-enforced by any coverage validator — `check:spec-coverage` only discovers `docx-primitives` and `mcp-server` spec paths. Now that the capability is archived into `openspec/specs/cross-implementation-conformance/`, extend a coverage script so its scenarios are mapped to tests (the self-check at `packages/docx-core/src/integration/cross-implementation-suite.test.ts`) the same way other capabilities are.

Ref: #283, #390

stevenobiajulu:
Done in #468 (`fix(docx-core): cover cross-implementation spec scenarios`, merged 2026-06-12), which extended `packages/docx-core/scripts/validate_openspec_coverage.mjs` to discover and enforce the `cross-implementation-conformance` capability and mapped its scenarios to the self-check at `packages/docx-core/src/integration/cross-implementation-suite.test.ts`.

Verified: `npm run check:spec-coverage -w @usejunior/docx-core` reports `PASS cross-implementation-conformance: 5/5 scenarios mapped (100.0%)`.

PR #513, opened 2026-06-20
test(docx-core): raise docx-comparison spec coverage to 68/69 and split coverage scripts

### LINKED ISSUE #469: check:spec-coverage never passes --strict to validate_openspec_coverage.mjs, so the docx-core matrix gate is WARN-only in CI
## Problem

The root `check:spec-coverage` script intends strict enforcement:

```json
"check:spec-coverage": "npm run check:spec-coverage -w @usejunior/docx-core -- --strict && ..."
```

but the docx-core package script is a chained command:

```json
"check:spec-coverage": "node scripts/validate_openspec_coverage.mjs && node scripts/validate_primitives_openspec_coverage.mjs"
```

npm appends `-- --strict` only to the **end of the chain**, so `validate_openspec_coverage.mjs` runs **non-strict** in CI (ci.yml `spec-coverage` job and release.yml). It prints `WARN docx-comparison: 59/69 scenarios mapped` and exits 0. Missing scenario coverage in docx-comparison or cross-implementation-conformance can therefore never fail CI — the gate is visibility-only.

Surfaced by peer review of #468 (verified by running `npm run check:spec-coverage -w @usejunior/docx-core -- --strict`: exit 0, with npm echoing the expanded command showing `--strict` reaching only the primitives validator).

## Why it can't be a one-line fix

Flipping `--strict` on for `validate_openspec_coverage.mjs` today turns CI red: docx-comparison has 10 missing scenarios (85.5%), pre-existing on `main` @ 066c930:

- Accept-side round-trip lemma is closed
- Bridge case provides a falsifiability layer for the new axiom
- Cross-run pass rescues inplace output
- Field-bearing arbitrary drives INV-FIELD-001 across operations *(phantom — see below)*
- Per-operation assertion strength matches the post-#217 engine *(phantom — see below)*
- Rebuild fallback only after all inplace passes fail
- Reject-side round-trip lemma is closed
- Residual obligations and the normalizeText modeling gap are documented
- Table-heavy run-fragmented templates preserve tracked table structure
- `inv_rt_001` sorry is replaced by a proof composing the named residual axiom and the lemmas

Two of those are **phantom misses caused by a multiline `.openspec(...)` tag** defeating the coverage parser at `packages/docx-core/src/integration/lean-spec-bridge.test.ts:1601-1604` (`[LEAN-FBA-01]` is split across lines; the regex swallows `', )` into the story name — visible as a mangled entry in the validator's "Extra scenario stories" output). Single-lining that tag should reclassify both as covered.

## Tasks

- [ ] Single-line the multiline `.openspec('[LEAN-FBA-01] ...')` tag in `lean-spec-bridge.test.ts` (reclaims 2 of the 10)
- [ ] Triage the remaining 8 missing docx-comparison scenarios: add mappings or mark `pending_impl` per the matrix conventions
- [ ] Re-wire the npm scripts so `--strict` reaches `validate_openspec_coverage.mjs` (split the chained script into two npm scripts, or pass the flag explicitly per command)
- [ ] Optional: add a `#### Scenario: [XIMPL-06] Protocol version mismatch exits with code 3` heading to `openspec/specs/cross-implementation-conformance/spec.md` — the requirement text and the test (`[XIMPL-06]` tag) both exist, but the scenario heading doesn't, so it shows up only as an "extra story"

Ref: #468, #418

stevenobiajulu:
## Summary

Groundwork toward strict spec-coverage enforcement (#469), shipped as an **honest partial** — it does **not** close #469 (see Residual).

The root `check:spec-coverage` chained both docx-core validators with `&&`, so a trailing `-- --strict` only reached the *primitives* validator; `validate_openspec_coverage.mjs` ran WARN-only and `docx-comparison` coverage could never fail CI. This PR splits the script so `--strict` *can* reach the openspec validator, and raises real `docx-comparison` coverage from 59/69 → **68/69** with genuine mappings only.

## Implementation

- **Split scripts**: docx-core `check:spec-coverage` → `check:spec-coverage:openspec` + `check:spec-coverage:primitives`; root invokes each so each validator gets its own flags. Script name kept stable for `ci.yml`, `release.yml`, `preflight:ci`.
- `[LEAN-FBA-01]`: single-lined the `.openspec` tag (Prettier trailing comma on the multiline form defeated the parser → 2 phantom-missing scenarios reclaimed).
- `[LEAN-RT-01..04]`: mapped on the field-bearing round-trip property test (empirical proxy / repo convention, cf. existing `[LEAN-FBA-05]`).
- `[LEAN-RT-05]`: **moved** to the single-fixture falsifiability test it actually describes (was mis-tagged on the 100-case property test).
- **New genuine test** `table-heavy-run-fragmented-inplace.test.ts` for "Table-heavy run-fragmented templates preserve tracked table structure" (asserts `reconstructionModeUsed: 'inplace'`, no fallback, `w:tbl` preserved, split-run insertion intact).
- `[XIMPL-06]`: added the scenario heading to the cross-impl spec (test tag already existed) → cross-implementation-conformance 6/6.

## Honest residual (why this doesn't close #469)

"Cross-run pass rescues inplace output" stays **unmapped**. The engine only surfaces cross-run pass diagnostics *on fallback* (`compare-types.ts`, `pipeline.ts`), so a test can't prove the cross-run pass is what *rescued* the output without an engine change to expose those diagnostics — and a deterministic "no-cross-run fails → cross-run rescues" fixture couldn't be constructed. Rather than tag-stuff a clean test, this scenario is left honest-WARN. **`--strict` is intentionally NOT yet enabled** for the openspec validator; `docx-comparison` sits at 68/69 (WARN). Closing #469 needs the diagnostics surface as a follow-up.

## Verification
- [x] `npm run lint -w @usejunior/docx-core` (tsc --noEmit) clean
- [x] Affected tests green (lean-spec-bridge, table-heavy-run-fragmented-inplace, field-cross-story-pipeline) — 18/18
- [x] `npm run check:spec-coverage` → `WARN docx-comparison: 68/69`, `PASS cross-implementation-conformance: 6/6`, exit 0
- [x] Peer-reviewed (agy + codex); tag-stuffing flagged and corrected before this commit

## Refs
- Refs #469 (partial — does not close), #468, #418


stevenobiajulu:
## Peer review (agy + codex, both dynamic) — round 2 (corrected diff)

A first review flagged tag-stuffing (scenarios mapped to tests that didn't exercise them). This commit corrects it; both reviewers re-ran dynamically against the corrected diff.

**Verdict: no blockers, mergeable as a coverage-improvement partial.**

- **New `table-heavy-run-fragmented-inplace.test.ts` — GENUINE** (both). Builds Delaware vs New York table-heavy docs, surgically splits revised table text across run boundaries, compares `inplace` with `premergeRuns:false`, asserts `reconstructionModeUsed==='inplace'`, no `fallbackReason`/`fallbackDiagnostics`, exactly one `<w:tbl>`, and the split-run insertion preserved. (codex nit, non-blocking: doesn't also assert `w:ins`/`w:del` wrappers.)
- **`[LEAN-RT-05]` placement — confirmed** moved to the single-fixture "axiom falsifiability layer" test, off the 100-case property test (satisfies the spec's "single fixture case, NOT empirical grounding").
- **Honest-exit wiring — confirmed, no enforcement regression.** On `origin/main` the `&&` chain already left the openspec validator effectively non-strict; this PR untangles the scripts and keeps openspec non-strict (WARN 68/69) while passing `--strict` to primitives/mcp/generation. `ci.yml`/`release.yml`/`preflight:ci` still resolve by name.

**Resolution**: tag-stuffing removed; two scenarios re-grounded (one genuine new test, one honest-WARN residual). Ships as a partial — does not close #469.

Issue #514, opened 2026-06-20
spec-coverage: deterministic THEN-keyword check so a tagged test must reference its scenario's observable


stevenobiajulu:
## Problem

`validate_openspec_coverage.mjs` checks tag **presence**, not tag **validity** — it confirms every `#### Scenario:` has an `.openspec('[ID] …')` somewhere, but has no idea whether the tagged test actually *asserts* that scenario. So a `PASS 69/69` can **oversell**: a tag slapped on an unrelated/clean test counts toward the ratio identically to a real mapping. The coverage number goes green while real coverage doesn't move.

This is not hypothetical. In #513 (partial fix for #469), the first implementation pass tag-stuffed several `docx-comparison` scenarios — e.g. "Cross-run pass rescues inplace output" tagged onto `SDX-GEN-103`, a *clean* author→compare round-trip that never sets up the scenario's GIVEN ("no-cross-run inplace passes fail round-trip safety"). Tests passed; the parser was satisfied; only peer review caught it. Late and expensive.

## Proposal

Add a deterministic, no-LLM check to `validate_openspec_coverage.mjs` that runs in the existing `spec-coverage` CI gate:

1. For each scenario, parse its `**THEN**` / `**AND**` lines and extract **code-like tokens** — identifiers, dotted property paths, enum/string literals, OOXML element names (e.g. `reconstructionModeUsed`, `'rebuild'`, `fallbackReason`, `w:tbl`).
2. For the test carrying that scenario's `.openspec` tag, require the test body (AST or source slice for that `test(...)` call) to mention **≥1** of those tokens.
3. **Fail-closed only when the scenario's THEN contains ≥1 recognizable code token.** Scenarios whose THEN is pure prose (documentation/Lean-build scenarios like `[LEAN-RT-03]`/`[LEAN-RT-04]`) have no extractable token → skip the assertion check for them (allowlist / auto-detect) so there are **no false positives** on doc/proof-style mappings.

### Why this shape
- Catches the most common, mechanical tag-stuffing (`SDX-GEN-103` would have failed instantly: "tagged with a scenario referencing `reconstructionMode`/`fallback`, but the test body mentions neither").
- Deterministic, fast, runs in CI — no LLM in the loop.
- Asymmetry-of-rot friendly: a stuffed tag tends to **fail** rather than silently pad the ratio.

### Acceptance criteria
- The check is wired into `check:spec-coverage:openspec` (or the validator itself) and gates CI under `--strict`.
[...]

PR #519, opened 2026-06-21
feat(docx-core): deterministic THEN-keyword spec-coverage check

stevenobiajulu:
## Summary

`validate_openspec_coverage.mjs` checked tag **presence**, not tag **validity** — a `.openspec('[ID] …')` slapped on an unrelated test counted toward the coverage ratio exactly like a real mapping, so `PASS 68/69` could oversell. This happened in #513 (a docx-comparison scenario tagged onto a clean `SDX-GEN` round-trip); only peer review caught it, late and expensive.

This adds a deterministic, no-LLM **THEN-keyword check** to the existing `check:spec-coverage:openspec` CI gate. It **fails closed** (independent of `--strict`, since validity is not negotiable the way the coverage ratio is).

## How it works

1. **Token mining** — for each `#### Scenario`, mine code-like tokens from its backtick code spans: `gateTokens` from the observable (THEN/AND), `matchTokens` = gate ∪ the action (WHEN). The check fires only when `gateTokens` is non-empty, so a pure-prose observable yields no token and is **auto-exempt** — the issue's "fail-closed only when THEN has a code token" rule, with no false positives on doc/proof-style mappings.
2. **Body-scoped matching** — slice per tagged test, group chained `.openspec().openspec()(…)` tags so every tag shares its test's body, and **drop the tag/title preamble** (the title echoes the scenario name and would otherwise satisfy a token like `inplace` on its own — defeating the check for exactly the kind of stuffing it targets).
3. **Allowlist** — a short, justified list exempts the few genuine mappings whose observable is asserted *outside* the TS body (external suite runner, compiled Lean proof, or a wrapping assertion helper).

The WHEN is in the match set (and GIVEN excluded) because a genuine test almost always invokes the subject under test even when it asserts the THEN through a returned value rather than the literal element name; this is what keeps real unit tests from being flagged.

## Verification

- **Zero false positives** against current `main`.
- An injected stuffed tag fails with a clear, actionable message; a same-scenario genuine mapping still passes (per-slice precision).
- Helpers are exported with a `realpathSync(argv[1])` entry guard so the pure logic is unit-tested without filesystem side effects. Tests live under `test-primitives/` (outside the validator scan roots) and build the `.openspec(` literal at runtime so the fixture file is not itself mis-scanned.

## Acceptance criteria
- [x] Wired into the `spec-coverage` gate, fails closed.
- [x] Token-extraction heuristic + auto-detected prose exemption + documented allowlist.
- [x] Unit tests: stuffed mapping FAILS, genuine PASSES, prose-only EXEMPT (plus title-echo, chained-tag, allowlist cases) — 13 tests.
- [x] Re-running against current `main` produces no false positives.

Ref: #514

### DIFF
--- a/packages/docx-core/scripts/validate_openspec_coverage.mjs
+++ b/packages/docx-core/scripts/validate_openspec_coverage.mjs
@@ -1,9 +1,10 @@
 #!/usr/bin/env node
 
 import fs from 'node:fs/promises';
+import { realpathSync } from 'node:fs';
 import path from 'node:path';
 import process from 'node:process';
-import { fileURLToPath } from 'node:url';
+import { fileURLToPath, pathToFileURL } from 'node:url';
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
@@ -24,7 +25,7 @@ const SPEC_CONFIGS = [
   },
 ];
 
-function normalizeScenarioName(value) {
+export function normalizeScenarioName(value) {
   return value
     .replace(/^Scenario:\s*/i, '')
     .replace(/^\[[^\]]+\]\s*/, '')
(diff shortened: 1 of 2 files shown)


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `d4ca1adce732de66af2079db41ddadcd3714a16c` (squash)
Built from: `main` @ `d4ca1ad`
Smoke: `npm run build` + `npm run check:spec-coverage` + `npm run lint:workspaces` + `npm run test:run -w @usejunior/docx-core`

Non-document PR (spec-coverage validator + unit tests; no DOCX code path), so the realistic entry point is the validator itself run against the real repo specs/tests.

**Steps**
- ✅ build (all workspaces)
- ✅ lint:workspaces (0 errors)
- ✅ **feature entry point** — `check:spec-coverage` ran the new THEN-keyword check against real specs/tests: `PASS THEN-keyword check: every tagged test references its scenario observable` (docx-comparison 68/69 coverage WARN unchanged; primitives canonical scenarios PASS)
- ✅ new unit test `test-primitives/validate_openspec_coverage.test.ts` (13 tests) passed
- ✅ docx-core tests: 1595 passed, 2 expected-fail, 1 skipped

**Note — one flaky failure, confirmed unrelated**
The full docx-core run showed 1 failure: `generation-compare-roundtrip.test.ts > Scenario: self-compare of an authored document is empty` (SDX-GEN-100). This is the documented `generateDocx` byte-nondeterminism flake, not a regression — this PR touches no generation code. Re-ran the file in isolation 3×: **8/8 passed each time**.

Log: `/tmp/automerge-smoke-519-20260621-060422.log` (local).

**Cleanup**
- ✅ remote branch `514-spec-coverage-then-keyword-20260620` deleted (repo auto-deletes head branches)
- ✅ removed worktree `.worktrees/514-then-keyword`
- ✅ deleted local branch + pruned origin

Issue #471, opened 2026-06-12
fix(docx-core): preserve merge-source commentsIds entries


stevenobiajulu:
## Summary

Neither reconstruction mode currently merges the merge-source side's `commentsIds.xml` entries into the comparison result. Merged-in comments therefore lose durable comment IDs; Word can regenerate them, so this is cosmetic, but the output drops existing durable-ID metadata from the source document.

Found during peer review follow-up for #448 / #465.

## Expected behavior

When comments are merged from the source side into the result, their matching `commentsIds.xml` rows should be carried forward or regenerated consistently so merged comments retain durable IDs.

## Reproduction

1. Create a DOCX with comments and a `word/commentsIds.xml` part containing `w16cid:commentId` rows for those comment paragraph paraIds.
2. Use it as the merge-source side in a comparison path that reconstructs comments into the result.
3. Inspect the resulting DOCX package.
4. Observe that `commentsIds.xml` entries from the merge-source side are absent, so the merged comments no longer have durable IDs until Word regenerates them.

Affected package: packages/docx-core

## ECMA-376 / OOXML impact

This concerns the Microsoft Word `commentsIds.xml` extension part ([MS-DOCX]) rather than ECMA-376 conformance-covered behavior. The impact is loss of Word durable-ID metadata for merged comments.

## Logs or artifacts

Ref: #448
Ref: #465

stevenobiajulu:
Shipped in #524.

PR #524, opened 2026-06-26
fix(docx-core): preserve merge-source commentsIds durable-ID rows

stevenobiajulu:
## Summary

The comment ancillary post-pass (`mergeCommentAncillaryParts` in `packages/docx-core/src/baselines/atomizer/pipeline.ts`) merged the merge-source side's `comments.xml`, `commentsExtended.xml`, and `people.xml`, but never merged `commentsIds.xml`. Merged-in comments therefore lost their Word durable-ID metadata ([MS-DOCX] `w16cid:commentId` rows) until Word regenerated them — cosmetic, but it drops existing durable-ID metadata from the source document.

This adds `mergeCommentsIds`, mirroring the existing `mergeCommentsExtended` shape:
- Carries forward source rows keyed by `w16cid:paraId` (the same paraIds threaded through `commentsExtended.xml`) for the expanded comment set.
- Appends into an existing result `commentsIds.xml`, or bootstraps the part (cloning the source root for namespaces, dropping non-matching rows) plus OPC metadata via `ensureOpcMetadata` / a new `COMMENTS_IDS_DESCRIPTOR`.
- Wired into `mergeCommentAncillaryParts` so it runs for both reconstruction modes (the post-pass picks `mergeSourceArchive` by mode).

## Tests

Added a focused regression describe block in `rebuild-auxiliary-merge.test.ts` (issue #471), covering both paths:
- rebuild **bootstraps** `commentsIds.xml` with the merge-source durable ID + content-type override + relationship.
- rebuild **appends** a reply's durable-ID row into an existing `commentsIds.xml` (both root and reply IDs survive).

A `addCommentsIdsPart` helper splices the part in, since the synthetic fixture does not emit `commentsIds.xml`.

## Gates run

- `npm run build` — pass
- `npm run lint:workspaces` — pass (0 errors; only pre-existing warnings)
- `docx-core` full vitest suite — pass (1580 passed, 2 expected-fail, 11 skipped)
- `npm run check:conformance-citations` — pass (JSDoc cites `[MS-DOCX]`, not ECMA-376)

## Reviewer notes

- No OpenSpec proposal: this is a bug fix following the existing auxiliary-merge pattern, not a new capability. No `.openspec()` tags added, so no `TEST_FEATURE` const required.
- The fix runs in both inplace and rebuild modes; the regression test exercises rebuild (bootstrap + append). Inplace shares the same code path.

Ref: #471

### DIFF
--- a/packages/docx-core/src/baselines/atomizer/pipeline.ts
+++ b/packages/docx-core/src/baselines/atomizer/pipeline.ts
@@ -1235,8 +1235,9 @@ async function mergeCommentAncillaryParts(
   // (and any roots not yet present in the result, defensively).
   await mergeMissingCommentDefinitions(resultArchive, commentById, includedCommentIds);
 
-  // Merge commentsExtended and people for the expanded set.
+  // Merge commentsExtended, commentsIds, and people for the expanded set.
   await mergeCommentsExtended(sourceArchive, resultArchive, includedParaIds);
+  await mergeCommentsIds(sourceArchive, resultArchive, includedParaIds);
   await mergePeople(sourceArchive, resultArchive, includedAuthors);
 }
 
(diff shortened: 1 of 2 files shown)


stevenobiajulu:
## Verdict: Approve (with nits)

Traced the implementation and tests end-to-end against issue #471. This is a clean, correctly-scoped bug fix that mirrors the established `mergeCommentsExtended` shape. No blocking findings.

### Why it's sound
- **Correct keying.** `includedParaIds` is built from each comment's first `<w:p>` `w14:paraId` (pipeline.ts:1176-1180), and `w15:paraId` / `w16cid:paraId` share that same value space, so matching merged comments against `w16cid:paraId` (pipeline.ts:1396-1401) is right.
- **Runs in both modes.** `mergeCommentsIds` is wired into `mergeCommentAncillaryParts` (pipeline.ts:1240), which is invoked unconditionally for both reconstruction modes at pipeline.ts:908. The "inplace shares the code path" claim is accurate.
- **OPC metadata.** Content-type `...wordprocessingml.commentsIds+xml` and relationship type `.../2016/09/relationships/commentsIds` are the correct MS-DOCX values; `ensureOpcMetadata` ignores the empty `referenceTag`/`idBearingTags`, so the descriptor is fine.
- **Tests genuinely exercise both branches and would fail pre-fix.** Bootstrap test: original has no comments → result is a clone of original with no `commentsIds.xml` → bootstrap branch; `expect(commentsIdsXml).not.toBeNull()` would fail without the fix. Append test: rebuild clones `originalArchive` (pipeline.ts:880,887), and the original side carries paraId `00000001`, so the result already has `commentsIds.xml` and the revised reply (`00000002`) takes the append branch (pipeline.ts:1411-1427). Confirmed the synthetic fixture emits `w14:paraId="00000001"/"00000002"` (synthetic-docx-fixture.ts:293,301), so the paraIds actually line up rather than passing vacuously. Hardcoded durableIds are first-principles expected values, not re-derived from the SUT.

### Conventions
- No OpenSpec proposal needed: bug fix following an existing pattern, no new public API (Check 17 clean). No `.openspec()` tags → no `TEST_FEATURE` const required. Correct call.
- Commit footer uses `Ref: #471` (no `Fixes #`), avoiding the auto-close trap.

### Nits (non-blocking)
1. The append test asserts both durable IDs survive but doesn't guard against *duplicate* rows for a paraId. Adding a count assertion (exactly one `w16cid:commentId` per paraId) would pin the dedup logic in rebuild-auxiliary-merge.test.ts.
2. Inplace mode is only covered transitively (shared code path) — no direct inplace regression test.
3. Since `Ref: #471` won't auto-close, remember to close #471 manually on merge.

stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `1ee82fc` (squash)
Built from: `main` @ `1ee82fc`
Smoke: real-document `compareDocuments` reconstruction + merged integration regression

**Steps**
- ✅ build (`@usejunior/docx-core`)
- ✅ lint (0 errors)
- ✅ integration test `rebuild-auxiliary-merge.test.ts` — 21 passed (incl. the new count-guard against duplicate `w16cid:commentId` rows per paraId)
- ✅ real end-to-end `compareDocuments` on a genuine commented DOCX

**Real-world fixture**
- `generation-phase6-drafting-notes.docx` — a genuine commented document (real `word/comments.xml` "Confirm the survival period with the client.", `commentsExtended.xml`, `people.xml`). Used as `original`; `revised` = the same doc with one real body word edited, comment parts left untouched.
  - `compareDocuments` ran reconstruction in **`rebuild` mode** (`fallbackReason: none`) — the exact mode this fix wires `mergeCommentsIds` into.
  - Output preserved all comment ancillary parts end-to-end: `comments.xml` ✅, `commentsExtended.xml` ✅, `people.xml` ✅ — no parts dropped through the rebuild merge.

**Honest scope note** — the new `mergeCommentsIds` branch keys off `word/commentsIds.xml`, a Microsoft-only (`w16cid`) durable-ID part that **no in-corpus fixture carries and LibreOffice (this machine's oracle) does not emit** — only Word authors it. So this smoke exercises the surrounding rebuild comment-merge path on a real document, while the durable-ID-preservation branch itself stays covered by the merged 166-line integration suite (which splices a real-shaped `commentsIds.xml` to drive the branch). A live Word-authored fixture would be needed to re-verify that specific branch end-to-end.

Visual evidence: n/a (comment-metadata merge — structurally invisible on a rendered page).
Log: `/tmp/automerge-smoke-524-20260630-130133.log` (local).

**Cleanup**
- ✅ deleted remote branch `471-preserve-merge-source-commentsids-20260626` (auto-delete)
- ✅ deleted local branch

PR #319, opened 2026-06-04
refactor(docx-core): split inPlaceModifier into focused modules (closes #318)

stevenobiajulu:
## Summary
Behavior-preserving split of `inPlaceModifier.ts` (3,185 LOC) — the in-place track-change reconstruction engine — into seven focused modules plus an orchestrator that re-exports the full public surface. Second half of the Workstream A modularization pass (follows #317's `document_view` split).

## Implementation
| Module | Lines | Responsibility |
|---|---|---|
| `inPlaceModifier-shared.ts` | 165 | revision-id state, DOM utilities |
| `inPlaceModifier-containers.ts` | ~200 | container topology + `ContainerResolutionError` |
| `inPlaceModifier-bookmarks.ts` | 243 | bookmark-marker survival/cloning |
| `inPlaceModifier-wrappers.ts` | ~615 | `w:ins`/`w:del`/move/format wrapping primitives |
| `inPlaceModifier-deletion.ts` | 391 | deleted/move/field content insertion |
| `inPlaceModifier-presplit.ts` | 306 | mixed/interleaved run pre-splitting |
| `inPlaceModifier-postprocess.ts` | 448 | grouping, no-op suppression, coalescing |
| `inPlaceModifier.ts` | 949 | `modifyRevisedDocument` orchestrator + barrel |

Code moved **verbatim** — no logic rewritten. Import graph is an acyclic DAG (`shared → {containers, bookmarks, wrappers} → deletion → presplit`; `postprocess → wrappers`; core → all leaves); the core barrel is the only importer of the leaf modules.

## Lean traceability (named anchors)
The pre-existing `Spec.lean` / `lean-spec-bridge.test.ts` citations referenced `inPlaceModifier.ts` by **line number** — and those numbers had already drifted onto blank/comment lines before this PR (unenforced by any test/script). Replaced them with greppable `@lean-segment:` anchors so traceability survives this and future splits:
- `@lean-segment: container-topology` → `inPlaceModifier-containers.ts`
- `@lean-segment: field-wrapper-emission` → `inPlaceModifier-wrappers.ts`

## Verification
- [x] `npm run build -w @usejunior/docx-core` clean
- [x] `npm run lint -w @usejunior/docx-core` clean
- [x] `npm run test:run -w @usejunior/docx-core` — **1302 passed, 11 skipped**
- [x] Public export surface byte-identical: **27 names**, diffed against pre-split `.d.ts`
[...]

Issue #272, opened 2026-05-27
ci(llm-gate): matrix jobs failing with 'Exit prior to config file resolving' on ~50% of recent runs


stevenobiajulu:
## Symptom

The LLM-Based Quality Gate (`.github/workflows/llm-based-quality-gate.yml`, added in #253) has been failing at the workflow level on ~5 of the last 6 PRs. The failures aren't the LLM finding issues — they're the Gemini CLI failing to start, so every matrix job exits with no result and the aggregate posts a `NO RESULTS` comment.

## Evidence

Last 6 LLM-gate runs on `main` and recent PRs (collected 2026-05-27):

| Run | PR | Conclusion | Aggregate verdict |
|---|---|---|---|
| [26490355255](https://github.com/UseJunior/safe-docx/actions/runs/26490355255) | #271 | success | ✅ PASS (14/14 with real justifications) |
| [26487987643](https://github.com/UseJunior/safe-docx/actions/runs/26487987643) | (PR closed) | failure | `NO RESULTS` |
| [26481977858](https://github.com/UseJunior/safe-docx/actions/runs/26481977858) | (PR closed) | failure | `NO RESULTS` |
| [26481976773](https://github.com/UseJunior/safe-docx/actions/runs/26481976773) | (PR closed) | failure | not checked |
| [26479483400](https://github.com/UseJunior/safe-docx/actions/runs/26479483400) | (`fix/allure-labels-validator-false-positives`) | failure | not checked |
| [26447905249](https://github.com/UseJunior/safe-docx/actions/runs/26447905249) | (`llm-gate-phase1`) | failure | not checked |

`NO RESULTS` comments observed on PR #270 and #269 ([safe-docx#270 LLM-gate comment](https://github.com/UseJunior/safe-docx/pull/270#issuecomment-4550837889), [safe-docx#269 LLM-gate comment](https://github.com/UseJunior/safe-docx/pull/269#issuecomment-4550007026)).

## Root cause snippet

From a representative failed matrix job (`Check 02 — Live DOM namespace-safe OOXML writes` on run 26487987643, job 77999406077):

```
[...checkout + setup steps succeed...]
Exit prior to config file resolving
##[error]Process completed with exit code 1.
```

The error is emitted by the Gemini CLI (`@google/gemini-cli@0.39.1`) BEFORE any model invocation. It happens before the `.gemini/settings.json` file the composite action writes is even read.

[...]

PR #533, opened 2026-07-07
ci(llm-gate): stop matrix flakiness from config-resolution and oversized diffs

stevenobiajulu:
## Problem

The LLM-Based Quality Gate is a required, blocking check (`LLM_GATE_BLOCKING=1`). It has been failing ~50% of runs in **two distinct ways**, and because it now blocks, each flake hard-blocks a clean, execution-verified PR and costs a manual `llm-gate/override`. Ref: #272.

## Failure mode 1 — `Exit prior to config file resolving` → `NO RESULTS`

The install step's `gemini --version` ran while the forced-empty `NPM_CONFIG_*` files (needed to harden `npm install` against `.npmrc` poisoning) were still in the environment. Gemini's startup update-check shells out to npm, and npm 11 aborts on that forced-empty config with `Exit prior to config file resolving`. Under `set -e` that aborted the whole install step **before any result artifact was uploaded**, so the aggregate posted `NO RESULTS`.

**Fix:** drop the `NPM_CONFIG_*` vars after `npm install` completes (mirroring the defense the runtime step already applies before the real model call), and make the version print non-fatal — it is a diagnostic, and a broken binary should fall through to the runtime step's determinate WARN rather than crash the job into `NO RESULTS`.

## Failure mode 2 — oversized diff → every row `WARN`

When `pr_diff.patch` exceeded `LLM_GATE_MAX_DIFF_BYTES`, the prompt **omitted** the diff and told the model to reconstruct it with `git diff`. That never bounded the payload — it moved a multi-hundred-KB diff from the prompt into tool output, which floods the model's context and, across a parallel matrix wave each retrying, burns token quota until the CLI returns empty responses and every check falls back to WARN (observed on PR #319's 3,185-line split).

**Fix:** bound the input instead — inline a `head -c`-truncated slice up to the byte budget with an explicit truncation marker, and still point the model at a scoped `git diff` to inspect the remainder on demand.

## Security properties preserved

- The `NPM_CONFIG_*` hardening still wraps the only line that needs it (`npm install`); the `unset` happens *after* the install completes.
- The truncated diff remains framed as untrusted data inside a `~~~diff` fence.

## Validation

- `actionlint` clean on the workflow; both YAML files parse.
- Every `bash` `run:` block passes `bash -n`.
- Truncation logic exercised in isolation: a 361-byte diff with a 100-byte budget produces a bounded 245-byte body with a correct `[diff truncated at 100 of 361 bytes]` marker and a scoped git-diff hint.
- Peer-reviewed with Codex before merge.

The gate running on this PR is itself the most direct smoke signal for the change.

https://claude.ai/code/session_01Bu5mvUQFLkhsz9MrvZGwBZ

### DIFF
--- a/.github/actions/llm-gate-check/action.yml
+++ b/.github/actions/llm-gate-check/action.yml
@@ -72,7 +72,22 @@ runs:
         npm install --silent --no-audit --no-fund --global "@google/gemini-cli@${GEMINI_CLI_VERSION}" --prefix "$GEMINI_PREFIX"
         echo "$GEMINI_PREFIX/bin" >> "$GITHUB_PATH"
         export PATH="$GEMINI_PREFIX/bin:$PATH"
-        gemini --version
+        # The forced-empty NPM_CONFIG_* files above are needed ONLY for the
+        # `npm install` line — they harden it against `.npmrc` poisoning. But
+        # they must not survive into the `gemini` invocation below: the CLI's
+        # startup update-check shells out to npm, and npm 11 aborts on the
+        # forced-empty config with "Exit prior to config file resolving". When
+        # that abort propagates through `set -e` it fails the whole install
+        # step, so the matrix job never uploads a result artifact and the
+        # aggregate posts NO RESULTS (issue #272, failure mode 1). Drop the
+        # vars before exercising the binary; the runtime step below applies the
+        # same defense before the real model call.
+        unset NPM_CONFIG_USERCONFIG NPM_CONFIG_GLOBALCONFIG NPM_CONFIG_REGISTRY
+        # Diagnostic only — never let a version print fail the job. If the
+        # binary is genuinely broken the runtime step falls back to a WARN
+        # result (a determinate verdict), which is strictly better than the
+        # job crashing into NO RESULTS.
+        gemini --version || echo "::warning::gemini --version failed after install (non-fatal diagnostic); proceeding to runtime check"
 
     - name: Build settings.json (tool allowlist)
       shell: bash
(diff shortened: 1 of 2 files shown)


stevenobiajulu:
## Peer review (Codex, dynamic) — addressed

Codex reviewed the two original fixes read-only, executing local probes:

- **`unset` is effective** — verified via a `/tmp` bash probe that child processes no longer inherit the `NPM_CONFIG_*` vars; the `npm install` hardening is preserved because the unset runs *after* the pinned install and before any `GEMINI_API_KEY` is present.
- **Truncation branch is shell-correct** — quoted, numeric-validated `head -c`, closed `~~~` fence. (A mid-UTF-8 cut can yield an invalid byte but the fence still closes byte-for-byte; Codex rated this a non-blocker for git patch text.)
- **Verdict: no blocking correctness or security issue** in the two targeted fixes.

**Substantive improvement Codex raised** (and my own stated uncertainty): the install-step fix cures only the *observed* pre-runtime crash; any other setup failure still yields `NO RESULTS`. Addressed in `5a596c6` — a defense-in-depth `if: always()` step backfills a determinate `WARN` result when a matrix item fails before the gate writes its verdict, and the artifact upload is now `if: always()`. Every check always contributes a row, so the aggregate never posts `NO RESULTS`. The fallback step carries no `GEMINI_API_KEY`.

The gate is running on this PR itself as the live smoke signal.

stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `dff32108f24228be63ea2d1ea89dfa2d28cc167c` (squash of #533)
Built from: `main` @ `8f01e5d`
Smoke: actionlint + behavioral shell probes against the merged files + live-CI evidence (CI-workflow-only change; no TS runtime surface, so the full package suite is not the relevant entry point — GitHub Actions is).

**Steps**
- ✅ `actionlint` clean on merged `.github/workflows/llm-based-quality-gate.yml`; both YAML files parse
- ✅ both mode-1 fixes present in merged main (`unset NPM_CONFIG_*`, non-fatal `gemini --version`), mode-2 truncation present, defense-in-depth fallback present
- ✅ **truncation behavioral probe** (compose_prompt extracted from merged action.yml): a 1250-byte diff at a 200-byte budget yields an **873-byte** bounded prompt (system-prompt + framing + 200B slice), exact `_[diff truncated at 200 of 1250 bytes]_` marker, closing `~~~` fence — confirms failure-mode-2 payloads are now bounded, not moved to an unbounded git-diff
- ✅ **live end-to-end**: the gate ran on this PR itself — **`Aggregate and post review` = SUCCESS**, and **all 17 individual checks = SUCCESS** (real verdicts, no `NO RESULTS`, no WARN cascade). This is the direct proof of the failure-mode-1 fix: every matrix job's install step completed instead of crashing.

Visual evidence: n/a — CI/config change, no rendered output.
Prod: n/a — this change is CI-only and does not affect the deployed site (Vercel check passed independently).
Log: `/tmp/automerge-smoke-533-20260706-225610.log` (local).

**Cleanup**
- ✅ deleted remote branch `272-fix-llm-gate-matrix-flakiness-20260706` (repo auto-delete had already removed it)
- ✅ removed worktree `.worktrees/272-fix-llm-gate-matrix-flakiness-20260706`
- ✅ deleted local branch

PR #594, opened 2026-07-21
feat(docx-core): accept any paragraph bookmark name as an edit anchor

### LINKED ISSUE #593: Accept any paragraph bookmark name as an edit anchor, not just _bk_* ids
## Summary

`replace_text` / `insert_paragraph` (and the other paragraph-anchored tools) only resolve safe-docx's own `_bk_*` bookmark ids. A host application that already stamps its own stable per-paragraph bookmarks cannot use them as anchors — it gets `ANCHOR_NOT_FOUND`.

Please accept **any bookmark name attached to the paragraph** as an anchor, not just `_bk_*`.

## Why this matters

Consumers that maintain their own paragraph identity have no sound way to translate their ids into `_bk_*` ids:

- `_bk_*` is `sha12(seed)` where the seed is `intrinsic:w14:<paraId>` when the paragraph has a `w14:paraId`, else a text/prev/next/ancestors fallback (`packages/docx-core/src/primitives/bookmarks.ts`). A consumer *can* recompute the intrinsic form — **but only if `w14:paraId` survives in their pipeline.**
- In the Junior harness it does not: **Aspose strips `w14:paraId` on load+save**, and the harness persists documents through Aspose. Measured:

  ```
  BEFORE (authored):        ['AAAA1111', 'BBBB2222']
  AFTER aspose load+save:   []
  AFTER safe-docx save:     ['AAAA1111', 'BBBB2222']   # safe-docx preserves them
  ```

- That leaves matching by **paragraph text**, which is not a sound identity. Two parsers can extract different text for the same paragraph (run concatenation without `xml:space="preserve"`, evaluated field codes), so a "unique" text match can resolve to the *wrong* paragraph — silently editing the wrong clause. Duplicate/empty paragraphs make it worse. For a legal-document editor that is not an acceptable failure mode.

safe-docx already **tolerates** stacked foreign bookmarks around a paragraph (`getParagraphBookmarkId` scans across them); it simply filters resolution to `_bk_*`. Accepting the other names is a small change that removes an entire class of unsound consumer workarounds.

## Proposed behavior

- `findParagraphByBookmarkId(doc, id)` matches if `id` equals **any** bookmark name attached to the paragraph (sibling-wrapped or inside), exact match.
- The **reported/canonical** id stays `_bk_*` (no change to what `read_file`/`grep` return), so nothing downstream shifts.
- Unknown names still resolve to `null` (no guessing).

## Consumer impact

Junior's harness can then pass its existing `jr_para_*` bookmarks straight through as anchors and delete its text-matching bridge entirely.

Related: UseJunior/legal-agent-harness#194, and #586 (accept_ai_edits docs/save-format hazard).

stevenobiajulu:
Closes #593.

## Summary
Paragraph-anchored tools (`replace_text`, `insert_paragraph`, …) resolved only safe-docx's own `_bk_*` ids. A host application that already stamps stable per-paragraph bookmarks got `ANCHOR_NOT_FOUND` passing them. They now work as anchors.

## Change
- **`getParagraphBookmarkNames(p)`** (new, exported): every bookmark name attached to a paragraph — sibling-wrapped and inside styles — in discovery order.
- **`findParagraphByBookmarkId`** matches any of those names (exact).
- **`getParagraphBookmarkId` is unchanged in meaning** — it still reports the canonical `_bk_*` id, so `read_file`/`grep` output and downstream consumers are unaffected.
- Unknown names still resolve to `null` — no guessing.
- Anchor param descriptions updated; `tool-reference.generated.md` regenerated (`check:tool-docs` clean).

## Why a text-matching workaround isn't acceptable
Consumers can't soundly reconstruct `_bk_*`. It's `sha12(seed)` where the seed is `intrinsic:w14:<paraId>` when present — recomputable, **but only if `w14:paraId` survives the consumer's pipeline.** Measured in the Junior harness:

```
BEFORE (authored):       ['AAAA1111', 'BBBB2222']
AFTER aspose load+save:  []                          ← Aspose strips them
AFTER safe-docx save:    ['AAAA1111', 'BBBB2222']    ← safe-docx preserves them
```

That leaves matching by paragraph **text**, which is not an identity: two parsers can extract different text for the same paragraph (run concatenation without `xml:space="preserve"`, evaluated field codes), so a "unique" text match can resolve to the wrong paragraph — silently editing the wrong clause. Duplicate and empty paragraphs make it worse. For a legal-document editor that failure mode is unacceptable.

safe-docx already *tolerated* stacked foreign bookmarks (the resolver scans across them); it just filtered to `_bk_*`. This removes an entire class of unsound consumer workarounds.

## Verification
- New `foreign_bookmark_anchors.test.ts` (4 scenarios): foreign anchor resolves; **two identical-text paragraphs correctly disambiguated** by distinct host bookmarks; stacked host + `_bk_` names resolve to the same paragraph while the canonical id stays `_bk_`; unknown name still fails.
- `docx-core`: **1190 passed**. `docx-mcp`: **876 passed**. `check:tool-docs`: clean.
- End-to-end against a locally built server — editing the *second* of two identical-text paragraphs by host bookmark only:

```
replace_text success: True        (previously ANCHOR_NOT_FOUND)
_bk_bda6144d0d1c | Identical clause text        ← untouched
_bk_eb4f9703a7b3 | EDITED VIA FOREIGN ANCHOR    ← correct paragraph
_bk_095ae280eb35 | Third clause
```

## Consumer follow-up
Once released, UseJunior/legal-agent-harness passes its `jr_para_*` bookmarks straight through and deletes its text-matching bridge (see legal-agent-harness#194 / #196).

### DIFF
--- a/packages/docx-core/src/primitives/bookmarks.ts
+++ b/packages/docx-core/src/primitives/bookmarks.ts
@@ -204,8 +204,11 @@ function paragraphsCoveredByBookmarkName(doc: Document, name: string): Element[]
     if (!p || !isParagraph(p)) continue;
     const pSpan = spans.get(p);
     if (!pSpan) continue;
-    // Intersect the covered content with the paragraph's subtree.
-    if (contentFirst <= pSpan[1] && contentLast >= pSpan[0]) covered.push(p);
+    // Intersect against the paragraph's CHILDREN (pSpan[0] + 1 …), not its own
+    // element position. A bookmark that closes at the very start of a paragraph
+    // (start before it, end as its first child) otherwise covers only the `w:p`
+    // boundary itself — no content — and would still claim the paragraph.
+    if (contentFirst <= pSpan[1] && contentLast >= pSpan[0] + 1) covered.push(p);
   }
   return covered;
 }
--- a/packages/docx-core/src/primitives/bookmarks.ts
+++ b/packages/docx-core/src/primitives/bookmarks.ts
@@ -125,14 +125,23 @@ function isParagraph(el: Element): boolean {
   return el.namespaceURI === OOXML.W_NS && el.localName === W.p;
 }
 
-export function getParagraphBookmarkId(p: Element): string | null {
-  // Supports:
-  // 1) sibling style: <w:bookmarkStart/> <w:p/> <w:bookmarkEnd/>
-  // 2) inside style: <w:p><w:bookmarkStart/> ... </w:p>
+/**
+ * Collect every bookmark name attached to a paragraph, in discovery order.
+ *
+ * Supports both attachment styles:
+ *   1) sibling: `<w:bookmarkStart/> <w:p/> <w:bookmarkEnd/>`
+ *   2) inside:  `<w:p><w:bookmarkStart/> ... </w:p>`
+ *
+ * Paragraphs routinely carry stacked bookmarks — our own `_bk_*`, `edit-*` spans,
+ * Word's own `_Toc*`/`_Ref*`, and names owned by an embedding application (e.g.
+ * a host that stamps its own stable paragraph ids). Returning all of them lets
+ * callers resolve an anchor by any of these names, not just ours.
+ */
+export function getParagraphBookmarkNames(p: Element): string[] {
+  const names: string[] = [];
 
-  // 1) Sibling style lookup.
-  // Handle stacked bookmarks (e.g. _bk_* plus edit-*) around the same paragraph.
-  // We scan backward across adjacent bookmark nodes until we hit another paragraph.
+  // 1) Sibling style. Scan backward across adjacent bookmark nodes until we hit
+  // another paragraph, so stacked bookmarks around one paragraph are all seen.
   const prev = prevElementSibling(p);
   const next = nextElementSibling(p);
   if (prev && next && isBookmarkEnd(next)) {
@@ -141,7 +150,7 @@ export function getParagraphBookmarkId(p: Element): string | null {
       if (isParagraph(cur)) break;
       if (isBookmarkStart(cur)) {
         const name = getAttr(cur, 'name');
-        if (name && name.startsWith('_bk_')) return name;
+        if (name) names.push(name);
       }
       cur = prevElementSibling(cur);
     }
@@ -152,9 +161,23 @@ export function getParagraphBookmarkId(p: Element): string | null {
   for (let i = 0; i < starts.length; i++) {
     const el = starts.item(i);
     const name = el ? getAttr(el, 'name') : null;
-    if (name && name.startsWith('_bk_')) return name;
+    if (name) names.push(name);
   }
 
+  return names;
+}
+
+/**
+ * The paragraph's canonical safe-docx id (`_bk_*`), or null when it has none.
+ *
+ * This intentionally stays `_bk_`-only: it is the id we *report* for a paragraph.
+ * To *resolve* an anchor the caller supplied, use `findParagraphByBookmarkId`,
+ * which accepts any bookmark name on the paragraph.
+ */
+export function getParagraphBookmarkId(p: Element): string | null {
+  for (const name of getParagraphBookmarkNames(p)) {
+    if (name.startsWith('_bk_')) return name;
+  }
   return null;
 }
 
@@ -273,11 +296,21 @@ export function insertSingleParagraphBookmark(doc: Document, p: Element): string
   return name;
 }
 
+/**
+ * Resolve an anchor to its paragraph by ANY bookmark name attached to it.
+ *
+ * Anchors are not limited to safe-docx's own `_bk_*` ids. An embedding
+ * application that already stamps stable per-paragraph bookmarks (and whose
+ * pipeline may not preserve `w14:paraId`) can pass those names directly, instead
+ * of having to reconstruct our content-derived ids — a reconstruction that has to
+ * fall back to matching paragraph text, which is not a sound identity and can
+ * silently target the wrong paragraph. Matching is exact on the bookmark name.
+ */
 export function findParagraphByBookmarkId(doc: Document, bookmarkId: string): Element | null {
   const paragraphs = Array.from(doc.getElementsByTagNameNS(OOXML.W_NS, W.p));
   for (const p of paragraphs) {
     if (!isParagraph(p)) continue;
-    if (getParagraphBookmarkId(p) === bookmarkId) return p;
+    if (getParagraphBookmarkNames(p).includes(bookmarkId)) return p;
   }
   return null;
 }
--- a/packages/docx-core/src/primitives/bookmarks.ts
+++ b/packages/docx-core/src/primitives/bookmarks.ts
@@ -156,6 +156,8 @@ function buildDocumentOrderSpans(doc: Document): Map<Node, [number, number]> {
  * a zero-length "point" bookmark sitting just before a paragraph, or a heading
  * bookmark spanning several paragraphs, masquerade as that paragraph's anchor.
  * We resolve the real range and report exactly which paragraphs it intersects.
+ *
+ * @conformance ECMA-376 edition 5, Part 1 § 17.13.6.2
  */
 function paragraphsCoveredByBookmarkName(doc: Document, name: string): Element[] {
   const starts = Array.from(doc.getElementsByTagNameNS(OOXML.W_NS, W.bookmarkStart)).filter(
@@ -168,6 +170,12 @@ function paragraphsCoveredByBookmarkName(doc: Document, name: string): Element[]
   const id = getAttr(start, 'id');
   if (!id) return [];
 
+  // The same w:id must not be opened twice — pairing would be ambiguous.
+  const startsWithId = Array.from(doc.getElementsByTagNameNS(OOXML.W_NS, W.bookmarkStart)).filter(
+    (el) => getAttr(el, 'id') === id,
+  );
+  if (startsWithId.length !== 1) return [];
+
   const ends = Array.from(doc.getElementsByTagNameNS(OOXML.W_NS, W.bookmarkEnd)).filter(
     (el) => getAttr(el, 'id') === id,
   );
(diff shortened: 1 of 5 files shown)


stevenobiajulu:
## Peer review (Codex, dynamic) — round 1 found a reproduced blocker; fixed in `dcb2917`

Codex ran a dynamic review with real fixtures and real MCP tool invocations. It **requested changes**, and it was right on both counts. Recording the findings and the fix.

### 🔴 Blocker — adjacency was treated as ownership (reproduced)
The first cut collected bookmark names by scanning siblings/descendants. Codex placed a **zero-length `_RefPointBefore` bookmark immediately before** a paragraph and the resolver attached it — then drove an actual `replace_text`:

```json
{ "target_paragraph_id": "_RefPointBefore", "success": true,
  "replacements_made": 1, "before_text": "Wrong target", "after_text": "WRONGLY EDITED" }
```

Multi-paragraph `_Toc`/`_Ref` ranges also resolved (to their first paragraph). I reproduced the point-bookmark case independently before fixing.

**Fix:** foreign names now go through `paragraphsCoveredByBookmarkName`, which pairs `bookmarkStart`/`bookmarkEnd` **by `w:id`** — ECMA-376 Part 1 §17.13.6.2 says pairing is by id, never by position — walks the real range using document-order spans, and accepts it **only when exactly one paragraph is covered**. A point bookmark covers none; a spanning bookmark covers many; an unpaired id or a name used twice is refused as ambiguous. All return `null` so the caller falls back rather than guessing.

### 🟠 High — my Assumption 2 was false: widening moved existing lookups
A paragraph can carry several `_bk_*` names while *reporting* only one. Matching "any name" made a canonical id resolve to a paragraph that doesn't report it:

```json
{ "oldResolvedText": "second", "newResolvedText": "first",
  "paragraphCanonicalIds": ["_bk_other", "_bk_target"] }
```

**Fix:** `findParagraphByBookmarkId` runs the original `getParagraphBookmarkId(p) === id` pass **first** and returns early for any `_bk_*` id, so canonical resolution is byte-for-byte unchanged and can never fall through to the foreign path.

### 🟡 Medium — docs were inconsistent
`include_fingerprint` still said *"Edit tools accept only `_bk_*` IDs"*; `style_source_id` was `_bk_*`-only; the new descriptions didn't say the behavior is DOCX-only. All four corrected to state the qualified contract, and the generated reference regenerated.

### 🟡 Medium — widened surface
Codex confirmed by invocation that `format_layout` and `clear_formatting` also accept foreign names (they share the resolver, along with `batch_edit`, `style_source_id`, `getParagraphElementById`, `extract_revisions`). With the range qualification these are now safe — an unqualified name resolves to `null`, preserving each tool's existing fallback — but the widening is intentional and now documented rather than silent.

### Tests
9 total (+5 from review): point bookmark adjacent to a paragraph · sibling multi-paragraph span · inline multi-paragraph span · inside-paragraph foreign bookmark (resolves) · unpaired `w:id` · duplicate foreign name · canonical-stability regression with two `_bk_*` names stacked · plus the original foreign-resolve, identical-text disambiguation, and unknown-name cases.

`docx-core` **1195 passed**, `docx-mcp` **876 passed**, tool docs regenerated, strict-TS build clean.

### Note on the earlier CI red
The `generation-numbering-level-justification` failure on the previous push was **not** from this change: the generation path contains no reference to any function this PR touches, it passes locally, and `main`'s own CI failed at this branch's base commit on a *different* test with identical code. Codex independently ran that test 250× without a structural failure. Fresh CI is running on `dcb2917`.

A round-2 review of the fix is in flight.

stevenobiajulu:
## Round 2 + 3 — two more escapes of the same defect class, both fixed

### Round 2 (`dcb2917` → reviewed): P1, reproduced again
Pairing by `w:id` was right, but I then measured the **marker span** `[start, end]`. For a point bookmark whose start/end are adjacent **inside** a paragraph, that span still intersected it — Codex drove a live `replace_text` and mutated the wrong paragraph a second time. It also caught a **mandatory conformance-lint failure** I'd introduced (JSDoc citing ECMA-376 without `@conformance`).

**Fix (`4073420`):** the covered interval is the content *strictly between* the markers — `[startSpan.exit + 1, endSpan.enter - 1]`. Empty ⇒ marks no content ⇒ refused, wherever the point sits; an end preceding its start falls out the same way. Also refuse two `bookmarkStart`s sharing one `w:id`. Conformance tags added — `check_conformance_citations: OK`.

### Round 3: aborted by the reviewer's own safety filter — so I ran its attack suite myself
Codex's round-3 run was killed mid-flight (`This content was flagged for possible cybersecurity risk`) before it produced a verdict — the adversarial malformed-XML fixtures tripped its classifier. It had already written a **17-fixture attack script**, so I executed that directly against the branch.

**4 mismatches. One was a real defect; three were not — judged individually rather than adopted wholesale:**

| fixture | result | call |
|---|---|---|
| `start-before-end-inside-at-start` | resolved `victim` | 🔴 **Real defect** — range enclosed only the `w:p` boundary, no content |
| `prooferr-only-inside-paragraph` | resolved `victim` | ✅ Correct — bookmark lies **entirely inside** that paragraph; that *is* the paragraph it marks |
| `xml-whitespace-only-inside-paragraph` | resolved `victim` | ✅ Same reasoning |
| `foreign-looking-canonical-prefix` (`_bk_notreal`) | resolved `marked` | ✅ Expectation was wrong — verified `getParagraphBookmarkId(p) === "_bk_notreal"`, so the paragraph genuinely reports it |

**Fix (`46f33d3`):** coverage now intersects a paragraph's **children** (`pSpan[0] + 1 …`), not its own element position, so a boundary-only range claims nothing. Positive controls added so content-covering ranges (partial-run, table-cell, whole-paragraph wrap) still resolve.

Declining the proofErr/whitespace expectations is deliberate: those bookmarks are unambiguously *within* one paragraph, so resolving there is the right answer, not a wrong-anchor. Requiring visible text would reject legitimate bookmarks around field codes and structural content.

### Final state
- **14 anchor tests**, covering every reproduction from all three rounds.
- Full attack suite: **16/17 match; the 3 remaining "mismatches" are the intentional decisions above** (and one reviewer error).
- `docx-core` 1198 passed · `docx-mcp` 876 passed · strict-TS clean · `check_conformance_citations` OK · tool docs fresh.
- **CI green** — `workspace-test` on Node 20 and 22 both pass.

### On the red CI you may have seen earlier
Three separate transient failures, none from this diff:
- `generation-numbering-level-justification` — passes locally; the generation path has **zero** references to anything here; Codex ran it 250× with no structural failure.
- `advanced-revision-classification` — fails intermittently *even in isolation*, and failed on `main` at this branch's base commit before the branch existed. Filed as **#602**.
- `codeql` — `read ECONNRESET` fetching the CodeQL bundle. Passed on re-run.

stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `b65600ff6f6e283ea9f967fbcd9f7d9eff60e2fa`
Built from: `main` @ `b65600f`
Smoke: real NVCA legal form → real Word bookmarks as edit anchors → real MCP `replace_text` → verified output

**Steps**
- ✅ clean build from merged `main` (`docx-core`, `docx-compare`, `docx-mcp`)
- ✅ `docx-core` **1200 passed** (2 expected fail, 11 skipped)
- ✅ `docx-mcp` **876 passed**
- ✅ `check:conformance-citations` OK · `check:tool-docs` OK
- ✅ primitive resolver against a real document
- ✅ end-to-end MCP edit through a real Word bookmark

**Real-world fixture**
- `nvca-indemnification-agreement` (via the open-agreements cache, SHA-256-verified download — not constructed by the smoke)
  - 153 paragraphs, **35 genuine Word bookmarks** (`_Hlk45869003`, `_Ref_ContractCompanion_*`, `_9kP1qJ9mv3AD`) — exactly the foreign-anchor shapes this PR resolves.
  - **33 resolved to exactly one paragraph**, and to semantically correct ones — e.g. `_Ref_ContractCompanion_9kb9Ur019` → *"Proceedings Other Than Proceedings by or in the Right of the Company"*.
  - **2 refused** by the range guard (point / spanning / ambiguous) — the ECMA-376 §17.13.6.2 pairing rule doing its job on real data rather than a synthetic fixture.
  - **0 unstable resolutions** across all 35.
  - Canonical `_bk_*` round-trip after indexing: **153/153 correct, 0 wrong** — confirming the regression fix (widening never moves an existing lookup) holds on a real document.

**End-to-end through the MCP server**
```
replace_text  target_paragraph_id = "_Ref_ContractCompanion_9kb9Ur019"   (a REAL Word bookmark)
              success=true  replacements_made=1
              before_text="Proceedings Other Than Proceedings by or in the Right of the Company. "
save          save_format=tracked  success=true
after         paragraphs with new text: 1 | still with old text: 0
              paragraph count 149 → 149   (no structural damage)
unknown anchor "_ThisBookmarkDoesNotExist" → success=false, ANCHOR_NOT_FOUND
```

Tracked markup in the saved file, verified at the XML level:
```
w:ins author='Junior AI' text='ROCEEDINGS OTHER THAN'
w:del author='Junior AI' text='roceedings Other Than'
visible text contains NEW: True | OLD: False
```
A correct minimal tracked change (the shared leading “P” is untouched), authored as `Junior AI`.

Visual evidence: **skipped as evidence** — this PR changes anchor *resolution*, which is structurally invisible on a rendered page; the meaningful proof is the tracked markup and the "exactly one paragraph changed" assertions above. Renders of the source and edited document were produced and opened in-session anyway for reference: `/tmp/automerge-smoke-594-20260722-075927/` (local).

Prod: n/a — safe-docx publishes on a `v*.*.*` tag, not on merge to `main`. This change reaches consumers with the next release; the harness (legal-agent-harness#196) currently pins `@usejunior/safe-docx@0.16.0` and will pass its `jr_para_*` bookmarks straight through once a release carrying this lands.

Log: `/tmp/automerge-smoke-594-20260722-075927.log` (local).

**Cleanup**
- ⚠️ remote branch `feat/foreign-bookmark-anchors` already deleted — auto-delete is enabled
- ✅ removed worktrees (`safedocx-anchors`, `sd594-r3`, `sd594-smoke`)
- ✅ deleted local branch, pruned refs

PR #612, opened 2026-07-23
fix(docx-mcp): guard selective revisions on clean save

### LINKED ISSUE #586: accept_ai_edits: session-scoped mutation is undocumented, and save_format="clean" silently discards the revisions it promises to preserve
## Summary

Two related gaps around `accept_ai_edits` / `reject_ai_edits` make it easy for a consumer to write code that looks correct, passes its own tests, and silently discards exactly the revisions the tool promises to preserve.

Both were hit for real while wiring `accept_ai_edits` into the Junior harness finalizer (UseJunior/legal-agent-harness#194).

## 1. The tool reference implies file mutation; the mutation is session-scoped

`packages/docx-mcp/docs/tool-reference.generated.md` documents `accept_ai_edits` as:

- `destructive: true`
- `file_path` — **required** — "Path to the DOCX or ODT file."

with no mention of `save`. Read plainly, that says "this writes to the file at `file_path`."

It doesn't. `packages/docx-mcp/src/tools/accept_ai_edits.ts` resolves a session, calls `session.doc.acceptAIEdits(...)`, then `manager.markEdited(session)` — and `markEdited` (`session/manager.ts:544`) only bumps `editCount`/`editRevision` and clears caches. Nothing touches disk. Persistence requires a separate `save` with `save_to_local_path`.

The realistic failure this produces in a consumer: copy the source to an output path, call `accept_ai_edits` on the copy, never call `save` → the output is a byte-identical copy of the fully-marked-up input. It's a **silent no-op**, not an error, and any test that stubs the MCP boundary will pass. We wrote exactly this before catching it by reading the source.

**Ask:** state in the tool description (and therefore the generated reference) that the mutation applies to the in-memory session and requires a subsequent `save` to persist. Consider whether `destructive: true` is the right annotation for a tool that doesn't write, or whether the required-looking `file_path` should be documented as a *session selector* rather than a write target.

## 2. `save_format: "clean"` silently nullifies the tool's headline guarantee

`accept_ai_edits` is described as "leaving all other (e.g. third-party reviewer) revisions **byte-untouched**" — that byte-preservation of a counterparty's redlines is the entire reason to choose it over `accept_all_revisions`.

But the very next call the consumer is *required* to make can throw that away. In `packages/docx-mcp/src/tools/save.ts`:

- `save_format: "clean"` strips all remaining tracked changes.
- `track_changes: false` aliases to `"clean"` (`save.ts:166`).
- omitting `save_format` defaults to `"both"` (`save.ts:169`).

So `accept_ai_edits({author: "AI"})` followed by `save({save_format: "clean"})` produces a document with the third-party revisions **gone** — the precise outcome the author/id selector existed to prevent. Nothing in the API surface flags the contradiction.

In our case this is not hypothetical: the harness's existing `save()` helper passed no `save_format`, so the obvious "reuse the existing helper" implementation would have landed on a non-`tracked` save and discarded opposing counsel's open redlines from a document going back to a client. We now pin `save_format="tracked"` explicitly.

**Ask:** when a session has had `accept_ai_edits`/`reject_ai_edits` applied with an `author` or `revision_ids` selector, and a subsequent `save` requests `save_format: "clean"`, either emit a warning in the save report or require an explicit opt-in acknowledging that selectively-preserved revisions are being discarded. A field on the save report noting how many non-targeted revisions were dropped would also make the loss visible after the fact.

## What we're not asking for

We considered requesting a `save_to_local_path` on `accept_ai_edits` for a one-shot functional form (symmetric with `compare_documents`), and decided **against** proposing it: every mutating tool here (`replace_text`, `insert_paragraph`, …) is session-based with an explicit `save`, and adding an output path to one of them would break a coherent model to paper over what is really a documentation problem. Item 1 above is the fix.

## Context

- Consumer side: UseJunior/legal-agent-harness#194 (routes the document-edit finalizer's "revised clean" step through `accept_ai_edits` with `author="Junior AI"`, so counterparty revisions survive for human adjudication).
- Observed against `@usejunior/safe-docx@0.15.0`.
- The harness PR does not block on this — we pin `save_format="tracked"` explicitly. Filing so the next consumer doesn't have to read the source to find it.

stevenobiajulu:
## Summary

- make `accept_ai_edits` / `reject_ai_edits` explicitly report that successful mutations are session-scoped and require `save`
- record successful selective revision dispositions in the DOCX session
- fail closed when a later `clean`/`both` save would auto-accept remaining revisions by the session AI author
- add an explicit `allow_discard_preserved_revisions` acknowledgement for callers that intentionally want that clean-output behavior
- preserve tracked-save behavior and document the contract in the generated tool reference

## Why

Selective revision tools can intentionally leave some AI-authored revisions unresolved. A later clean save previously auto-accepted all remaining revisions by that author, silently defeating the caller's selective intent. This makes that irreversible disposition explicit and agent-readable.

## Validation

- `npm run test:run -w @usejunior/docx-mcp -- src/tools/add_selective_ai_accept_reject.test.ts` (9/9)
- `npm run build -w @usejunior/docx-mcp`
- `npm run lint -w @usejunior/docx-mcp`
- `npm run check:tool-docs`
- repository test run: all workspaces passed except five environment-blocked docx-core probes; those five passed when rerun outside the sandbox (14 passed, 1 expected skip)

Fixes #586

### DIFF
--- a/packages/docx-mcp/docs/tool-reference.generated.md
+++ b/packages/docx-mcp/docs/tool-reference.generated.md
@@ -110,7 +110,7 @@ Insert a paragraph before/after an anchor paragraph by paragraph id. Supports DO
 
 ## `save`
 
-Save document. For DOCX: saves clean and/or tracked changes output. For ODT: saves an .odt package. For Google Docs: checkpoint (default) returns revisionId, or snapshot exports as DOCX. Surface: revisionable — the save report lists both the AI revisions applied and a non-revision change manifest of any package-level mutations (comment/footnote side parts, relationships) that have no tracked-change wrapper.
+Persist the current in-memory document session. For DOCX: saves clean and/or tracked changes output. For ODT: saves an .odt package. For Google Docs: checkpoint (default) returns revisionId, or snapshot exports as DOCX. Surface: revisionable — the save report lists both the AI revisions applied and a non-revision change manifest of any package-level mutations (comment/footnote side parts, relationships) that have no tracked-change wrapper.
 
 - readOnly: `false`
 - destructive: `true`
@@ -123,6 +123,7 @@ Save document. For DOCX: saves clean and/or tracked changes output. For ODT: sav
 | `clean_bookmarks` | `boolean` | no |  |
 | `save_format` | `enum("clean", "tracked", "both")` | no |  |
 | `allow_overwrite` | `boolean` | no |  |
+| `allow_discard_preserved_revisions` | `boolean` | no | Explicitly allow a clean artifact to auto-accept remaining revisions by the session AI author after accept_ai_edits/reject_ai_edits selectively left revisions unresolved. Default: false. |
 | `tracked_save_to_local_path` | `string` | no |  |
 | `tracked_changes_author` | `string` | no |  |
 | `tracked_changes_engine` | `enum("auto", "atomizer")` | no | Deprecated and ignored (#126). The redline is now the session's write-time tracked markup, serialized directly — there is no comparison engine to select. Use the compare_documents tool for comparison-based redlines. |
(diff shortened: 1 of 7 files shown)


stevenobiajulu:
Post-merge real-document smoke passed on `tests/test_documents/open-agreements/mutual-nda.docx`. With two existing runs marked as SafeDocX AI insertions, selectively accepting revision 9001 armed the persistence guard; an unacknowledged clean save was blocked with `SELECTIVE_REVISIONS_WOULD_BE_DISCARDED`; tracked save succeeded with 9001 accepted and unresolved revision 9002 preserved.

PR #613, opened 2026-07-23
fix(docx-compare): resolve insertion provenance collisions

### LINKED ISSUE #359: Pre-tracked insertion content that collides with the other input's plain text loses provenance — reject round-trip diverges (inline w:ins revised, paragraph-insert either side)

stevenobiajulu:
## Summary

- promote revised-side pre-tracked insertions to settled content when their matched original lineage is plain
- remove the corresponding paragraph insertion mark for matched settled paragraphs
- keep unmatched revised insertions tracked and fail closed on ambiguous wrapper shapes
- re-enable insertion-collision pairs in the generated Lean bridge
- flip the two pinned fallback characterizations into in-place projection regressions

## Why

When revised calls matched text inserted but original treats it as settled, preserving the revised `w:ins` makes Reject All drop text that `rejectAllChanges(original)` keeps. The shared lineage must be represented as settled so both Accept All and Reject All projections remain lawful without an unnecessary rebuild fallback.

Fixes #359

## Validation

- `npm run build`
- `npm run lint:workspaces`
- `npm run test:run` (all workspaces green except nine environment/contention-sensitive docx-core cases in the restricted sandbox)
- unrestricted rerun of all nine reported cases: 27 passed, 1 expected external-suite skip
- Lean spec bridge: 14/14 passed, including generated collision coverage
- related atomizer/provenance suites: 151/151 passed

### DIFF
--- a/packages/docx-compare/src/baselines/atomizer/inPlaceModifier.ts
+++ b/packages/docx-compare/src/baselines/atomizer/inPlaceModifier.ts
@@ -327,7 +327,7 @@ function removeParagraphInsertionMarker(paragraph: Element | undefined): void {
  * @conformance ECMA-376 edition 5, Part 1 § 17.13.5.20
  * @see https://github.com/UseJunior/safe-docx/issues/359
  */
-function resolveRevisedInsCollisionOnRun(
+export function resolveRevisedInsCollisionOnRun(
   atom: ComparisonUnitAtom,
   run: Element
 ): void {
--- a/packages/docx-compare/src/baselines/atomizer/inPlaceModifier.ts
+++ b/packages/docx-compare/src/baselines/atomizer/inPlaceModifier.ts
@@ -13,7 +13,12 @@
 
 import type { ComparisonUnitAtom } from '@usejunior/docx-core';
 import { CorrelationStatus } from '@usejunior/docx-core';
-import { childElements, findChildByTagName, insertAfterElement } from '@usejunior/docx-core';
+import {
+  childElements,
+  findChildByTagName,
+  insertAfterElement,
+  unwrapElement,
+} from '@usejunior/docx-core';
 import { enforceConsumerCompatibility } from './consumerCompatibility.js';
 import { XMLSerializer } from '@xmldom/xmldom';
 import { findChild } from '@usejunior/docx-core';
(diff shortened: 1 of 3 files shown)


stevenobiajulu:
Post-merge real-document smoke passed on `tests/test_documents/open-agreements/mutual-nda.docx`. I wrapped the existing title run (`Mutual Non-Disclosure Agreement`) in a revised-side pre-tracked insertion, compared in inplace mode, and verified: mode remained `inplace`; the stale insertion author was removed; Accept matched revised text; Reject matched original text.

Issue #582, opened 2026-07-15
Rebuild fallback loses rsids, sectPr, fields, and content controls — close the fidelity gap with the in-place path


stevenobiajulu:
`restoreUntouchedBlocks()` (docx-core `primitives/minimal_save.ts`) gives the
in-place save path byte-level preservation for every untouched block: after a
save-time rebuild from the spec model, any block that is
serialization-equal to the normalized original is swapped back to the
pristine original element (descending into edited tables so untouched
rows/cells restore too). When alignment fails it returns 0 and the output
degrades to the full re-serialization.

That fallback is lossy by construction: the spec model does not represent
rsids, `sectPr`, complex/simple fields, or content controls (`sdt`), so a
full rebuild silently drops or flattens them. Today the damage is limited to
edited blocks and alignment-failure cases, but it caps how far the rebuild
path can ever be trusted.

Two candidate directions (not mutually exclusive):

1. **Model the features.** Add first-class spec-model representations for
   rsids, section properties, fields, and content controls, so
   `generateDocx()` can emit them. Highest fidelity ceiling, largest surface
   area — each feature is its own project.
2. **Opaque passthrough.** Attach unrecognized XML (attributes and child
   elements the model does not understand) to spec nodes as opaque blobs and
   re-emit them verbatim on rebuild. Lower ceiling (no semantic edits to the
   preserved content) but generalizes to everything not yet modeled,
   including future OOXML extensions.

Suggested acceptance test either way: round-trip a corpus of real-world DOCX
files through a forced full rebuild (bypass `restoreUntouchedBlocks`) and
diff `document.xml` — the gap between that diff and the in-place path's diff
is the metric to drive down.

Context: this would also de-risk a planned Markdoc-as-authoring-layer
workflow that compiles Markdoc diffs to `batch_edit` operations — the more
the rebuild fallback preserves, the less catastrophic an alignment failure
becomes.

stevenobiajulu:
[...]

PR #623, opened 2026-07-24
feat(docx-compare): verify ancillary field stories

stevenobiajulu:
## Summary

- relationship-select valid section header/footer bindings from the assembled package instead of globbing part names
- strictly validate every selected header/footer and final note entry with independent field state
- emit source-first PAGE/NUMPAGES/REF/PAGEREF preservation evidence with base/import provenance
- reject unsafe targets, ambiguous note IDs, malformed selected stories, and terminal evidence mismatches before publication
- allow one ancillary-triggered inplace-to-rebuild fallback with typed diagnostics
- keep Lean protocol v3 and its fixed main/footnotes/endnotes, inplace-only scope unchanged

## Evidence

- real NVCA source-derived comparison passes in true inplace and forced rebuild with nonzero footer PAGE and footnote REF evidence
- original NVCA source-vs-filled large-diff regression remains covered
- focused docx-core: 52/52
- focused docx-compare: 57 passed, 23 Lean-binary skips
- full docx-core excluding three local-only LibreOffice tests: 1,242 passed, 2 expected failures, 3 skipped
- full docx-compare: 602 passed, 24 skipped
- build and workspace lint passed; 9 pre-existing warnings, 0 errors
- strict OpenSpec, spec coverage, conformance citations, generated conformance idempotence, and diff checks passed

LibreOffice-backed local-only tests were not rerun after final repair because macOS sandbox startup aborts inside AppKit even under `--headless`; CI does not install LibreOffice. No runtime dependency is added.

## Scope boundaries

This does not synthesize or compare ancillary revisions, evaluate fields, paginate, resolve bookmarks, establish complete note-reference integrity, or broaden the Lean certificate. Exact canonical range preservation is recorded as a SafeDocX metamorphic invariant, not an ECMA-376 requirement.

Refs #582

### DIFF
--- a/openspec/changes/verify-ancillary-field-stories/design.md
+++ b/openspec/changes/verify-ancillary-field-stories/design.md
@@ -1,0 +1,362 @@
+## Context
+
+`evaluateSafetyChecks` currently applies the Lean-pinned
+`validateFieldStructure` predicate to `word/document.xml` and footnote/endnote
+entries extracted from both input sidecars. That runs before reconstruction
+mode selection and before auxiliary definitions are merged into the result.
+Header/footer parts are excluded because selecting them requires section-binding
+and relationship resolution.
+
+The final package is mode-dependent. True inplace output clones the revised
+archive and imports still-referenced definitions from the original; forced
+rebuild and inplace fallback clone the original archive and import definitions
+from the revised. Before either assembly, auxiliary ID collision handling may
+mutate the revised in-memory archive and renumber definitions and references.
+Final ancillary validation and preservation evidence must therefore use those
+post-collision in-memory archives and the actual merge results, not the original
+input buffers or a reconstructed guess at provenance.
+
+## Goals / Non-Goals
+
+- Goals: reuse or extract the existing `sectPrAudit` binding contract to select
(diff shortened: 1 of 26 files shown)


stevenobiajulu:
Post-merge smoke passed on merged commit `4225b4851aa3eb6c5ea11ace23fe149f4c331d1f`.

- Used the real checked-in NVCA COI source DOCX and a minimal edit made through the shipped document primitive.
- Built merged `main`, then exercised both true `inplace` and forced `rebuild`.
- 213/213 assertions passed.
- Both modes emitted passing ancillary evidence with 5 relationship-selected footer `PAGE` ranges and 30 final footnote `REF` ranges. Every range had a canonical match and expected source/provenance.
- Accept-all recovered revised body text; reject-all recovered original body text.
- Both outputs were ZIP-readable with 45 entries and 44 parseable XML/relationship parts.
- LibreOffice was intentionally not invoked. This path is structurally invisible, and local macOS sandbox startup aborts during LibreOffice AppKit registration.

Artifacts are retained locally under `/tmp/safe-docx-pr623-artifacts`.

Issue #641, opened 2026-07-25
docx-compare: deleting a bookmarked paragraph detaches its bookmark (markers hoisted to body level)


stevenobiajulu:
## Summary

When the atomizer INPLACE reconstruction recreates a deleted paragraph whose content was enclosed by bookmark boundary markers, the `<w:bookmarkStart>`/`<w:bookmarkEnd>` pair is hoisted **out of the paragraph to body level**, collapsing to a zero-length range that no longer covers the text it named.

## Reproduction

Original body:

```xml
<w:p><w:r><w:t>Leading survivor</w:t></w:r></w:p>
<w:p><w:pPr><w:keepNext/></w:pPr>
  <w:bookmarkStart w:id="41" w:name="DeletedBoundary"/>
  <w:r><w:t>Bookmarked deleted text</w:t></w:r>
  <w:bookmarkEnd w:id="41"/></w:p>
<w:p><w:r><w:t>Trailing survivor</w:t></w:r></w:p>
```

Revised body drops the middle paragraph. Compare with `{ engine: 'atomizer', reconstructionMode: 'inplace' }`.

## Actual output

```xml
<w:p><w:r><w:t>Leading survivor</w:t></w:r></w:p>
<w:bookmarkStart w:id="41" w:name="DeletedBoundary"/>
<w:bookmarkEnd w:id="41"/>
<w:p><w:pPr><w:keepNext/><w:rPr><w:del .../></w:rPr></w:pPr>
  <w:del ...><w:r><w:delText>Bookmarked</w:delText></w:r>...</w:del>
</w:p>
<w:p><w:r><w:t>Trailing survivor</w:t></w:r></w:p>
```

Both markers sit adjacent at body level, **before** the recreated paragraph.

## Expected

The bookmark range should continue to enclose the deleted content, i.e. the markers should remain inside the recreated paragraph surrounding the `<w:del>` run — the same way the move-source path already keeps them (verified: a moved bookmarked paragraph retains its markers inside the `<w:p>`, around `<w:moveFrom>`).

## Notes

- Body-level `bookmarkStart`/`bookmarkEnd` is **schema-legal** (`CT_Body → EG_BlockLevelElts → EG_ContentBlockContent → EG_RunLevelElts` in both the vendored strict and transitional ECMA-376 schemas), so no validator catches this. The defect is semantic: the bookmark survives by name but stops referring to anything.
[...]

Issue #639, opened 2026-07-25
Raise docx-compare inPlaceModifier.ts test coverage (74.5% → as high as feasible)


stevenobiajulu:
## Goal
`packages/docx-compare/src/baselines/atomizer/inPlaceModifier.ts` (1091 lines) applies tracked-change revisions in **inplace reconstruction mode** — the per-status handlers that mutate the original tree in place rather than rebuilding. It sits at **74.5% statement coverage (79 uncovered statements)**, now the lowest-covered atomizer file after the pipeline sweep. Add characterization tests to raise it as high as cleanly feasible. **Exhaustive intent** — multiple test files fine; use real DOCX fixtures.

## Uncovered clusters (statement lines, from a v8 run)
`142, 278-279, 289, 358-362, 395-398, 403, 418, 424-428, 436-442, 482, 499-500, 506, 515, 523-526, 536-537, 541-544, 577, 593, 611-614, 629, 646-647, 653-656, 662, 670-673, 683-684, 688-691, 717, 750, 829-831, 837, 846, 875-877, 915, 983-1009, 1025`
Likely themes (VERIFY by reading the code): per-status inplace handlers (insert/delete/move/format edge cases), whole-paragraph vs inline handling, provenance/auxiliary interactions, and defensive guards. The `983-1009` block is the single largest contiguous gap — read it first.

## Existing sibling tests (match their conventions, don't duplicate)
- `inPlaceModifier.test.ts`, `inPlaceModifier-handlers.test.ts`, `inPlaceModifier-split.test.ts`, `inPlaceModifier.traceability.test.ts` — read these to see what's ALREADY covered (word insert/delete, whole-para insert/delete, format change, move, replacement are done in `-handlers`). Target the branches those miss.

## How to measure coverage (exact recipe)
```bash
cd packages/docx-compare
npm run build -w @usejunior/docx-core -w @usejunior/docx-compare   # REQUIRED first — spawn-based tests false-fail on missing dist/
npx vitest run --coverage --coverage.reportOnFailure=true --coverage.reporter=json
node -e '
const cov=require("./coverage/coverage-final.json");
for(const[f,d]of Object.entries(cov)){if(!/\/inPlaceModifier\.ts$/.test(f))continue;
const s=d.s,m=d.statementMap;const u=Object.entries(s).filter(([k,v])=>v===0).map(([k])=>m[k].start.line);
[...]

PR #637, opened 2026-07-25
test(docx-compare): cover atomizer pipeline auxiliary-merge + safety paths (closes #636)

stevenobiajulu:
## Summary
Characterization tests for the atomizer comparison pipeline (`compareDocumentsAtomizer` and its internal helpers), lifting `pipeline.ts` statement coverage **47.0% → 89.5%** (672/751). Test-only; no production changes. Implemented via `/codex-implement`, supervised + independently verified.

## What's covered
- **pipeline-auxiliary-notes.test.ts** — footnote/endnote auxiliary-part merging through the comparison + reconstruction path.
- **pipeline-comment-ancillary.test.ts** — comment-range ancillary-story merging and comment-part reconciliation.
- **pipeline-safety-guards.test.ts** — input/publication guards: missing-`w:body` rejection, `AncillaryStorySafetyError` fail-closed on a malformed note part (`NOTE_PART_XML_INVALID`), and `computeAtomizerStats` on paragraph-less orphan atoms.

Assertions are semantic — typed error names, diagnostic codes + locators, emitted OOXML markup, and `stats` shapes — not snapshot- or existence-only.

## Verification (independently re-run by supervisor)
- [x] `pipeline.ts` coverage 47.0% → **89.5%** (672/751 statements)
- [x] Full `@usejunior/docx-compare` suite: 43 files, **651 passed**, 24 skipped
- [x] eslint clean; `check:allure-labels` exit 0 (features: Pipeline Auxiliary Note Merge / Pipeline Comment Ancillary Merge / Pipeline Safety Guard Diagnostics)
- [ ] Peer review (agy + Codex) — pending; appended below

## Remaining uncovered
~79 lines: rare diagnostic-detail combinations, forced `ContainerResolutionError` recovery, rebuild round-trip failure, malformed-comment rethrow, and defensive early returns — candidates for a later targeted pass.

## Closes
- #636

https://claude.ai/code/session_01P4C7hFonxn6KXJqb6GF1Th

PR #633, opened 2026-07-24
test(docx-compare): cover documentReconstructor provenance + structure paths

stevenobiajulu:
## Summary
Characterization tests for previously-uncovered branches of the atomizer `documentReconstructor`, lifting its statement coverage **84.3% → 88.9%**. Test-only; no production code changes.

## What's covered
**`documentReconstructor-provenance.test.ts`** — the issue-#358 rebuild-side original-`w:ins` provenance threading. When merged atoms whose ORIGINAL lineage sat inside a pre-tracked `<w:ins>` are reconstructed, the deletion must nest as `<w:ins original-author><w:del>…</w:del></w:ins>` so reject-all drops the content with the original insertion (INV-RT-001) while accept-all unwraps the emptied insertion. Exercises `partitionAtomsByInsProvenance`, `buildRunGroupXmlWithInsProvenance`, and the whole-paragraph provenance branch of `buildParagraphXml`. These are reachable only by handing `reconstructDocument` atoms carrying an original `w:ins` `revTrackElement` — no pipeline-driven sibling test produced them.
- whole-paragraph delete of pre-tracked-inserted text → nested `w:ins`/`w:del`
- a provenance boundary splits one run group into separate `w:del` spans (inline path)
- same-author atoms collapse into a single restored insertion span

**`documentReconstructor-structure.test.ts`** — structural helpers:
- legacy exported `buildDocument` splicer (happy path + `Could not find w:body` throw)
- `reconstructDocument` no-`w:body` guard throw
- `shouldStartNewParagraph` / `shouldStartNewRunGroup` boundary fallbacks (undefined `paragraphIndex`; differing `moveName`)

## Verification
- [x] Both new files green (8 tests)
- [x] Full `@usejunior/docx-compare` suite: 40 files, 644 passed, 24 skipped
- [x] `documentReconstructor.ts` coverage: 84.3% → 88.9% statements
- [x] eslint clean; `check:allure-labels` exit 0; `check:allure-quality` 0 errors

## Notes
Remaining uncovered lines are mostly defensive `OpaquePassthroughError` / body-block validation throws and specialized opaque-scaffold / hyperlink-resolver-absent paths that need elaborate malformed fixtures — candidates for a targeted follow-up. Two lines the analysis flagged as dead (a both-null rPr fast-path shadow and an empty-atoms guard) were intentionally not chased.

https://claude.ai/code/session_01P4C7hFonxn6KXJqb6GF1Th

PR #642, opened 2026-07-25
test(docx-compare): raise inPlaceModifier coverage 74.5% → 86.5% (refs #639)

stevenobiajulu:
Refs #639.

Adds four characterization test files for the INPLACE reconstruction handlers in `packages/docx-compare/src/baselines/atomizer/inPlaceModifier.ts`. **Test-only — no production code changed.**

## Coverage

`inPlaceModifier.ts` statements: **231/310 (74.5%) → 268/310 (86.5%)**, +12.0pp.

Completes the atomizer coverage sweep alongside #622 (auxiliaryIdCollision 100%), #630 (inPlaceModifier handlers), #633 (documentReconstructor 89.1%), and #637 (pipeline 89.5%).

## What is covered

| File | Scenarios |
|---|---|
| `inPlaceModifier-bookmark-boundaries.test.ts` | Deleted paragraph whose text is enclosed by bookmark boundary markers; nested trailing `bookmarkEnd`s |
| `inPlaceModifier-empty-and-cell-paragraphs.test.ts` | Paragraph-mark-only insert/delete of empty paragraphs; removed first table-cell paragraph recreated after `w:tcPr` |
| `inPlaceModifier-moved-bookmark-paragraph.test.ts` | Styled bookmarked move source under `detectMoves` |
| `inPlaceModifier-original-insertion-provenance.test.ts` | Restoration of an original `w:ins` wrapper around text the revision left untracked |

Assertions are semantic — emitted OOXML elements, document order, and `result.stats`. No snapshots. Every helper asserts `reconstructionModeUsed === 'inplace'`, which both reviewers confirmed is load-bearing: mutating the handlers trips the `leanXmlVerifier` guard and falls back to rebuild, failing the test.

## A defect this pins: #641

The bookmark-boundary tests characterize **known-defective** behavior. Deleting a bookmarked paragraph hoists `<w:bookmarkStart>`/`<w:bookmarkEnd>` out to body level, where they collapse into a zero-length range detached from the text they named:

```xml
<w:bookmarkStart w:id="41" w:name="DeletedBoundary"/>
<w:bookmarkEnd w:id="41"/>
<w:p><w:pPr>…</w:pPr><w:del …><w:r><w:delText>Bookmarked</w:delText></w:r>…</w:del></w:p>
```

This is schema-legal (`CT_Body → EG_BlockLevelElts → EG_ContentBlockContent → EG_RunLevelElts`), so no validator flags it — the loss is semantic, and a cross-reference to the bookmark now resolves to an empty span. Filed as #641. The test names, a file-level comment, and the BDD step text all say this is a defect being pinned rather than endorsed; when #641 is fixed those tests must be rewritten to assert the enclosing range. Notably the move-source path already gets this right, which this PR asserts.

## Peer review

Both reviewers ran dynamic reviews (built, ran the suite, traced branches, mutation-tested) and initially returned **REQUEST CHANGES**. Both findings were real and are fixed here:

1. **Vacuous source-fidelity assertion.** The move test checked `expect(xml).toContain('<w:keepNext/>')` against the whole document. Because the move *destination* carries the same `w:pPr`, the check passed off the destination — Codex proved it by deleting the recreated source's `w:pPr` and watching the test still pass. Assertions are now scoped to the paragraph enclosing `<w:moveFrom>` via an `enclosingParagraph` helper, plus the bookmark range is asserted to stay *inside* the source and the source position is anchored between the leading and trailing paragraphs. Re-running that exact mutation now fails the test.
2. **Defect pinned under a name that understated it.** Test names and step text now lead with `DEFECT #641`, and #641 was filed.

## Verification

| Check | Result |
|---|---|
| `inPlaceModifier.ts` coverage | 74.5% → **86.5%** |
| `npx vitest run --root packages/docx-compare` | 47 files passed, 661 passed / 48 skipped |
| eslint | clean |
| `check:allure-labels` | pass |
| Production diff | none |

Remaining uncovered lines are defensive fallbacks requiring internally inconsistent atom/context states that valid `compareDocuments` inputs do not produce.

https://claude.ai/code/session_01P4C7hFonxn6KXJqb6GF1Th

### DIFF
--- a/packages/docx-compare/src/baselines/atomizer/inPlaceModifier-bookmark-boundaries.test.ts
+++ b/packages/docx-compare/src/baselines/atomizer/inPlaceModifier-bookmark-boundaries.test.ts
@@ -1,0 +1,129 @@
+import JSZip from 'jszip';
+import { describe, expect } from 'vitest';
+import { compareDocuments } from '../../index.js';
+import { testAllure, type AllureBddContext } from '../../testing/allure-test.js';
+import { buildDocxFromBodyXml } from '../../testing/ooxml-fixtures.js';
+
+const test = testAllure
+  .epic('Document Comparison')
+  .withLabels({ feature: 'Inplace Deleted Paragraph Bookmark Boundaries' });
+
+async function compareInplace(originalBody: string, revisedBody: string) {
+  const original = await buildDocxFromBodyXml(originalBody);
+  const revised = await buildDocxFromBodyXml(revisedBody);
+  const result = await compareDocuments(original, revised, {
+    engine: 'atomizer',
+    reconstructionMode: 'inplace',
+  });
+  expect(result.reconstructionModeUsed).toBe('inplace');
+
+  const documentPart = (await JSZip.loadAsync(result.document)).file('word/document.xml');
+  if (!documentPart) throw new Error('comparison result omitted word/document.xml');
+  return { result, xml: await documentPart.async('string') };
+}
+
+/**
+ * Characterization of KNOWN-DEFECTIVE behavior — see issue #641.
+ *
+ * When a deleted paragraph's text was enclosed by bookmark boundary markers, the markers are
+ * hoisted out of the paragraph to body level and collapse into a zero-length range that no
+ * longer covers the text it named. The markers survive by name only; a cross-reference to the
+ * bookmark resolves to an empty span.
+ *
+ * These tests pin what the engine does today so the behavior cannot change unnoticed. They do
+ * NOT endorse it. When #641 is fixed, they must be rewritten to assert that the bookmark range
+ * still encloses the recreated `<w:del>` content — the way the move-source path already does
+ * (see inPlaceModifier-moved-bookmark-paragraph.test.ts).
+ *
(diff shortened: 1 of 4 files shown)


stevenobiajulu:
Heads-up on a cross-PR interaction: #644 fixes the defect this PR's
`inPlaceModifier-bookmark-boundaries.test.ts` pins as characterization (#641).

The file is new on this branch, so there is no textual merge conflict — but the
two `DEFECT #641: …` tests fail against #644's fix. I verified by running this
branch's copy of that file against #644:

```
× DEFECT #641: hoists both boundary markers to body level, detaching the bookmark from the deleted text
× DEFECT #641: hoists nested trailing bookmark ends ahead of the deleted fragments, preserving only their relative order
  AssertionError: expected 831 to be less than 593
  AssertionError: expected 1003 to be less than 958
```

Both were written to be flipped once the bug was fixed, so whichever PR lands
second needs the update: the markers now sit *inside* the recreated paragraph
(`pPr, bookmarkStart, w:del, bookmarkEnd`) rather than at body level ahead of it.

#644 is green on all gates; this PR's CI is currently red, so #644 is going in
first unless you'd rather sequence it the other way.

Issue #643, opened 2026-07-25
docx-compare: a bookmark boundary partway inside a revision wrapper cannot be repositioned faithfully without splitting the wrapper


stevenobiajulu:
## Summary

`enforceConsumerCompatibility` lifts bookmark boundary markers out of inline
revision wrappers (`w:ins` / `w:del` / `w:moveFrom` / `w:moveTo`), because a
marker nested inside `<w:del>` disappears on Accept All even though its
paragraph survives.

When the wrapper owns the **whole** range, that is exact: start before the
wrapper, end after it, range unchanged (fixed in #641).

When only **one** boundary is inside the wrapper, no placement outside the
wrapper reproduces the original span, because the boundary sat partway through
content that one projection removes:

```xml
<w:p><w:bookmarkStart w:id="50" w:name="Partial"/>
  <w:r><w:t>kept</w:t></w:r>
  <w:del w:id="1">
    <w:r><w:delText>inside-range</w:delText></w:r>
    <w:bookmarkEnd w:id="50"/>
    <w:r><w:delText>outside-range</w:delText></w:r>
  </w:del></w:p>
```

The range covers `kept` + `inside-range`. Placing the end **before** the wrapper
shrinks it to `kept`; placing it **after** grows it to
`kept` + `inside-range` + `outside-range`. Current behavior shrinks (the
long-standing placement, kept deliberately in #641 so that fix introduced no new
inaccuracy).

## Expected

Split the wrapper at the boundary, copying its attributes, so the range keeps
its original span:

```xml
<w:del w:id="1"><w:r><w:delText>inside-range</w:delText></w:r></w:del>
<w:bookmarkEnd w:id="50"/>
<w:del w:id="N"><w:r><w:delText>outside-range</w:delText></w:r></w:del>
```

The post-processing passes that merge adjacent revision siblings
(`mergeAdjacentTrackChangeSiblings`, `coalesceDelInsPairChains`,
`mergeWhitespaceBridgedTrackChanges` in `inPlaceModifier-postprocess.ts`) must
then not re-merge across that boundary, or the split is undone.

## Scope

Both reconstruction modes, since they share the pass:
`packages/docx-compare/src/baselines/atomizer/consumerCompatibility.ts`
(`liftMarkersAroundWrapper`).

## Status

Current behavior is pinned as characterization in
`packages/docx-compare/src/baselines/atomizer/consumerCompatibility-bookmark-ranges.test.ts`
[...]

PR #644, opened 2026-07-25
fix(docx-compare): keep a deleted paragraph's bookmark range around its content (closes #641)

### LINKED ISSUE #641: docx-compare: deleting a bookmarked paragraph detaches its bookmark (markers hoisted to body level)

stevenobiajulu:
## Problem

Deleting a bookmarked paragraph detached its bookmark: both boundary markers
were lifted out to body level, adjacent to each other. That is schema-legal, so
no validator objected, but it collapses the range to zero length — the bookmark
survives by name and stops covering any text, so a `REF` / `PAGEREF` aimed at it
resolves to an empty span.

## Root cause

Not in the atomizer's paragraph-recreation code, which places the markers inside
the recreated paragraph correctly. The detaching happens afterwards, in
`enforceConsumerCompatibility` — the post-reconstruction pass that **both**
`inplace` and `rebuild` share, so both modes were affected and both are fixed
here. Two defects in its bookmark step:

1. **Paragraph-level lift was unconditional.** Every boundary of a wholly
   inserted or deleted `w:p` was moved to body level.
2. **Wrapper lift dropped both boundaries in front of the wrapper.** Same
   collapse in miniature. This is why the rebuild path still emitted a
   zero-length range even after fixing (1) — its reconstructor emits the markers
   inside the `w:del`.

## Fix

The lift now depends on how much of a range the container owns:

- A range with **both** boundaries inside the revised paragraph stays inside it,
  wrapped around the revision marker. It covers only that paragraph's text and
  is meant to travel with it.
- A range with one boundary in **surviving** content still gets that boundary
  anchored outside, because the paragraph vanishes in one projection. (Guarded by
  a test that passes both before and after, so it catches over-fixing.)
- Inside an inline wrapper, a range the wrapper **encloses** is placed around it
  — start before, end after. A boundary only **partway** inside keeps the
  long-standing placement ahead of the wrapper; see Known limitation.

Output now matches the issue's Expected, in both modes:

```xml
<w:p><w:pPr><w:keepNext/><w:rPr><w:del .../></w:rPr></w:pPr>
  <w:bookmarkStart w:id="41" w:name="DeletedBoundary"/>
  <w:del ...><w:r><w:delText>Bookmarked deleted text</w:delText></w:r></w:del>
  <w:bookmarkEnd w:id="41"/></w:p>
```

Reject All restores the enclosing range. Accept All leaves the emptied bookmark
at the paragraph merge point instead of detached at body level.

## Known limitation (issue #643)

A boundary sitting partway inside a revision wrapper cannot be repositioned
faithfully by moving it: before the wrapper shrinks the range, after it grows
it. Repositioning it exactly means splitting the wrapper at the boundary, which
is a distinct change — filed as #643 and pinned here as characterization, so
this PR introduces no new inaccuracy of its own.

## Bookmark-name-on-accept is policy, not measured Word behavior

An earlier draft of the tests claimed Word leaves a zero-length bookmark when
all of a range's text is deleted. No Word-generated oracle backs that, and the
repo's LibreOffice oracle drops the bookmark outright on accept. Preserving the
name is safe-docx policy that this pass already had, unchanged here; the test
comments now say so rather than attributing it to a consumer. Capturing real
Word output for this scenario is worth doing separately.

## Interaction with #642

#642 (`cover-inplacemodifier2-20260725`) adds
`inPlaceModifier-bookmark-boundaries.test.ts`, which pins the old hoisting as
characterization. It is a **new file**, so there is no textual merge conflict
either way — but its two `DEFECT #641: …` tests fail against this fix (verified
by running that file against this branch). They were written to be flipped:
"When this issue is fixed, those tests must be updated to assert the enclosing
range." Whichever PR lands second needs that flip.

## Verification

- 11 new tests (`consumerCompatibility-bookmark-ranges.test.ts`): deleted,
  inserted, table-cell, and spanning paragraph cases across both reconstruction
  modes, plus wrapper-boundary cases for `w:del`, `w:moveFrom`, nested
  `<w:ins><w:del>`, and the partial boundary.
- Red/green: 8 fail against the parent commit; the 3 that pass in both states are
  the deliberate guards.
- Full suite green. Also `lint:workspaces`, `check:spec-coverage`,
  `check:conformance-citations`, `check:conformance-doc`, `check:allure-labels`,
  `check:allure-quality`, `check:allure-filenames`, `check:bdd-coverage`,
  `check:test-narratives`, `check:cycles`, `check:control-bytes`,
  `check:emitted-schema`, and the coverage ratchet (docx-compare lines +27.37%).

Peer-reviewed by Codex, which ran the fixtures rather than reading them; both of
its blocking findings are addressed above.

Closes #641

https://claude.ai/code/session_01NNSDJd4kNg5wgBnKmL69WS

### DIFF
--- a/packages/docx-compare/src/baselines/atomizer/consumerCompatibility-bookmark-ranges.test.ts
+++ b/packages/docx-compare/src/baselines/atomizer/consumerCompatibility-bookmark-ranges.test.ts
@@ -1,0 +1,242 @@
+/**
+ * Bookmark ranges stay attached to the content they name.
+ *
+ * The consumer-compatibility pass repositions `w:bookmarkStart` /
+ * `w:bookmarkEnd` so a range survives both the Accept All and the Reject All
+ * projection. It used to lift both boundaries of a wholly deleted or inserted
+ * paragraph out to body level, which is schema-legal but collapses the range to
+ * a zero-length span: the bookmark keeps its name and stops covering any text,
+ * so a `REF` / `PAGEREF` field aimed at it resolves to nothing.
+ *
+ * A range whose two boundaries both sit inside the revised paragraph now stays
+ * inside it, wrapped around the revision marker. A range with one boundary in
+ * surviving content still gets that boundary anchored outside, because the
+ * paragraph disappears in one of the two projections.
+ *
+ * @conformance ECMA-376 edition 5, Part 1 § 17.13.6.1
+ * @conformance ECMA-376 edition 5, Part 1 § 17.13.6.2
+ * @see https://github.com/UseJunior/safe-docx/issues/641
+ */
+
+import { describe, expect } from 'vitest';
+import { DocxArchive, childElements, parseXml } from '@usejunior/docx-core';
+import { buildDocxFromBodyXml } from '../../testing/ooxml-fixtures.js';
+import { testAllure, type AllureBddContext } from '../../testing/allure-test.js';
+import { compareDocumentsAtomizer } from './pipeline.js';
+import { acceptAllChanges, rejectAllChanges } from './trackChangesAcceptorAst.js';
+import type { ReconstructionMode } from '../../compare-types.js';
+
+const TEST_FEATURE = 'Consumer Compatibility Bookmark Ranges';
+
+const test = testAllure
+  .epic('Document Comparison')
+  .withLabels({
+    feature: TEST_FEATURE,
+    story: 'Bookmark Range Stays Around Revised Content',
+    severity: 'critical',
+  })
+  .conformance(
+    { spec: 'ECMA-376', edition: 5, part: 1, section: '17.13.6.1' },
+    { spec: 'ECMA-376', edition: 5, part: 1, section: '17.13.6.2' },
+  );
+
+const AUTHOR = 'Bookmark Range Test';
+const DATE = new Date('2026-07-25T00:00:00Z');
+
+function textParagraph(text: string): string {
+  return `<w:p><w:r><w:t>${text}</w:t></w:r></w:p>`;
+}
+
+/** Paragraph whose whole text is enclosed by one bookmark range. */
+function bookmarkedParagraph(id: string, name: string, text: string, pPr = ''): string {
+  return (
+    `<w:p>${pPr}` +
+    `<w:bookmarkStart w:id="${id}" w:name="${name}"/>` +
+    `<w:r><w:t>${text}</w:t></w:r>` +
+    `<w:bookmarkEnd w:id="${id}"/>` +
+    `</w:p>`
+  );
+}
+
+async function compare(
+  originalBody: string,
+  revisedBody: string,
+  reconstructionMode: ReconstructionMode
+): Promise<string> {
+  const original = await buildDocxFromBodyXml(originalBody);
+  const revised = await buildDocxFromBodyXml(revisedBody);
+  const result = await compareDocumentsAtomizer(original, revised, {
+    author: AUTHOR,
+    date: DATE,
+    reconstructionMode,
+  });
+  expect(result.reconstructionModeUsed).toBe(reconstructionMode);
+  return (await DocxArchive.load(result.document)).getDocumentXml();
+}
+
+function bodyOf(documentXml: string): Element {
+  const doc = parseXml(documentXml);
+  const body = doc.getElementsByTagName('w:body')[0];
+  if (!body) throw new Error('no w:body in document');
+  return body as Element;
+}
+
+/** Tag names of the direct children of `w:body`, in document order. */
+function bodyChildTags(documentXml: string): string[] {
+  return childElements(bodyOf(documentXml)).map((child) => child.tagName);
+}
+
+/**
+ * Tag names of the direct children of the paragraph carrying `text`, in
+ * document order, with runs of the same tag collapsed to one entry. Word-level
+ * atomization splits a sentence across several sibling `w:r`, and how many it
+ * produces is beside the point here.
+ */
+function paragraphChildTags(documentXml: string, text: string): string[] {
+  const body = bodyOf(documentXml);
+  for (const paragraph of Array.from(body.getElementsByTagName('w:p')) as Element[]) {
+    if (paragraph.textContent?.includes(text)) {
+      return childElements(paragraph)
+        .map((child) => child.tagName)
+        .filter((tag, index, tags) => tag !== tags[index - 1]);
+    }
+  }
+  throw new Error(`no paragraph containing ${JSON.stringify(text)}`);
+}
+
+const ORIGINAL_WITH_BOOKMARKED_PARAGRAPH =
+  textParagraph('Leading survivor') +
+  bookmarkedParagraph('41', 'DeletedBoundary', 'Bookmarked deleted text', '<w:pPr><w:keepNext/></w:pPr>') +
+  textParagraph('Trailing survivor');
+
+const REVISED_WITHOUT_BOOKMARKED_PARAGRAPH =
+  textParagraph('Leading survivor') + textParagraph('Trailing survivor');
+
+describe('Bookmark ranges survive paragraph-level revisions', () => {
+  for (const mode of ['inplace', 'rebuild'] as const) {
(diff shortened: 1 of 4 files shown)


stevenobiajulu:
## Post-merge smoke: PASS

Merge commit `562296f0e4181b556c372d36eb730dbb7e1edb22` (merged 2026-07-25T23:48:02Z). Smoked on a
clean build of that commit: all 7 prebuilt `packages/*/dist` directories deleted, then
`npm run build` from scratch (exit 0).

### Real-world fixtures

Untouched, SHA-256-verified NVCA model agreements fetched via `ensureSourceDocx` and cached at
`~/.open-agreements/cache/<recipe-id>/source.docx` — real Word-authored legal documents, not
constructed fixtures. For each one the smoke located a clause whose text is bracketed by a bookmark
range that is both (a) the live target of a `REF`/`PAGEREF` cross-reference elsewhere in the
agreement and (b) enclosing every text-bearing run in its paragraph — the issue #641 shape — then
**struck that clause**, an authentic lawyer redline.

| fixture | paras | struck clause | bookmark | inplace |
|---|---|---|---|---|
| `nvca-certificate-of-incorporation` | 234 | "a subsidiary of the Corporation is a constituent party…" | `_Ref175118432` | ✅ |
| `nvca-investors-rights-agreement` | 420 | "in the event of any underwritten public offering…" | `_Ref42252487` | ✅ |
| `nvca-rofr-co-sale-agreement` | 199 | "Shares Includable. Each Participating Investor may include…" | `_Ref264907037` | ✅ |
| `nvca-stock-purchase-agreement` | 405 | "[__________] shares of common stock, $[____] par value…" | `_Ref145428096` | ✅ |
| `nvca-voting-agreement` | 180 | "[As a Preferred Director, [Example 1: one person designated…" | `_Ref146109502` | ✅ |

Per fixture, asserted on the emitted `word/document.xml`:

- the struck clause is recorded as a deletion, `reconstructionModeUsed === 'inplace'`
- **both boundary markers survive**, located by `w:name` and re-paired on whatever `w:id` the start
  ended up carrying (`enforceConsumerCompatibility` may reassign ids)
- **the range stays non-empty**: `bookmarkStart` < `w:delText` < `bookmarkEnd`
- the deleted clause text actually sits *between* the two markers
- **both markers sit inside one `w:p`, not hoisted to body level** — the #641 defect
- `REF` fields targeting the bookmark survive

**Schema validation.** All 5 emitted documents validate against the vendored ECMA-376 Transitional
WML schema via `scripts/check_emitted_document_schema.mjs --self-test` (exit 0). The self-test passed
first, so the harness is proven to both accept a known-good and reject a known-bad instance rather
than silently passing everything.

**Full suite** on the clean build: exit 0 — 295 test files, 3,118 passed, 0 failures, 2 expected-fail
(characterization pins), 97 skipped, across all 8 packages.

### Coverage gap I could not close: rebuild mode on real documents

The fix lives in the post-reconstruction pass shared by both modes, so unit tests cover
`rebuild`. **But rebuild could not be exercised on any real fixture**: it rejects a single-paragraph
deletion in every one of these documents with
`OpaquePassthroughError: boundary 0 changed paragraph ownership, moved, or mutated`. That guard is
pre-existing and edit-triggered, not document-triggered — a zero-edit rebuild self-compare passes,
and a plain `WHEREAS` clause with no bookmark and no field code fails identically. Filed as **#646**.
Real-document evidence for this fix is therefore **inplace-only**; rebuild rests on unit coverage.

### Other pre-existing findings surfaced (both unrelated to this PR)

- **#645** — `validateFieldStructure` rejects Word's nested-field serialization
  (`IF` guarding a `DOCPROPERTY`, a standard law-firm DMS footer stamp). Word writes the inner
  field's cached result as `w:instrText` because the *outer* field is still in its code region, but
  the check only inspects the innermost level. This makes `nvca-indemnification-agreement`
  **uncomparable in either mode** — its zero-edit self-compare throws. Includes a secondary defect:
  the inplace→rebuild fallback wraps only the first `assembleCandidate`, so a both-modes ancillary
  failure escapes uncaught.
- **#643 confirmed on real documents.** The partial-boundary limitation this PR deliberately left in
  place is not a synthetic edge case: `nvca-voting-agreement` has 15 cross-referenced ranges that end
  mid-paragraph (e.g. `_Ref444624639` covers only the "Board Composition" heading, with
  `bookmarkEnd` sitting between runs). Those still collapse to zero length on deletion, exactly as
  #643 documents. Counts per fixture: certificate-of-incorporation 3, stock-purchase 1,
  voting-agreement 15; the other three have none.

### Advisory-gate audit

Not trusting the green checkmark: the centralized LLM gate's posted verdict reads
`Overall: PASS (5 pass, 0 warn, 0 error, 12 skipped, 17 total)` with substantive per-check reasoning
(the unit-test-quality check specifically noted the tests "are mock-free and assert precise XML tag
structures constructed from first principles"). No `Unable to verify` / `provider failed` /
`Gemini CLI exited` markers — the gate genuinely ran. Codecov reported patch coverage 93.55% with 2
partials in `consumerCompatibility.ts` (non-blocking).

### Notes

- **Visual evidence: skipped** — bookmark marker placement is structurally invisible on a rendered page.
- **Prod: n/a** — safe-docx is a library; the repo does not deploy on merge.
- Artifacts (emitted + revised DOCX, suite log, schema-gate output) at
  `/tmp/automerge-smoke-644-20260725-2350/`. Not uploaded here.
- The PR body's note about #642 ("whichever PR lands second needs that flip") is stale — #642 landed
  first as `ca1666c`, and the characterization-test flip was done in this branch as `796414e`.

PR #659, opened 2026-07-26
fix(docx-core): accept nested field instruction results

### LINKED ISSUE #645: docx-core: validateFieldStructure rejects Word's nested-field serialization, making real documents uncomparable
## Summary

`validateFieldStructure` rejects the way Word serializes a **nested field**, so real Word-authored
documents whose headers/footers carry a nested field cannot be compared at all — the atomizer
throws `AncillaryStorySafetyError` for **inplace**, falls back to **rebuild**, fails there too, and
the second `assembleCandidate` call sits outside the try/catch, so the error escapes to the caller.

Reproduced during the post-merge smoke for #644 on a real fixture:
`~/.open-agreements/cache/nvca-indemnification-agreement/source.docx` (NVCA model Indemnification
Agreement). **A zero-edit self-compare throws**, so this is a property of the document, not of any edit:

```
compareDocuments(buf, buf, { engine: 'atomizer' })
  -> AncillaryStorySafetyError: Ancillary story safety check failed with 3 issue(s)
     code: INSTRUCTION_TEXT_OUTSIDE_FIELD_CODE
     detail: w:instrText must appear inside an open field code region
     parts: word/footer2.xml, word/footer4.xml, word/footer7.xml
```

## Root cause

`packages/docx-core/src/shared/field-structure.ts:254-257` flags `w:instrText` whenever
`pastSeparatorAtDepth[depth]` is set — i.e. whenever the *innermost* open field is past its
`separate`. But for a **nested** field, everything between the OUTER field's `begin` and its
`separate`/`end` is the outer field's instruction text, including the inner field's cached
result. Word therefore writes the inner field's result as `w:instrText`, and that is correct.

The real footer (verbatim from `word/footer2.xml`, a law-firm DMS stamp — `IF` guarding a
`DOCPROPERTY`):

```xml
<w:fldChar w:fldCharType="begin"/>              <!-- depth 1: IF          pastSep[1]=0 -->
<w:instrText>IF "0" = "1" "</w:instrText>       <!--                      legal          -->
  <w:fldChar w:fldCharType="begin"/>            <!-- depth 2: DOCPROPERTY pastSep[2]=0 -->
  <w:instrText> DOCPROPERTY "SWDocID" </w:instrText>
  <w:fldChar w:fldCharType="separate"/>         <!--                      pastSep[2]=1 -->
  <w:instrText>RLF1 23607329v.2</w:instrText>   <!-- FLAGGED — but depth 1 is still in its
                                                     code region, so this IS instruction text -->
  <w:fldChar w:fldCharType="end"/>
<w:instrText>" ""</w:instrText>
<w:fldChar w:fldCharType="end"/>
```

At the flagged node `depth=2, pastSep[2]=1` but `pastSep[1]=0`. The check looks only at the
innermost level.

## Suggested fix

`w:instrText` is inside a field code region when the **outermost enclosing field** is still in its
code region — not merely when the innermost one is. Guard on "some enclosing depth `d` has
`pastSeparatorAtDepth[d] === 0`" (equivalently: the outermost open field is pre-`separate`) rather
than on `pastSeparatorAtDepth[depth]` alone. The same reasoning applies to the
`DELETED_INSTRUCTION_TEXT_OUTSIDE_FIELD_CODE` branch at lines 262-264.

## Secondary defect

Even once the checker is right, the escape path is wrong: in
`packages/docx-compare/src/baselines/atomizer/pipeline.ts`, the inplace→rebuild fallback wraps only
the first `assembleCandidate` call. When rebuild *also* raises `AncillaryStorySafetyError`, the
second call (~line 1022) throws uncaught. An ancillary failure in both modes should surface as
diagnostics on the result, not as an exception, or at minimum carry both modes' issue lists.

## Impact

Nested `IF`/`DOCPROPERTY` footers are a very common document-management-system stamp in law-firm
documents, which is the core corpus. Any such document is currently uncomparable.

Found while running the post-merge smoke for #644 (issue #641). Unrelated to that change — neither
`field-structure.ts` nor `ancillaryFieldSafety.ts` is touched by it.

stevenobiajulu:
## Summary

- treat `w:instrText` / `w:delInstrText` as inside a field-code region when **any** enclosing field remains pre-separator
- mirror the corrected semantics in both Lean field validators and pin Lean↔TypeScript parity
- retain diagnostics from both inplace and rebuild ancillary-publication attempts on terminal failure
- add the canonical Word `IF` / `DOCPROPERTY` nested-field fixture and positive/negative regression coverage

This fixes Word-authored DMS footer fields where an inner field's cached result is serialized as `w:instrText` while the outer field is still in its instruction region.

## Validation

- full mandatory pre-submit sequence passed:
  - `npm run build`
  - `npm run lint:workspaces` (0 errors; 9 pre-existing warnings)
  - `npm run test:run`
- compiled Lean targets: `Tier2.FieldStructure`, `Tier2.XmlTripleChecker`, `leanDocxChecker`, and `leanHelperDifferential`
- Lean↔TypeScript differential: 12/12 passed with LibreOffice oracle
- real NVCA Indemnification Agreement zero-edit self-compare:
  - inplace: PASS, zero revisions
  - rebuild: PASS, zero revisions

One pre-existing move-emission test hit its 5-second timeout during the first heavily loaded full run; it passed in isolation (610 ms), and the complete workspace test command then passed on rerun.

Fixes #645

### DIFF
--- a/packages/docx-compare/src/baselines/atomizer/ancillaryFieldSafety.ts
+++ b/packages/docx-compare/src/baselines/atomizer/ancillaryFieldSafety.ts
@@ -46,13 +46,27 @@ const IGNORED_STRICT_ISSUES = new Set([
   'DELETED_TEXT_OUTSIDE_DELETION',
 ]);
 
+export interface AncillaryStorySafetyAttempt {
+  reconstructionMode: ReconstructionMode;
+  issues: AncillaryStorySafetyIssue[];
+}
+
 export class AncillaryStorySafetyError extends Error {
   readonly issues: AncillaryStorySafetyIssue[];
+  readonly attempts?: AncillaryStorySafetyAttempt[];
 
-  constructor(issues: AncillaryStorySafetyIssue[]) {
-    super(`Ancillary story safety check failed with ${issues.length} issue(s)`);
+  constructor(
+    issues: AncillaryStorySafetyIssue[],
+    attempts?: AncillaryStorySafetyAttempt[],
+  ) {
+    super(
+      attempts
+        ? `Ancillary story safety check failed in ${attempts.length} reconstruction attempt(s)`
+        : `Ancillary story safety check failed with ${issues.length} issue(s)`,
+    );
     this.name = 'AncillaryStorySafetyError';
     this.issues = issues;
+    this.attempts = attempts;
   }
 }
 
--- a/packages/docx-compare/src/baselines/atomizer/pipeline-safety-guards.test.ts
+++ b/packages/docx-compare/src/baselines/atomizer/pipeline-safety-guards.test.ts
@@ -174,8 +174,7 @@ describe('pipeline safety and input guards', () => {
   }: AllureBddContext) => {
     // The inplace attempt throws AncillaryStorySafetyError; the pipeline then
     // retries in rebuild mode, whose base archive still carries the corrupted
-    // footnotes.xml, so rebuild fails closed with the same typed diagnostic —
-    // which is what surfaces to the caller. We assert on that terminal error.
+    // footnotes.xml. The terminal error retains diagnostics from both attempts.
     let original: Buffer;
     let revised: Buffer;
     let failure: unknown;
@@ -218,6 +217,24 @@ describe('pipeline safety and input guards', () => {
               normalizedPartPath: 'word/footnotes.xml',
             }),
           }),
+        ],
+        attempts: [
+          {
+            reconstructionMode: 'inplace',
+            issues: [
+              expect.objectContaining({
+                code: 'NOTE_PART_XML_INVALID',
+              }),
+            ],
+          },
+          {
+            reconstructionMode: 'rebuild',
+            issues: [
+              expect.objectContaining({
+                code: 'NOTE_PART_XML_INVALID',
+              }),
+            ],
+          },
         ],
       });
     });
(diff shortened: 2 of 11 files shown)


stevenobiajulu:
Post-merge smoke: **PASS** ✅

Verified squash commit `02057cdd8d56282ace4f97c922dde7c4ebcacf1c` from a fresh `main` worktree.

- Installed lockfile-pinned dependencies with `npm ci --ignore-scripts`.
- Build and lint passed.
- Workspace tests completed green except for one load-sensitive 5 s timeout in the pre-existing advanced move-emission test; rerunning that exact test in isolation passed in **820 ms** (1 passed, 10 skipped). All remaining workspace suites passed.
- `npm run check:spec-coverage`, `npm run check:conformance-citations`, and `npm run check:conformance-doc` passed.
- Real-document acceptance: compared the genuine NVCA Indemnification Agreement source DOCX against itself in both explicit reconstruction modes.
  - `inplace`: requested/used `inplace`, all 10 comparison counters zero.
  - `rebuild`: requested/used `rebuild`, all 10 comparison counters zero.
- Consumer-open check: LibreOffice opened both output DOCX files and exported each to a valid **29-page PDF**.
- Production commit statuses: Vercel **success**, Codecov patch **success**.

No product regression observed.

Issue #650, opened 2026-07-26
CLI compare aborts with OpaquePassthroughError (w:sdt placement) on a document compare_documents handles fine


stevenobiajulu:
## Summary

On one real document (a ~430 KB, ~1,000-paragraph form package),
`safe-docx compare` aborts with an uncaught `OpaquePassthroughError`, while
`compare_documents` compares **the same pair** successfully and writes a redline
Word opens.

```
OpaquePassthroughError: Opaque passthrough: w:sdt placement is outside inline-run and direct body-block support
    at validateSdtNamespaceOwnership (packages/docx-compare/dist/baselines/atomizer/opaquePassthrough.js:482:19)
    at atomizeTree (packages/docx-compare/dist/atomizer.js:632:9)
    at runComparisonPass (packages/docx-compare/dist/baselines/atomizer/pipeline.js:463:42)
    at compareDocumentsAtomizer (packages/docx-compare/dist/baselines/atomizer/pipeline.js:634:34)
    at async Object.runCompareCommand [as compare] (packages/docx-mcp/dist/cli/commands/compare.js:35:20)
    at async Object.parseAsync (packages/docx-mcp/dist/cli/index.js:113:32)
    at async runCli (packages/docx-mcp/dist/cli/index.js:147:5)
```

Two separate problems: the `w:sdt` placement is unsupported, and the two entry
points disagree about whether it is fatal.

## Environment

- safe-docx `0.17.0`, `c33d86e`, Node v26, macOS

## No minimal repro yet

I built a minimal fixture with a body-level `w:sdt` (a `docPartObj` /
"Table of Contents" gallery block wrapping one paragraph) and it compares fine
through both entry points — so that placement is not the trigger.

The document that does trigger it contains, in the same file: content controls,
~50 `sectPr`, VML text boxes, tables, and a nested-field table of contents.
I cannot attach it. Happy to run any narrowing script against it and report back —
that is probably the fastest route to a fixture, and I would rather do that than
guess.

Suggested first probe: log the offending node's ancestor chain in
`validateSdtNamespaceOwnership` before throwing. Right now the message says which
placements are supported but not which one was found or where.

## Expected

1. Whatever `compare_documents` does to survive this, the CLI should do too.
2. The error should name the ancestor chain and a paragraph identifier, so the
   input can be narrowed without patching the library.
3. Preferably: passthrough rather than abort, with the affected region reported
   as unrepresented (cf. #648).

## Related

[...]

PR #662, opened 2026-07-26
fix(docx-compare): preserve table-scoped content controls

### LINKED ISSUE #660: Rebuild reconstruction rejects table-scoped w:sdt (w:tr / w:tc), removing the inplace safety fallback (root cause of #650)
## Summary

The rebuild reconstruction path rejects **table-scoped `w:sdt` placements** — content controls whose parent is `w:tr` (row-level) or `w:tc` (block-level inside a cell). Both are legal OOXML. Any document containing one can never be compared in `rebuild` mode, and — because `rebuild` is also the hard safety fallback for `inplace` — such a document has **no working fallback** when the inplace passes fail their round-trip safety check.

This is the root cause of #650. That issue frames the failure as CLI-only; it isn't. The CLI just reaches the rebuild path more often, because the library default is `reconstructionMode: 'rebuild'` (`pipeline.ts:599`) while the MCP tool hardcodes `'inplace'` (`docx-mcp/src/tools/comparison_defaults.ts:9`).

## Reproduction

Self-contained, no external fixture. Generates a 1.1 KB document pair whose only content control is a `w:tc` wrapped by `w:sdt` inside a table row:

```python
import os, zipfile
W = 'xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main"'

def doc(cell_text):
    return f'''<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<w:document {W}><w:body>
<w:p><w:r><w:t>Lead paragraph.</w:t></w:r></w:p>
<w:tbl>
<w:tblPr><w:tblW w:w="0" w:type="auto"/></w:tblPr>
<w:tr>
<w:sdt>
<w:sdtPr><w:alias w:val="CellControl"/><w:id w:val="101"/></w:sdtPr>
<w:sdtContent>
<w:tc><w:tcPr><w:tcW w:w="4000" w:type="dxa"/></w:tcPr>
<w:p><w:r><w:t>{cell_text}</w:t></w:r></w:p>
</w:tc>
</w:sdtContent>
</w:sdt>
<w:tc><w:tcPr><w:tcW w:w="4000" w:type="dxa"/></w:tcPr><w:p><w:r><w:t>Static cell.</w:t></w:r></w:p></w:tc>
</w:tr>
</w:tbl>
<w:p><w:r><w:t>Trailing paragraph.</w:t></w:r></w:p>
<w:sectPr><w:pgSz w:w="12240" w:h="15840"/></w:sectPr>
</w:body></w:document>'''

CT = '''<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<Types xmlns="http://schemas.openxmlformats.org/package/2006/content-types">
<Default Extension="rels" ContentType="application/vnd.openxmlformats-package.relationships+xml"/>
<Default Extension="xml" ContentType="application/xml"/>
<Override PartName="/word/document.xml" ContentType="application/vnd.openxmlformats-officedocument.wordprocessingml.document.main+xml"/>
</Types>'''
RELS = '''<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<Relationships xmlns="http://schemas.openxmlformats.org/package/2006/relationships">
<Relationship Id="rId1" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/officeDocument" Target="word/document.xml"/>
</Relationships>'''

for name, text in (('rowsdt-original.docx','Original cell text.'), ('rowsdt-revised.docx','Revised cell text.')):
    with zipfile.ZipFile(name, 'w', zipfile.ZIP_DEFLATED) as z:
        z.writestr('[Content_Types].xml', CT)
        z.writestr('_rels/.rels', RELS)
        z.writestr('word/document.xml', doc(text))
```

```js
import { readFileSync } from 'node:fs';
import { compareDocuments } from '@usejunior/docx-compare';

const o = readFileSync('rowsdt-original.docx');
const r = readFileSync('rowsdt-revised.docx');
for (const mode of ['rebuild', 'inplace']) {
  try {
    const res = await compareDocuments(o, r, { reconstructionMode: mode });
    console.log(`${mode}: OK (used=${res.reconstructionModeUsed ?? 'n/a'})`);
  } catch (e) {
    console.log(`${mode}: THREW ${e.constructor.name}: ${e.message}`);
  }
}
```

**Actual:**

```
rebuild: THREW OpaquePassthroughError: Opaque passthrough: w:sdt placement is outside inline-run and direct body-block support
inplace: OK (used=inplace)
```

**Expected:** both modes produce a redline; a legal cell- or row-scoped content control should not abort comparison.

## Root cause

1. `baselines/atomizer/pipeline.ts:723` — SDT passthrough capture is enabled only for rebuild:
   ```ts
   const effectiveAtomizeOptions = outputMode === 'rebuild'
     ? { ...atomizeOptions, captureInlineSdtPassthrough: true }
     : atomizeOptions;
   ```
2. `atomizer.ts:840` — that flag runs `validateSdtNamespaceOwnership(node)` over the **entire body**, not just changed regions.
3. `baselines/atomizer/opaquePassthrough.ts:530-538` (and the same test again in `captureSdtPassthrough` at `599-606`) — an SDT is accepted only when its parent is the enclosing `w:p` (inline run) or `w:body` (direct body block):
   ```ts
   const isInline = paragraph && parent === paragraph;
   const isBodyBlock = parent?.nodeType === 1 &&
     (parent as Element).namespaceURI === OOXML.W_NS &&
     (parent as Element).localName === 'body';
   if (!isInline && !isBodyBlock) {
     throw new OpaquePassthroughError('w:sdt placement is outside inline-run and direct body-block support');
   }
   ```

`w:sdt` parented by `w:tr` or `w:tc` matches neither and throws.

## Why it looks intermittent

On a real document the failure appears to depend on which edits are present, which is misleading. It doesn't. `reconstructionMode: 'inplace'` tries four passes (`pipeline.ts:844-884`) and only falls back to rebuild when all four fail the round-trip safety check (`pipeline.ts:934-937`). So for the same file pair:

- edits where some inplace pass survives → SDT validator never runs → comparison succeeds
- edits where all four fail → rebuild fallback → validator runs over the whole body → hard throw

The SDT placements are identical in both cases. I confirmed this on a ~1,000-paragraph document with 20 `w:sdt` elements (4 inline in `w:p`, 10 parented by `w:tr`, 6 parented by `w:tc`): identical placements in original and revised, comparison succeeding or failing purely according to which reconstruction path was taken.

## Impact

- Documents with table content controls — common in subscription agreements, intake forms, and any questionnaire-style contract — lose the rebuild fallback entirely.
- When inplace can't produce a safe result, `compare_documents` returns `COMPARE_ERROR` with no redline at all, rather than degrading.
- The guard itself is correct in spirit: rebuild reconstructs `document.xml` from atoms, so an SDT it cannot model would be silently dropped. The supported placement set is what's too narrow.

## Fix direction

`placementKind` is currently `'inline-run' | 'body-block'`. Supporting table-scoped controls needs:

- new placement kinds for cell-scoped and row-scoped controls;
- container-relative anchoring instead of `bodyChildOrdinal` (`opaquePassthrough.ts:635`), which assumes a `w:body` child index;
- corresponding handling in the rebuild reconstruction so the boundary is re-emitted inside the right `w:tr` / `w:tc`.

Row-level is the harder half: `w:sdtContent` there holds `w:tc` elements rather than paragraphs, so `validateBlockSdtKnownStructure`'s paragraph-ownership contract (`opaquePassthrough.ts:499-525`) doesn't apply and needs a cell-ownership analogue.

A smaller interim step worth considering on its own: when the rebuild fallback aborts, surface the placement diagnostic through the `compare_documents` result rather than only as `COMPARE_ERROR`, so callers can tell "unsupported content control" apart from a generic comparison failure.

## Environment

- `safe-docx` @ `c33d86e`
- Node v26.0.0
- macOS 15 (Darwin 25.4.0)

Related: #650 (symptom report — CLI framing should be corrected to point here).

stevenobiajulu:
## Summary

- preserve legal table-row (`w:tr`) and table-cell (`w:tc`) scoped structured document tags during rebuild reconstruction
- identify opaque controls relative to their owning table container while allowing tracked paragraph-content edits
- fail closed when the control wrapper or surrounding table scaffold changes
- register the exact ECMA-376 conformance sections and add accept/reject regression coverage

## Root cause

Rebuild reconstruction classified opaque `w:sdt` nodes only as inline runs or direct body blocks. Legal controls nested directly under a table row or cell therefore aborted rebuild reconstruction, which also removed the inplace safety fallback and caused the comparison failure reported in #650.

## Validation

The repository-required build, workspace lint, package test suites, spec coverage, conformance citation, conformance document, and strict OpenSpec checks were run on the dedicated worktree. Sandbox-only IPC/LibreOffice failures were rerun outside the sandbox and passed.

A post-merge real-world DOCX smoke test will be run against the merged build and its output posted here.

Fixes #660
Related #650

### DIFF
--- a/openspec/changes/preserve-block-sdt-on-rebuild/design.md
+++ b/openspec/changes/preserve-block-sdt-on-rebuild/design.md
@@ -18,20 +18,27 @@ payload such as DrawingML.
 - Goals: pair multiple or identical controls deterministically from local body
   placement and ownership, with precomputed/memoized group identity and
   deterministic complexity evidence.
-- Non-Goals: row/cell/nested/editable controls; controls outside `document.xml`;
-  footer or ancillary reconstruction; tables inside a supported control; rsids
-  outside the preserved subtree; fields; `sectPr`; arbitrary package parts.
+- Goals: preserve row- and cell-scoped control wrappers through their existing
+  table scaffold while allowing ordinary tracked paragraph reconstruction
+  inside `w:sdtContent`.
+- Non-Goals: nested controls; controls outside `document.xml`;
+  footer or ancillary reconstruction; tables inside a direct-body opaque
+  control; rsids outside the preserved subtree; fields; `sectPr`; arbitrary
+  package parts.
 
 ## Decisions
 
 ### Placement is explicit
 
 `OpaquePassthroughNode` records `placementKind` as `inline-run` or
-`body-block`. Inline descriptors retain one paragraph owner and paragraph-run
+`body-block`, plus `row-block` and `cell-block`. Inline descriptors retain one paragraph owner and paragraph-run
 emission. Body-block descriptors record their direct body-child ordinal and a
 closed, contiguous interval of global paragraph slots. A block descriptor is
 attached to every atom in every owned paragraph, but its subtree is emitted only
-by the body scaffold.
+by the body scaffold. Row/cell descriptors record their direct parent-child
+ordinal and a wrapper fingerprint that replaces controlled paragraphs with
+structural markers, so paragraph content can change without laundering wrapper
+or table/cell scaffold changes.
 
 ### Correlation is local and fail closed
 
@@ -72,20 +79,24 @@ also fail closed before correlation.
 ### Claims remain bounded
 
 Block structure is cited to ECMA-376 Part 1 §§17.5.2.29, 17.5.2.34, and
-17.5.2.38. Exact opaque preservation and package-part stability are SafeDocX
+17.5.2.38; cell-level structure inside a row is cited to §§17.5.2.32 and
+17.5.2.33. Exact opaque preservation and package-part stability are SafeDocX
 metamorphic invariants, not requirements imposed by ECMA-376.
 
 ## Risks / Trade-offs
 
-- An otherwise safe edit inside the control is rejected. This avoids silently
-  preserving stale controlled content or flattening unsupported structure.
-- Block controls containing tables are rejected in this slice because paragraph
-  slot ownership alone cannot represent table/cell placement safely.
+- An otherwise safe edit inside a direct-body opaque control is rejected. This
+  avoids silently preserving stale controlled content or flattening unsupported
+  structure.
+- Direct body controls containing tables remain rejected because their opaque
+  paragraph-slot ownership cannot represent nested table placement safely.
+- Row/cell controls allow paragraph edits but reject wrapper, cell-property, or
+  table-shape changes that the paragraph reconstructor cannot safely represent.
 - Footer SDTs remain outside reconstruction support; package-clone identity may
   be observed only as no-regression evidence.
 
 ## Migration Plan
 
-No API migration is required. Existing inline controls keep their current
-behavior. Newly supported direct body-level controls pass through only when the
-strict block ownership contract is satisfied.
+No API migration is required. Existing inline and direct-body controls keep
+their current behavior. Row/cell controls use the existing table scaffold only
+when their wrapper and structural identity correlate.
--- a/openspec/changes/preserve-block-sdt-on-rebuild/proposal.md
+++ b/openspec/changes/preserve-block-sdt-on-rebuild/proposal.md
@@ -14,10 +14,12 @@ even when the control itself is unchanged.
   and contiguous paragraph-slot ownership while retaining the inline contract.
 - Capture unchanged direct body-level block SDTs, pair them deterministically,
   and preserve the validated original subtree as one scaffold-owned block.
+- Correlate row- and cell-scoped SDT wrappers by their container-relative
+  scaffold placement while allowing their controlled paragraphs to rebuild.
 - Bind every relationship-namespace attribute in a block to a memoized package
   closure covering relationship metadata, normalized targets, internal part
   hashes, and recursively referenced XML-part relationships.
-- Reject mutation, movement, ownership loss, nesting, unsupported placement, or
+- Reject wrapper mutation, movement, ownership loss, nesting, unsupported placement, or
   correlation loss before any lossy rebuilt XML is emitted.
 - Replace the ILPA count-only measurement with forced-rebuild corpus evidence
   that makes real unrelated edits and validates the complete SDT subtree,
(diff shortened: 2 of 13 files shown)


stevenobiajulu:
## ✅ Post-merge feature smoke passed

Functional merge: `4b82b172dbf9869133929cc6dfb6516484f6bda9` (#662 squash)
Validated `main`: `752e000cf7a3b02f9e7ff2c71dfa88ba5e617f97` (includes the Allure-label follow-up from #664)

**Real-document end-to-end path**

- Built all workspaces from merged `main` with clean `dist/` outputs.
- Used the repository's real 170,765-byte ILPA limited partnership agreement, wrapping two existing table structures in legal table-row (`w:tr`) and table-cell (`w:tc`) scoped content controls.
- Ran the built production CLI in forced rebuild mode:
  - engine: `atomizer`
  - requested/used mode: `rebuild`
  - output: 211,374-byte DOCX
  - stats: 2 insertions, 2 deletions, 2 modifications
- Verified the emitted control parents are exactly `w:tr` and `w:tc`.
- Verified each control contains one `w:ins` and one `w:del`.
- Accept-all contains only the revised row/cell markers; reject-all contains only the original markers.
- `unzip -t` reports no errors for the output package.

**Suite evidence**

- New table-SDT regression file: 2/2 passed in the full run.
- Full workspace run exercised all packages: docx-core 1,246 passed; docx-compare 674 passed before two load-sensitive failures; docx-mcp 897 passed; Google Docs 116 passed; ODF 136 passed before four LibreOffice timeouts; remaining workspace suites passed.
- Both load-sensitive compare failures passed immediately in isolated single-worker reruns (real ILPA corpus in 12.6s; verifier process-group case in 414ms).
- Three of four timed-out LibreOffice oracle files passed serially. One ODF paragraph-roundtrip oracle still times out against local LibreOffice; it is unrelated to this DOCX comparison change and the production DOCX smoke does not invoke LibreOffice.

**Gate audit**

- LLM aggregate: PASS (6 pass, 0 warn, 0 error).
- Lean Build: success.
- The original CI run's missing Allure feature label was fixed and merged in #664; #664 CI is green.
- Advisory Codecov patch report on #662 remains 73.08% (21 changed lines not fully covered).

Visual evidence: n/a — structural OOXML comparison fix, no visual surface.
Prod: n/a — no production deployment required.
Log: `/tmp/automerge-smoke-662-20260726T1853.log`
Artifacts: `/private/tmp/safe-docx-660-smoke`

**Cleanup**

- Deleted local feature/follow-up branches.
- Deleted the recreated remote feature branch; GitHub auto-deleted the merged follow-up branch.
- Working tree is clean on `main`.

PR #667, opened 2026-07-27
fix(docx-core): support cached field result edits

### LINKED ISSUE #651: replace_text refuses any edit inside a cached field result, and reports "spans multiple runs" even for a single-run edit
## Summary

`replace_text` / `batch_edit` refuse **any** edit that touches a cached field
result:

```
Failed to apply batch edit: Edit spans multiple runs and intersects a field result.
This is currently unsupported in Safe-Docx TS.
```

Two problems:

1. **The message is wrong for the single-run case.** An edit entirely inside one
   `w:t` of the field result is refused with the same "spans multiple runs" text.
2. **There is no supported way to do it.** Updating the cached page numbers in a
   table of contents is ordinary work, and no other tool in the surface reaches them.

## Environment

- safe-docx `0.17.0`, `c33d86e`, Node v26, macOS

## Reproduction

```python
#!/usr/bin/env python3
"""A TOC-style paragraph: literal text, then a PAGEREF field whose cached result is a page number."""
import zipfile

NS = ('xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main" '
      'xmlns:r="http://schemas.openxmlformats.org/officeDocument/2006/relationships" '
      'xmlns:w14="http://schemas.microsoft.com/office/word/2010/wordml" '
      'xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" mc:Ignorable="w14"')
CT = ('<?xml version="1.0" encoding="UTF-8" standalone="yes"?>'
      '<Types xmlns="http://schemas.openxmlformats.org/package/2006/content-types">'
      '<Default Extension="rels" ContentType="application/vnd.openxmlformats-package.relationships+xml"/>'
      '<Default Extension="xml" ContentType="application/xml"/>'
      '<Override PartName="/word/document.xml" ContentType="application/vnd.openxmlformats-officedocument.wordprocessingml.document.main+xml"/>'
      '</Types>')
RELS = ('<?xml version="1.0" encoding="UTF-8" standalone="yes"?>'
        '<Relationships xmlns="http://schemas.openxmlformats.org/package/2006/relationships">'
        '<Relationship Id="rId1" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/officeDocument" Target="word/document.xml"/>'
        '</Relationships>')
SECT = ('<w:sectPr><w:pgSz w:w="12240" w:h="15840"/><w:pgMar w:top="1440" w:right="1440" '
        'w:bottom="1440" w:left="1440" w:header="720" w:footer="720" w:gutter="0"/></w:sectPr>')

def t(s): return f'<w:r><w:t xml:space="preserve">{s}</w:t></w:r>'

def field_para(cached_runs):
    return ('<w:p w14:paraId="40000001" w14:textId="40000001">'
            + t("Section One")
            + '<w:r><w:fldChar w:fldCharType="begin"/></w:r>'
              '<w:r><w:instrText xml:space="preserve"> PAGEREF _Toc1 \\h </w:instrText></w:r>'
              '<w:r><w:fldChar w:fldCharType="separate"/></w:r>'
            + cached_runs
            + '<w:r><w:fldChar w:fldCharType="end"/></w:r></w:p>')

def pack(path, body):
    z = zipfile.ZipFile(path, "w", zipfile.ZIP_DEFLATED)
    z.writestr("[Content_Types].xml", CT)
    z.writestr("_rels/.rels", RELS)
    z.writestr("word/_rels/document.xml.rels",
               '<?xml version="1.0" encoding="UTF-8" standalone="yes"?>'
               '<Relationships xmlns="http://schemas.openxmlformats.org/package/2006/relationships"/>')
    z.writestr("word/document.xml",
               f'<?xml version="1.0" encoding="UTF-8" standalone="yes"?><w:document {NS}><w:body>{body}</w:body></w:document>')
    z.close(); print("wrote", path)

pack("field-single-run.docx", field_para(t("8")) + SECT)          # cached result "8" in one run
pack("field-split-runs.docx", field_para(t("1") + t("2")) + SECT)  # cached result "12" across two runs
```

## Observed

| edit | target | result |
|---|---|---|
| `"8"` → `"9"` | `field-single-run.docx` | REFUSED — "Edit spans multiple runs and intersects a field result" |
| `"12"` → `"13"` | `field-split-runs.docx` | REFUSED — same message |
| `"One8"` → `"One9"` | `field-single-run.docx` | REFUSED — same message |

The first case spans exactly one run. The message is inaccurate there.

On a real table of contents the guard also behaves inconsistently in a way the
message does not explain: within one document some TOC entries accepted an edit
(a heading label, and one page number) while others were refused. My reading is
that the accepted ones were entries whose live `PAGEREF` field had been dropped
by earlier hand-editing, so the text was no longer inside a field result — but
nothing in the API lets a caller distinguish the two beforehand.

## Expected

Short term:

- Make the message accurate, and name what was hit: which field, and whether the
  span was inside the result, inside the instruction, or crossing a boundary.
- Expose the condition as a query (does this paragraph, or this offset range, sit
  inside a field result?) so a caller can branch before attempting the edit.

Longer term: support editing a cached field result. Word treats it as a cache and
rewrites it on F9, so a targeted edit is legitimate — the caller usually knows the
field will not be refreshed, or is deliberately setting the value a reader sees
before the next refresh.

## Related

- #645 (`validateFieldStructure` and Word's nested-field serialization)

stevenobiajulu:
## Summary

- fix half-open text range mapping when a cached result starts at a run boundary
- track nested complex-field ownership and allow replacements wholly inside one cached result, including split runs
- keep cross-field rewrites fail-closed with field-specific diagnostics
- cover both `replace_text` and `batch_edit` while preserving all `w:fldChar` markers

## Verification

- `npm run build`
- `npm run lint:workspaces`
- `npm run test:run`
- LibreOffice identity and PDF probes passed in the full suite

Fixes #651

### DIFF
--- a/packages/docx-core/src/primitives/text.test.ts
+++ b/packages/docx-core/src/primitives/text.test.ts
@@ -7,6 +7,11 @@ import { SafeDocxError } from './errors.js';
 import { createRevisionContext, createRevisionIdState } from './track-changes-emitter.js';
 import { revisionEvidence, revisionEvidenceCases } from '../testing/revision-evidence.js';
 import {
+  fldChar,
+  instrText,
+  resultText,
+} from '../testing/ooxml-fixtures.js';
+import {
   getParagraphRuns,
   getParagraphText,
   splitRunAtVisibleOffset,
@@ -159,6 +164,37 @@ describe('getParagraphRuns', () => {
       expect(runs).toHaveLength(1);
       expect(runs[0]!.text).toBe('InlineResult');
       expect(runs[0]!.isFieldResult).toBe(true);
+    });
+  });
+
+  test('restores the enclosing result identity after a nested field ends', async ({ given, then }: AllureBddContext) => {
+    testAllure.conformance({ spec: 'ECMA-376', edition: 5, part: 1, section: '17.16.18' });
+    let doc: Document;
+
+    await given('an outer result containing a complete nested field', () => {
+      doc = makeDoc(
+        `<w:p>` +
+        fldChar('begin') +
+        instrText(' IF 1 = 1 ') +
+        fldChar('separate') +
+        resultText('Outer A') +
+        fldChar('begin') +
+        instrText(' REF X ') +
+        fldChar('separate') +
+        resultText('Inner') +
+        fldChar('end') +
+        resultText('Outer B') +
+        fldChar('end') +
+        `</w:p>`,
+      );
+    });
+
+    await then('the outer runs share an identity distinct from the nested result', () => {
+      const runs = getParagraphRuns(firstParagraph(doc));
+      expect(runs.map((run) => run.text)).toEqual(['Outer A', 'Inner', 'Outer B']);
+      expect(runs[0]!.fieldResultId).toBe(runs[2]!.fieldResultId);
+      expect(runs[1]!.fieldResultId).not.toBe(runs[0]!.fieldResultId);
+      expect(runs.map((run) => run.fieldInstruction)).toEqual(['IF 1 = 1', 'REF X', 'IF 1 = 1']);
     });
   });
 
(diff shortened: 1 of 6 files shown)


stevenobiajulu:
Post-merge smoke: **PASS** on merged commit `9ddfb2f03d60290525b070d97c577db82f840aee`.

- Built the exact `origin/main` merge commit in a detached smoke worktree.
- Exercised the real Word-authored fixture `tests/test_documents/nvca-regression/source.docx` end-to-end through the built CLI/MCP path.
- Replaced the cached `REF _Ref137575642` field result `Exhibit A` → `Schedule A` in paragraph `_bk_f53731c23556`, then saved clean + tracked variants.
- Re-read the saved clean DOCX and confirmed the visible result is `Schedule A`.
- Confirmed field topology was preserved: 68 `begin`, 68 `separate`, 68 `end` markers before and after; all 14 matching `REF _Ref137575642` instructions remained.
- `unzip -t` and `xmllint --noout` passed on the saved document.
- LibreOffice headless opened and converted the saved DOCX to PDF successfully.

Issue #646, opened 2026-07-26
docx-compare: rebuild mode rejects any paragraph deletion in a document containing REF fields


stevenobiajulu:
## Summary

In **rebuild** reconstruction mode, deleting *any* paragraph from a document that contains
REF/PAGEREF fields is rejected by the opaque-passthrough guard:

```
OpaquePassthroughError: Opaque passthrough: boundary 0 changed paragraph ownership, moved, or mutated
```

This is not specific to the deleted paragraph — it fires even when the struck paragraph contains no
bookmark and no field code of its own. It appears to be triggered by the paragraph-ordinal shift that
any deletion causes, which reads as every downstream field boundary "changing paragraph ownership".

## Reproduction

Real fixture: `~/.open-agreements/cache/nvca-voting-agreement/source.docx` (NVCA model Voting
Agreement, 180 paragraphs, REF fields throughout).

```js
// (a) zero-edit self-compare in rebuild mode — PASSES
await compareDocuments(buf, buf, { engine: 'atomizer', reconstructionMode: 'rebuild' });
//   -> OK, reconstructionModeUsed = 'rebuild'

// (b) strike one plain WHEREAS clause: no bookmarkStart, no instrText inside it — THROWS
//     ("WHEREAS, the parties desire to enter into this Agreement...")
await compareDocuments(buf, revised, { engine: 'atomizer', reconstructionMode: 'rebuild' });
//   -> OpaquePassthroughError: boundary 0 changed paragraph ownership, moved, or mutated
```

(a) passing and (b) failing pins it to the **edit**, not the document.

## Breadth

Observed on every real NVCA fixture carrying REF fields, in rebuild mode, for a single-paragraph
deletion:

| fixture | rebuild verdict |
|---|---|
| `nvca-certificate-of-incorporation` | `boundary 0 changed paragraph ownership, moved, or mutated` |
| `nvca-indemnification-agreement` | `boundary 0 changed paragraph ownership, moved, or mutated` |
| `nvca-investors-rights-agreement` | `unsupported REF field instruction shape` |
| `nvca-rofr-co-sale-agreement` | `boundary 0 changed paragraph ownership, moved, or mutated` |
| `nvca-stock-purchase-agreement` | `boundary 0 changed paragraph ownership, moved, or mutated` |
| `nvca-voting-agreement` | `boundary 0 changed paragraph ownership, moved, or mutated` |

[...]

PR #671, opened 2026-07-27
test(docx-compare): gate real-corpus paragraph deletion

### LINKED ISSUE #658: docx-compare: no real-corpus paragraph-deletion gate — #643/#645/#646 all hid behind a green suite
## Problem

Three engine defects (#643, #645, #646) were all found by an ad-hoc script during the post-merge
smoke for #644, not by CI. At the moment they were found the full suite was green: **295 test files,
3,118 tests passed, 0 failures.** On the real NVCA corpus at the same commit:

- **1 of 7** documents is uncomparable in *either* reconstruction mode (#645)
- **6 of 6** REF-bearing documents reject an ordinary single-paragraph deletion in `rebuild` (#646)
- **19** real cross-referenced bookmark ranges collapse to zero length on deletion (#643)

Each of those has its own issue. This one is about the reason none of them was caught: there is no
gate that drives the atomizer over the real corpus with a **paragraph deletion**.

## What exists vs. what is missing

| | today |
|---|---|
| Real-document comparison tests | one: `packages/docx-core/src/integration/nvca-structural-regression.test.ts` |
| …its edit shape | `source.docx` vs `filled.docx` — placeholder **filling** (text replacement) |
| …its reconstruction modes | `inplace` only |
| …its breadth | a single committed fixture pair |
| Real-fixture tests in `packages/docx-compare` | **none** — the package that owns the engine |
| Any real-document test performing a **paragraph deletion** | **none** |

Paragraph deletion is the most common redline operation in this corpus and the operation all three
defects hinge on. Nothing exercises it against a real document, in either mode.

The synthetic unit tests are not the problem — they are precise and they pass. The problem is that
every one of them constructs the document it then feeds to the engine, so none of them can encounter
the shapes real Word output actually contains: nested `IF`/`DOCPROPERTY` footers (#645), heading-only
`_Ref`/`_Toc` bookmark pairs that end mid-paragraph (#643), or REF boundary bookkeeping across a
paragraph-ordinal shift (#646).

## Proposed gate

A matrix in `packages/docx-compare` over **cached corpus document × `{inplace, rebuild}` × edit
shape**, with paragraph deletion added to the edit shapes. Per cell, assert the comparison completes
and that bookmark ranges targeted by `REF`/`PAGEREF` survive the edit — `bookmarkStart` <
`w:delText` < `bookmarkEnd`, markers inside a `w:p` rather than at body level.

Known failures get **characterization pins** in the established G-case style, so the gate is green
today and goes red in *both* directions:

- a pinned cell that changes behaviour → regression
- a pinned cell that starts **passing** → the underlying fix landed; remove the pin

The second direction matters here specifically: #643, #645 and #646 are all expected to be fixed
independently, and without it the pins rot into permanently-accepted failures.

## Fixture sourcing — the part that needs care

The corpus lives uncommitted at `~/.open-agreements/cache/<recipe-id>/source.docx`, populated by
`ensureSourceDocx` with SHA-256 verification. Follow the precedent already in the repo:
`packages/docx-core/src/integration/cross-implementation-suite.test.ts` env-gates on
`DOCX_PLATFORM_TESTS_DIR`, emits a **logged** skip warning naming the missing variable, and has a
unit test asserting that warning mentions the variable (line 488). Do the same here, and have CI set
the variable so the gate actually runs.

A silent skip would reproduce the exact failure this issue exists to prevent.
`scripts/check_emitted_document_schema.mjs` already states the principle in its own header: *"a gate
that silently skips is not a gate."*

## Fold in: a latent instance of that antipattern

`packages/docx-core/src/integration/nvca-structural-regression.test.ts:19-25` handles missing
fixtures with `console.warn` followed by `return`, which leaves `sourceBuf`/`filledBuf` undefined —
so the `when` block throws rather than skipping. It never fires today because the fixture pair is
committed, so this is latent, not broken. Worth fixing while in the area, and worth *not* copying
into the new gate.

## Also worth folding in (minor)

The orphan-bookmark repair in `enforceConsumerCompatibility` reports via `console.warn` and silently
synthesizes a counterpart. It fires on real documents — observed on the NVCA Indemnification
Agreement: `[Consumer Compatibility] Warning: Orphaned bookmarkStart name=_Hlk45869003. Inserting
synthetic end.` A repair that mutates the document should surface as structured diagnostics on the
result, not only on stdout, so a gate like the one proposed here can assert on it.

## Scope note

This is deliberately a **coverage/gating** issue, not a bug report. The three defects are tracked
separately in #643, #645 and #646; fixing any of them does not close this, and this closing does not
depend on them being fixed — the pins encode their current behaviour either way.

stevenobiajulu:
## What changed

- add a SHA-256-pinned seven-document NVCA corpus manifest and downloader
- run a real paragraph-deletion matrix across inplace and rebuild reconstruction modes
- pin the exact current outcomes for #643, #645, and #646 in both directions
- add a dedicated fail-closed CI job that materializes and verifies the corpus
- make the committed NVCA structural fixture test fail clearly when its files are absent

## Why

Synthetic fixtures did not contain the Word-authored bookmark, REF/PAGEREF, and nested-field layouts that exposed #643, #645, and #646 after a fully green suite. This gate exercises the actual corpus operation that found all three defects and requires maintainers to remove a characterization pin when an underlying fix lands.

## Validation

- fresh public-source download and SHA-256 verification for all seven documents
- real-corpus paragraph-deletion matrix in both reconstruction modes
- docx-compare build and typecheck
- docx-compare full suite: 675 passed
- NVCA structural regression
- spec coverage and conformance checks
- actionlint and git diff --check

Fixes #658
Refs #643, #645, #646

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -408,6 +408,43 @@ jobs:
             packages/docx-core/allure-results/
             packages/docx-mcp/allure-results/
 
+  real-corpus-comparison:
+    name: Real-corpus paragraph deletion
+    runs-on: ubuntu-latest
+    needs: workspace-lint
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: 20
+          cache: npm
+      - name: Install dependencies (npm ci with retry)
+        run: |
+          for attempt in 1 2 3; do
+            echo "npm ci attempt ${attempt}/3"
+            if npm ci; then
+              exit 0
+            fi
+            if [ "$attempt" -lt 3 ]; then
+              sleep $((attempt * 20))
+            fi
+          done
+          echo "npm ci failed after 3 attempts"
+          exit 1
+      - name: Materialize SHA-256-verified comparison corpus
+        env:
+          SAFE_DOCX_REAL_CORPUS_DIR: ${{ runner.temp }}/open-agreements-cache
+        run: node scripts/prepare_real_comparison_corpus.mjs "$SAFE_DOCX_REAL_CORPUS_DIR"
+      - name: Run real-corpus paragraph-deletion gate
+        env:
+          SAFE_DOCX_REAL_CORPUS_DIR: ${{ runner.temp }}/open-agreements-cache
+          SAFE_DOCX_REAL_CORPUS_REQUIRED: "1"
+        run: >-
+          npm run test:run -w @usejunior/docx-compare --
+          src/integration/real-corpus-paragraph-deletion.test.ts
+
   mcpb-bundle:
     runs-on: ubuntu-latest
     needs: workspace-test
--- a/packages/docx-compare/src/integration/real-corpus-manifest.json
+++ b/packages/docx-compare/src/integration/real-corpus-manifest.json
@@ -1,0 +1,37 @@
+[
+  {
+    "id": "nvca-certificate-of-incorporation",
+    "sourceUrl": "https://nvca.org/wp-content/uploads/2025/10/NVCA-Model-COI-10-1-2025.docx",
+    "sha256": "d75600769c12724990de48149d7a2bb161f3522daa54b1783672f93697d87d29"
+  },
+  {
+    "id": "nvca-indemnification-agreement",
+    "sourceUrl": "https://nvca.org/wp-content/uploads/2021/12/NVCA-2020-Indemnification-Agreement.docx",
+    "sha256": "f9b61b4d99e245573de41d9469925b4332f871ec0e6ac6593055cdfe17825300"
+  },
+  {
+    "id": "nvca-investors-rights-agreement",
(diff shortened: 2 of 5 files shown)


stevenobiajulu:
Automerge is currently stalled on **Real-corpus paragraph deletion**. Three characterization cells that fail on the local Darwin/Node 26 runtime pass on the CI Linux/Node 20 runtime, so the current single expected-outcome map is not portable. The remaining 13 matrix cells behaved as pinned. Failing job: https://github.com/UseJunior/safe-docx/actions/runs/30283650631/job/90036435305

stevenobiajulu:
## Post-merge smoke: PASS

Validated from a clean detached checkout of current `main` at `dbdb6ef` (which includes #671 plus subsequent fixes):

- SHA-256-verified real corpus: 7 public NVCA DOCX files × `inplace`/`rebuild`; all 14 matrix cells matched their current characterization outcomes
- build and workspace lint: pass (9 pre-existing warnings, 0 errors)
- workspace tests: pass, including docx-compare 705 passed / 73 skipped, docx-core 1,256 passed / 2 expected failures / 14 skipped, docx-mcp 918 passed, and all remaining workspaces green
- strict spec coverage: pass
- conformance citations and generated conformance document: pass
- advisory LLM gate audited: PASS (6 pass, 0 warn, 0 error)

The temporary runtime-profile follow-up #690 was closed as superseded: #674/#681 fixed the three cross-runtime divergences before the follow-up could merge, and the current corpus matrix now passes on Darwin/Node 26 without runtime-specific pins.

PR #673, opened 2026-07-27
feat(docx-mcp): add paragraph numbering formatting

### LINKED ISSUE #653: No editing surface for paragraph numbering (w:numPr): cannot remove a stray auto-number or re-point a paragraph to another list
## Summary

There is no tool for a paragraph's numbering properties. The editing surface is
`replace_text`, `insert_paragraph`, `format_layout`, `clear_formatting`, comments
and footnotes — all of which reach run text, paragraph spacing and run formatting,
but none of which touches `w:numPr`.

Two operations that come up constantly in list-heavy documents are therefore
impossible without hand-editing `document.xml`:

**1. Remove a paragraph's auto-number.** A paragraph inherits a list definition it
should not be in and prints a stray label — e.g. a lead-in question that renders
as "(2) Does the applicant …" because a `w:numPr` was left on it. The fix is to
drop `<w:numPr>` from its `pPr`. There is no way to ask for that.

**2. Re-point a paragraph to a different list.** Two consecutive items each print
"(a)" because they carry different `w:numId` values, each starting its own
sequence. The fix is to set the second paragraph's `w:numId` to the first's. There
is no way to ask for that either.

`read_file` already surfaces exactly the state a caller needs to plan the edit:

```json
{"id": "_bk_…", "list_label": "(a)", "style": "top_level",
 "numbering": {"num_id": "33", "ilvl": 0, "is_auto_numbered": true},
 "paragraph_style_id": "ListParagraph"}
```

So the model is readable but not writable.

## Proposal

A `set_numbering` (or `format_numbering`, next to `format_layout`) tool taking a
paragraph anchor and one of:

- `{"remove": true}` — drop `w:numPr`, leaving the paragraph unnumbered
- `{"num_id": "<id>", "ilvl": <n>}` — re-point to an existing list definition
- optionally `{"match_paragraph_id": "_bk_…"}` — adopt another paragraph's
  numbering, which is what a caller usually means and avoids exposing raw ids

Revisionable, like the other editing tools: Word records this as a
`w:pPrChange`, so it can be a real tracked change rather than a silent rewrite.

## Why it matters

Numbering defects — a duplicated label, an orphan label, a sequence that restarts —
are among the most common things flagged on a long form document, and they are
visible to every reader. Hitting them today means dropping out of safe-docx
entirely, which forfeits tracked changes and the safety guarantees that are the
reason to use it.

## Related

- No editing surface for section properties (companion issue)
- No editing surface for headers and footers (companion issue)
- #562 (ECMA-376 §17.9 numbering coverage)

stevenobiajulu:
Closes #653

## Summary

- add a typed `docx-core` primitive for reading and mutating a paragraph's direct `w:numPr`
- add the revisionable, DOCX-only `format_numbering` MCP tool with remove, match-paragraph, and existing `num_id`/`ilvl` modes
- validate numbering instances, abstract definitions, and levels transactionally before mutation
- emit native `w:pPrChange` records for effective edits while keeping identical requests deterministic no-ops
- register the tool in the catalog, MCP bundle manifest, generated reference, and tutorial
- add the approved OpenSpec change and ECMA-376 §17.3.1.19, §17.9.18, and §17.9.3 evidence

## Deliberate v1 boundaries

- uses existing numbering definitions only; it does not create `word/numbering.xml` definitions
- changes direct paragraph numbering only; it does not rewrite style-inherited numbering
- guarantees structural/reference correctness, not a visible label in isolation from surrounding list context

## Validation

- `npm run build && npm run lint:workspaces && npm run test:run && npm run check:spec-coverage && npm run check:conformance-citations && npm run check:conformance-doc`
- `npm run check:tool-docs`
- `npm run check:site`
- `npx openspec validate add-paragraph-numbering-formatting --strict`
- focused regression set: 75 tests passed

## Real-DOCX smoke

Exercised `tests/test_documents/nvca-coi-regression/source.docx` end to end:

- removed direct numbering from a real numbered paragraph
- preserved paragraph count (234 → 234), paragraph text, and `word/numbering.xml` byte-for-byte
- clean save reread with no direct numbering/list label
- tracked save contained exactly one modification and one `w:pPrChange`
- clean and tracked DOCX archives passed `unzip -t`; `word/document.xml` passed `xmllint`
- source, clean, and tracked outputs rendered successfully in headless LibreOffice; inspected page 2 and confirmed the target label removal with surrounding text/layout intact

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -114,6 +114,36 @@ matches the complete paragraph text. Inspect the document structure before using
 it on a paragraph that carries section properties, is structurally required as
 the last paragraph in a table cell, or owns bookmark or comment anchors.
 
+### Repair DOCX Paragraph Numbering
+
+Use `format_numbering` when a paragraph has a stray direct list label:
+
+```text
+format_numbering(
+  file_path="~/docs/NDA.docx",
+  target_paragraph_id="_bk_4d71f33e924a",
+  remove=true
+)
+```
+
+To join a paragraph to an existing list sequence, copy another paragraph's
+explicit numbering:
+
+```text
+format_numbering(
+  file_path="~/docs/NDA.docx",
+  target_paragraph_id="_bk_9fc26ad74408",
+  match_paragraph_id="_bk_d39ac1a3b3f0"
+)
+```
+
+The tool also accepts an existing `num_id` with its `ilvl`, but matching a
+paragraph is usually safer because numbering IDs are local to each DOCX. The
+tool changes only direct `w:numPr`: it does not create numbering definitions,
+change numbering inherited only through a paragraph style, restart lists, or
+guarantee a rendered label without the surrounding list context. Effective
+changes are recorded as paragraph-property tracked changes.
+
 ## Step 6: Save Reviewable Outputs
 
 ```text
@@ -157,6 +187,7 @@ Visual review remains appropriate for material documents because Safe Docx is no
 | Inspect or accept revisions | `has_tracked_changes`, `extract_revisions`, `accept_changes` |
 | Insert a paragraph | `insert_paragraph` |
 | Delete an ordinary DOCX body paragraph | `replace_text` with the complete text and an empty `new_string` |
+| Remove or repair direct DOCX paragraph numbering | `format_numbering` |
 | Apply several edits together | `batch_edit` |
 | Convert DOCX to ODT | `convert_to_odt` |
 | Convert DOCX to Markdown | `export` with `format="markdown"` |
--- a/openspec/changes/add-paragraph-numbering-formatting/design.md
+++ b/openspec/changes/add-paragraph-numbering-formatting/design.md
@@ -1,0 +1,117 @@
+## Context
+
+Paragraph numbering in WordprocessingML is a paragraph property: direct list
+membership is represented by `w:pPr/w:numPr`, containing `w:ilvl` followed by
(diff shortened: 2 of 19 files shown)


stevenobiajulu:
Post-merge verification passed for merge commit `a72bb82f8748220b55d5c69774cfd68e09c51a26`.

- Clean checkout: `npm ci`, `npm run build`, and full `npm run test:run` passed.
- Real-document E2E: opened `tests/test_documents/nvca-coi-regression/source.docx`, removed direct `numId=2` / `ilvl=0` from paragraph `_bk_dc966e576f73`, and saved clean + tracked variants.
- Structural invariants: 234 paragraphs before/after; target rereads with `numbering: { num_id: null, ilvl: null, is_auto_numbered: false }`; `word/numbering.xml` is byte-identical across source/clean/tracked.
- Tracked evidence: one modification and one `w:pPrChange`, authored by `SafeDocX AI`.
- Package/XML checks: both DOCX archives pass `unzip -t`; both `word/document.xml` parts pass `xmllint`.
- Consumer smoke: source, clean, and tracked documents all rendered successfully through LibreOffice at 48 pages. Visual inspection confirms the selected paragraph loses its list label in clean output, later labels renumber, and tracked output visibly marks the property change without layout corruption.

Local evidence retained at:
- log: `/private/tmp/automerge-smoke-673-20260727T170005Z.log`
- artifacts: `/private/tmp/automerge-smoke-673-20260727T170005Z-artifacts/`

Production deployment: not applicable for this library/MCP change.

PR #677, opened 2026-07-27
fix(docx-compare): exclude rsid attributes from run-merge identity

### LINKED ISSUE #675: atomizer: rsid attributes participate in run-merge identity, producing phantom delete+insert on unchanged text
## Summary

The atomizer treats **rsid attributes as part of run-merge identity**. When a source document splits a token across runs whose rsids differ, those fragments never merge into one atom, so they cannot match the corresponding single atom in the revised document. The comparison then emits a **delete + insert over text that is byte-identical in both documents**.

rsids (`w:rsidR`, `w:rsidRPr`, …) are revision-save identifiers. They are editing-session bookkeeping with no rendering or semantic effect, so they must not influence whether two adjacent runs may be merged for atomization.

The user-visible result is a redline that marks an unchanged figure as changed. On a real document this showed up on currency amounts — a reviewer sees `$204,000.00` struck through and reinserted as `$204,000.00`, and has to stop and work out whether the number moved. That is exactly the class of no-op noise the atomizer path is supposed to avoid.

## Reproduction

Self-contained, no external fixture. Writes four ~1 KB documents: a repro pair and a control pair that differ **only** in whether the split runs carry matching rsids.

```python
import zipfile

W = 'xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main"'
CT = '''<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<Types xmlns="http://schemas.openxmlformats.org/package/2006/content-types">
<Default Extension="rels" ContentType="application/vnd.openxmlformats-package.relationships+xml"/>
<Default Extension="xml" ContentType="application/xml"/>
<Override PartName="/word/document.xml" ContentType="application/vnd.openxmlformats-officedocument.wordprocessingml.document.main+xml"/>
</Types>'''
RELS = '''<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<Relationships xmlns="http://schemas.openxmlformats.org/package/2006/relationships">
<Relationship Id="rId1" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/officeDocument" Target="word/document.xml"/>
</Relationships>'''

def wrap(body):
    return (f'<?xml version="1.0" encoding="UTF-8" standalone="yes"?>\n'
            f'<w:document {W}><w:body>{body}'
            f'<w:sectPr><w:pgSz w:w="12240" w:h="15840"/></w:sectPr></w:body></w:document>')

# Source: amount split across three runs. The middle run carries no w:rsidRPr;
# its neighbours do. Word produces this shape routinely when a figure is
# retyped in a later editing session.
SPLIT_ALTERNATING_RSID = '''<w:p>
<w:r w:rsidRPr="00F9719D" w:rsidR="00932CCC"><w:t xml:space="preserve">Alpha clause; beta clause; the amount shall not exceed $</w:t></w:r>
<w:r w:rsidR="00932CCC"><w:t>204</w:t></w:r>
<w:r w:rsidRPr="00F9719D" w:rsidR="00932CCC"><w:t xml:space="preserve">,000.00 or 0.50%, whichever is greater. Legacy Costs settle monthly.</w:t></w:r>
</w:p>'''

# Control: identical split, all three runs carry the same rsids.
SPLIT_UNIFORM_RSID = SPLIT_ALTERNATING_RSID.replace(
    '<w:r w:rsidR="00932CCC"><w:t>204</w:t></w:r>',
    '<w:r w:rsidRPr="00F9719D" w:rsidR="00932CCC"><w:t>204</w:t></w:r>')

# Revised: one run. "$204,000.00" is unchanged; the only real edits are the
# leading clause and the trailing term.
SINGLE_RUN = '''<w:p>
<w:r><w:t xml:space="preserve">Alpha clause and any further clause reasonably incurred; the amount shall not exceed $204,000.00 or 0.50%, whichever is greater. Updated Costs settle monthly.</w:t></w:r>
</w:p>'''

def write(path, body):
    with zipfile.ZipFile(path, 'w', zipfile.ZIP_DEFLATED) as z:
        z.writestr('[Content_Types].xml', CT)
        z.writestr('_rels/.rels', RELS)
        z.writestr('word/document.xml', wrap(body))

write('bug-a.docx', SPLIT_ALTERNATING_RSID)
write('bug-b.docx', SINGLE_RUN)
write('ctl-a.docx', SPLIT_UNIFORM_RSID)
write('ctl-b.docx', SINGLE_RUN)
```

```bash
node packages/docx-compare/dist/cli/index.js bug-a.docx bug-b.docx bug-out.docx --mode inplace
node packages/docx-compare/dist/cli/index.js ctl-a.docx ctl-b.docx ctl-out.docx --mode inplace
```

## Actual

`bug-out.docx` — the unchanged amount is struck and reinserted:

```xml
<w:delText xml:space="preserve">$</w:delText>
<w:delText>204</w:delText>
<w:delText xml:space="preserve">,000.00</w:delText>
<w:t xml:space="preserve">$204,000.00</w:t>
```

`ctl-out.docx` — same run split, matching rsids, amount passes through untouched:

```xml
<w:t xml:space="preserve"> the amount shall not exceed $204,000.00 or 0.50%, whichever is greater. </w:t>
```

Stats confirm the delta is confined to the phantom: repro `insertions 5 / deletions 5 / deletedAtoms 7`, control `insertions 4 / deletions 4 / deletedAtoms 4`.

## Expected

Both pairs produce the same output. rsid attributes are excluded from run-merge identity, the three fragments merge to one atom, and the amount is not marked.

## Notes

- Reproduced in `inplace` mode on a build of current `main`, and independently on a real-world document pair where the same three-run/alternating-rsid shape occurs on two currency amounts.
- The paragraph must contain at least one genuine edit for this to surface — with no other change the paragraph is skipped and the segmentation difference is correctly ignored. That is why it presents as "one figure in an otherwise legitimately-edited sentence is wrongly marked."
- Ruled out during isolation, in case it saves someone the same passes: differing `w:rPr`, position inside `w:tc`, and the `$` being glued to the end of the preceding run. None of those alone reproduces; only the rsid mismatch does.
- Related but distinct: #286 was `replace_text` *stripping* rsids in docx-core. This is the docx-compare atomizer *honouring* them where it should not. #409 and #42 cover no-ops from empty paragraphs and field boundaries respectively, not run-level rsid identity.

stevenobiajulu:
## Summary

The docx-compare atomizer emitted phantom delete+insert markup over byte-identical text when a source document split a token across runs whose rsid attributes differed (e.g. `$204,000.00` struck through and reinserted unchanged). Root cause: the run pre-merge step treated rsid attributes (`w:rsidR`, `w:rsidRPr`, `w:rsidDel`, ...) — revision-save session bookkeeping with no rendering or semantic effect — as part of run-merge identity, so rsid-fragmented runs never premerged and could not match the revised document's single atom.

Fixes: #675

## Changes

Three coupled changes, each empirically load-bearing (details in the commit bodies):

1. **rsid attributes excluded from run-merge identity** (`premergeRuns.ts` `attrsEqual`): compares only non-rsid attributes, symmetrically (equal filtered counts + per-attribute value check). rsids are recognised by w-namespace + `localName.startsWith('rsid')`, with a qualified-name (`w:rsid*`) fallback for attributes created without namespace metadata. Merged runs keep the target (first) run's rsid attributes — the conservative choice versus stripping or synthesising one. Non-rsid attribute differences still block merging.
2. **Whitespace-only direct text no longer marks a run unsafe to merge** (`runIsSafeToMerge`): pretty-printed documents put indentation text nodes inside every run; treating those as content made premerge a no-op on one side of a comparison and desynchronised run boundaries (NVCA COI `deletions` 385 → 540 with fix 1 alone). Every direct text child is scanned — not just the first — so stray non-whitespace text still blocks merging and can never be silently dropped (peer-review Medium finding).
3. **Edge punctuation split into its own atoms during word tokenization** (`atomizer.ts` `splitAtomIntoWords`): with premerge healing run fragmentation, whitespace-only word-splitting produced tokens like `[or` / `officer]`, so a punctuation-only edit (template bracket removal) struck and reinserted the unchanged neighbouring words (peer-review High finding). No premerge-local gate can separate the harmful glue (`[`+`or`) from the essential one (`$`+`204` — the issue-675 case): same character classes. Tokenization is now a pure function of paragraph text: leading `'"([{<` and trailing `,.:;!?'")]}>` punctuation become separate atoms on both sides; currency and interior punctuation stay glued (`$204,000.00` remains one atom). The punctuation-merge pass is skipped when word-splitting ran (edge splitting normalizes the same boundary differences symmetrically); run-level passes keep it unchanged.

## Verification

Issue repro (clean `main` build first, then fixed build):

| pair | before | after |
|---|---|---|
| alternating-rsid (bug) | insertions 5 / deletions 5 / deletedAtoms 7 | insertions 3 / deletions 3 / deletedAtoms 3 |
| uniform-rsid (control) | insertions 4 / deletions 4 / deletedAtoms 4 | insertions 3 / deletions 3 / deletedAtoms 3 |

Both pairs now produce byte-identical `document.xml`; `$204,000.00` never appears in a `w:delText`. (The control also improved: edge-punct tokenization keeps `clause` matched where `clause;` was previously deleted whole.)

Real-document check (NVCA COI regression pair, inplace) — final state is strictly better than the pre-PR baseline on every stat: insertions **205** / deletions **375** / modifiedParagraphs **26** vs baseline 224 / 385 / 27. The `[or officer]` "personally liable" paragraph emits exactly the baseline markup (brackets deleted, unchanged words untouched). Intermediate states that regressed this document (fix 1 alone: 401/540/79; fixes 1+2: 235/380/33 with word-consuming bracket edits) are why all three changes ship together.

Full pre-submit chain (`build`, `lint:workspaces`, `test:run` — 50/50 docx-compare files, 121/121 docx-core files — `check:spec-coverage`, `check:conformance-citations`, `check:conformance-doc`) passes locally.

## Peer review

Codex dynamic review: REQUEST-CHANGES with one High (bracket-adjacent unchanged words struck on NVCA — fixed by change 3), one Medium (first-text-child-only guard could silently drop stray run text — fixed by change 2's full scan). Both findings verified, accepted, and covered by new regression tests. Adjudication comment on this PR has the details.

## Follow-up candidates (not in this PR)

- `atomizer.ts` `elementIdentityString` hashes all attributes, including rsids, into empty-paragraph `pPr` identity (`atomizer.test.ts` asserts `w:rsidR` on `w:pPr` changes the hash). A pPr rsid difference on an otherwise-identical empty paragraph could produce the same phantom-diff class at paragraph level. `w:rPr` carries no rsid attributes in the vendored strict schema (confirmed by review), so `runPropertiesEqual` needs no change.

### DIFF
--- a/packages/docx-compare/src/atomizer.test.ts
+++ b/packages/docx-compare/src/atomizer.test.ts
@@ -10,6 +10,7 @@ import {
   createComparisonUnitAtom,
   atomizeTree,
   getAncestors,
+  splitAtomsIntoWords,
   EMPTY_PARAGRAPH_TAG,
 } from './atomizer.js';
 import { CorrelationStatus, OpcPart } from '@usejunior/docx-core';
@@ -1347,6 +1348,80 @@ describe('getAncestors', () => {
       expect(ancestors[1]).toBe(body);
       expect(ancestors[2]).toBe(para);
       expect(ancestors[3]).toBe(run);
+    });
+  });
+});
+
+describe('splitAtomsIntoWords edge punctuation', () => {
+  const PART: OpcPart = { uri: 'word/document.xml', contentType: 'text/xml' };
+
+  function wordTexts(text: string): string[] {
+    const textEl = el('w:t', {}, undefined, text);
+    const run = el('w:r', {}, [textEl]);
+    const paragraph = el('w:p', {}, [run]);
+    const atom = createComparisonUnitAtom({
+      contentElement: textEl,
+      ancestors: [paragraph, run],
+      part: PART,
+    });
+    return splitAtomsIntoWords([atom]).map((a) => a.contentElement.textContent ?? '');
+  }
+
+  test('splits bracket edges into their own atoms', async ({ given, when, then }: AllureBddContext) => {
+    let tokens: string[];
+
+    await given('text with template brackets "[or officer]"', () => {});
+
+    await when('the atom is word-split', () => {
+      tokens = wordTexts('[or officer]');
+    });
+
+    await then('brackets become separate atoms so the words can match a bracketless side', () => {
+      // Tokenization must be a pure function of paragraph text: "[or officer]"
+      // vs "or officer" must share the "or" and "officer" atoms (issue #675).
+      expect(tokens).toEqual(['[', 'or', ' ', 'officer', ']']);
+    });
+  });
+
+  test('keeps currency amounts as single atoms', async ({ given, when, then }: AllureBddContext) => {
+    let tokens: string[];
+
+    await given('text containing a currency amount', () => {});
+
+    await when('the atom is word-split', () => {
+      tokens = wordTexts('exceed $204,000.00 net');
+    });
+
+    await then('the amount stays one atom; interior and currency punctuation stay glued', () => {
+      expect(tokens).toEqual(['exceed', ' ', '$204,000.00', ' ', 'net']);
+    });
+  });
+
+  test('splits trailing sentence punctuation', async ({ given, when, then }: AllureBddContext) => {
+    let tokens: string[];
+
+    await given('a single-token atom with trailing punctuation', () => {});
+
+    await when('the atom is word-split', () => {
+      tokens = wordTexts('Conduct,');
+    });
+
+    await then('the comma becomes its own atom, matching a "Conduct" + "," run split', () => {
+      expect(tokens).toEqual(['Conduct', ',']);
+    });
+  });
+
+  test('leaves pure punctuation tokens intact', async ({ given, when, then }: AllureBddContext) => {
+    let tokens: string[];
+
+    await given('a token made only of punctuation', () => {});
+
+    await when('the atom is word-split', () => {
+      tokens = wordTexts('a ). b');
+    });
+
+    await then('the punctuation token stays a single atom', () => {
+      expect(tokens).toEqual(['a', ' ', ').', ' ', 'b']);
     });
   });
 });
--- a/packages/docx-compare/src/atomizer.ts
+++ b/packages/docx-compare/src/atomizer.ts
@@ -863,9 +863,15 @@ export function atomizeTree(
     : mergedAtoms;
 
   // Step 4: Merge punctuation-only atoms with preceding text
-  // This handles "Conduct" + "," vs "Conduct," split differences
-  // Must run AFTER word split since that's when punctuation becomes separate atoms
-  const atoms = mergePunctuationAtoms(wordSplitAtoms, normalizedOptions);
+  // This handles "Conduct" + "," vs "Conduct," split differences for
+  // run-level atomization. When word-splitting ran, edge punctuation was
+  // already split into its own atoms on BOTH sides (splitTokenAtPunctuationEdges),
+  // which normalizes the same boundary differences symmetrically; re-gluing
+  // here would depend on run structure and re-introduce asymmetric tokens
+  // (e.g. "officer]" on one side vs "officer" + "]" on the other).
+  const atoms = normalizedOptions.splitTextIntoWords
+    ? wordSplitAtoms
+    : mergePunctuationAtoms(wordSplitAtoms, normalizedOptions);
 
   console.log(
     `[DEBUG] atomizeTree: created ${rawAtoms.length} atoms, field-collapsed to ${fieldCollapsedAtoms.length}, merged to ${mergedAtoms.length}, word-split to ${wordSplitAtoms.length}, punct-merged to ${atoms.length}, ${state.emptyParagraphCount} empty paragraphs`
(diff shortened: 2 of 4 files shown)


stevenobiajulu:
## Peer review verdict + adjudication

Codex CLI dynamic review (`--sandbox workspace-write`, full prompt + raw output preserved in the worktree's `.peer-review/`): **REQUEST-CHANGES**, two findings. Both were verified against the working tree and **accepted**; fixes are in b28caa6.

### High — accepted and fixed: bracket-adjacent unchanged words struck on the NVCA fixture

Codex reproduced with real output: after the rsid+whitespace premerge fixes, the NVCA COI pair converted the punctuation-only edit `[or officer]` → `or officer` into `delText="[or"` + `delText="officer]"` + reinsertion of `or` / `officer` — three times in the "personally liable" paragraph. Mechanism confirmed: whitespace-only word-splitting tokenized the now-premerged run as `[or` / `officer]`, while the old output had only avoided this because the source's run boundaries (`[` | `or officer` | `]`) were doing accidental tokenization work.

I verified no premerge-local gate can fix this: the harmful glue (`[`+`or`) and the essential one (`$`+`204`, the issue-675 case) are the same character classes, so any boundary-based restriction on premerge un-fixes the issue. The durable fix (commit b28caa6) makes tokenization a pure function of paragraph text: `splitAtomIntoWords` splits leading/trailing edge punctuation into separate atoms on both sides, and the run-structure-dependent punctuation-merge pass is skipped when word-splitting ran. Result, measured:

- "personally liable" paragraph now emits exactly the baseline markup: `delText` = `[`, `]`, `[` — no word reinsertions.
- NVCA stats strictly better than the pre-PR baseline: insertions 205 / deletions 375 / modifiedParagraphs 26 (baseline 224 / 385 / 27; the reviewed state was 235 / 380 / 33).
- Issue repro improved 4/4/4 → 3/3/3, both pairs still byte-identical, `$204,000.00` still a single matched atom.
- Reduced regression added as requested: `[or officer] → or officer` with rsid-fragmented runs, asserting the brackets are deleted and `or`/`officer` are neither deleted nor reinserted.

### Medium — accepted and fixed: later direct text nodes bypassed the whitespace guard

Confirmed: `getLeafText()` returns only the first direct text child, so `<w:r>ws<w:t>B</w:t>STRAY</w:r>` passed the guard and `mergeRunInto` (element children only) silently dropped `STRAY`. `runIsSafeToMerge` now scans every direct text child (`hasNonWhitespaceDirectText`); test added with whitespace before `<w:t>` and stray text after it, asserting no merge and the stray text preserved.

### Review assertions independently confirmed

- Vendored strict schema: `CT_R` defines exactly `rsidRPr`/`rsidDel`/`rsidR`; `CT_RPr` has no rsid attributes — so `runPropertiesEqual` needs no filtering.
- No comparison/reconstruction code consumes run rsids for correlation; the general `mergeRuns` normalization already strips rsids by default.
- Full pre-submit chain green after the fixes (all workspaces, 50/50 + 121/121 test files in the touched packages).

stevenobiajulu:
## Post-merge smoke passed

Merged: `1caa9b1f2fd4ca7b63a2d2c26dcd4ffd0221d631`
Built from: fresh `git clone` of `main` @ `1caa9b1` (clean `npm install` + `npm run build`, both exit 0)
Smoke: issue-675 repro + control via `packages/docx-compare/dist/cli/index.js`, then a real-document comparison over the NVCA COI pair

**Steps**
- build (fresh clone, exit 0)
- issue repro pair (`bug-a`/`bug-b`, alternating-rsid three-run split): stats `insertions 3 / deletions 3 / deletedAtoms 3`, exit 0
- issue control pair (`ctl-a`/`ctl-b`, uniform rsids): identical stats, and `word/document.xml` **byte-identical** to the repro pair's output
- `$204,000.00` present in a `w:t`, never in a `w:delText`

**Real-world fixture**
- `tests/test_documents/nvca-coi-regression/` (real NVCA Certificate of Incorporation source + filled pair, inplace mode)
  - Stats: `insertions 205 / deletions 375 / modifiedParagraphs 26` — strictly better than the pre-PR baseline (224 / 385 / 27), matching the numbers verified during review
  - "personally liable" paragraph (the peer-review High finding site): delText = `[`, `]`, `[` only; zero insertions — unchanged words untouched
  - No bracket-glued phantom tokens (`[or`, `officer]`, `[that`) anywhere in the output
  - Output validates against the Transitional WML schema (`scripts/check_emitted_document_schema.mjs`: 1 of 1 pass)
  - LibreOffice headless renders the redline to a 50-page PDF; pages inspected visually — tracked-change strikethrough + change bars render correctly

Visual evidence: `$SCRATCH/lo-render/smoke-nvca-redline.pdf` (local; scratchpad path in session log)
Word check: **blocked by local environment, not the document** — Microsoft Word on this machine currently opens no documents at all (including the pre-existing checked-in `source.docx` fixture; a wedged zero-document instance was relaunched with no change, no repair dialogs raised). Flagging so the next `word-fidelity-check` run isn't misread as a document failure.

**Cleanup**
- deleted remote branch `675-atomizer-rsid-run-merge-20260727`

Issue #583, opened 2026-07-17
perf(docx-compare): compareDocuments spends ~42% in XML DOM, ~9% in GC, only ~9% in LCS — four optimization tracks, peer-reviewed for exactness


stevenobiajulu:
## Context

CPU-profiling `compareDocuments` on a real NVCA stock purchase agreement (375 paragraphs / ~16.5k words → ~39.1k vs ~34.1k word-level atoms, via the openagreements.org redline path, cf. UseJunior/openagreements-org-deploy#14) shows the DP LCS is **not** the bottleneck:

| Cost | Share | What |
|---|---|---|
| ~42% | XML DOM | `walkDOM`, live NodeList maintenance, SAX parse (`@xmldom/xmldom`) |
| ~9% | GC | atom objects, DP arrays, DOM churn |
| ~9% | LCS | `atomsEqual` + `computeAtomLcs` + `hierarchicalCompare` |
| rest | | per-atom crypto SHA-1, zip inflate/deflate |

DP-cell accounting on that fixture: Pass 1 group DP ≈ 152k cells; within-group atom DP ≈ 10.5M cells; Pass 2 similarity DP ≈ **492 cells** (~0.005%).

Four optimizations were proposed and then adversarially peer-reviewed for exactness by two independent tool-using reviewers (Codex + agy) executing probes against `92928bb` (differential alignment probes over all 3,969 binary sequence pairs ≤5, determinism runs, mutation traces, forced-hash-collision tests, full 525-test suite). **Two of the four survived as stated; two were refuted as stated and replaced with narrower exact variants.** Verdicts below are the reviewed consensus.

### A1. Suffix-only common-trim before within-group atom DP
Trimming the maximal common **suffix** per matched paragraph pair is alignment-preserving for the current backtracker (it starts at the lower-right corner and always consumes an equal trailing pair diagonally). Probe: suffix-trim reproduced the entire current `LcsResult` on all 3,969 sequence pairs (0 mismatches).
⚠️ **Prefix trimming is NOT exact** (see Track B): with duplicate identities (`[A(bold), A(italic)]` vs `[A(bold)]`) it flips which atom is deleted, and downstream `formatChanges` changes 1→0.

### A2. Full effective-identity interning to replace per-atom crypto SHA-1
Replace SHA-1 with a shared interner mapping the **complete effective atom identity** → integer; LCS equality becomes an int compare.
[...]

PR #680, opened 2026-07-27
test(docx-compare): pin LCS alignment so tie-breaker changes cannot ship silently

### LINKED ISSUE #584: test(docx-compare): LCS alignment is unpinned — suite passes with an inverted DP tie-breaker even though alignment changes formatChanges output
## Context

While peer-reviewing the four `docx-compare` optimization claims that became #583 (two executing reviewers: Codex CLI + agy, both run against a detached worktree at `92928bb`), the review surfaced three verification gaps in the package itself. They are filed here separately because they are test/robustness gaps in the current code, independent of whether any #583 optimization ever lands — but gap 1 is also the exact gate #583's Track B (Myers diff, general prefix trimming) is blocked on.

## Gap 1 (primary): LCS *alignment* is completely unpinned by the test suite

LCS length is unique; the alignment (which pairs get matched on ties) is not. The current backtracker in `packages/docx-compare/src/baselines/atomizer/atomLcs.ts` resolves ties with `dp[i-1][j] > dp[i][j-1]` (favoring deletions). That chosen alignment is **user-visible**: it feeds `comparisonUnitAtomBefore`, which drives format-change detection, move detection, and merged-output ordering.

**Executed evidence that no test pins it:**

- agy inverted the DP tie-breaker (`>` → `>=`) in `atomLcs.ts` — a change that provably alters which optimal alignment is chosen — and ran the full suite: **all 523 tests passed.**
- Codex searched for any test referencing the alignment: `rg -n "computeAtomLcs\(" packages/docx-compare/src -g '*.test.ts'` → exit 1 (no matches). The only `atomLcs.test.ts` asserts length/diff-minimality, not the alignment path.

**Executed evidence that alignment choice changes output** (so this is a real hole, not a don't-care):

- Codex's duplicate-identity probe: base contains two same-text/different-format atoms `[A(bold), A(italic)]`, modified contains `[A(bold)]`. Depending on which `A` the backtracker matches, `formatChanges` is **1 vs 0** and merged statuses flip between `[Deleted, FormatChanged]` and `[Equal, Deleted]`.
- Codex's Myers-vs-current-DP differential on the real NVCA fixture: **2,724 / 3,969** paragraph-pair alignments differ while every LCS *length* is identical.

**Consequence:** any refactor of the LCS internals (Myers, prefix trimming, or even an innocuous-looking loop rewrite) can silently change format/move stats and the visual redline, and CI will stay green.

**Proposed fix:**
1. Add duplicate-identity fixtures (same text, different `rPr`) with tests asserting the *chosen alignment* and the resulting `formatChanges`/merged statuses — the tie-breaker inversion above must make them fail.
2. Add a differential alignment harness: run `compareDocuments` on the real-template corpus and snapshot per-pair alignments, so an alignment-affecting change shows up as a reviewable diff instead of silence. This is the acceptance gate #583 Track B needs before Myers/prefix-trim can be evaluated.

## Gap 2: `atomsEqual` never re-checks attributes — the SHA-1 hash is silently load-bearing

`atomsEqual` (same file) re-verifies `textContent` and `tagName` after a hash match, but **not attributes**. Attribute identity is enforced *only* by `hashElement` having mixed sorted attributes into the SHA-1. Both reviewers independently flagged this; agy's forced-collision probe (equal hash, equal text, different attributes) produced a false "equal" verdict.

Not currently exploitable (SHA-1 collisions don't occur in practice), but it makes the hash function's *composition* a hidden correctness contract: anyone who "optimizes" identity to text+tag (as claim 4 of the review originally proposed) silently merges `<w:t>` runs with different formatting attributes. Either add the attribute re-check to `atomsEqual`, or document the contract at `hashElement` + `atomsEqual` and add a regression test with same-text/different-attribute atoms.

## Gap 3: output bytes are non-deterministic unless `date` is pinned

`compareDocuments` defaults `date` to `new Date()` (pipeline.ts, ~line 581), which lands in every `w:ins`/`w:del` `w:date` attribute — so two runs on identical inputs produce different bytes across a second boundary (agy's probe confirmed byte-equality only when runs land in the same second or `date` is passed). Stats are deterministic. Worth documenting (it blocks naive byte-level caching/snapshotting, relevant to #583 Track A), and a byte-determinism test with a pinned `date` would pin the invariant.

## References

- Review target: commit `92928bb`
- Companion perf issue: #583 (Track B is gated on Gap 1's harness)

stevenobiajulu:
## Summary

Test-only PR closing **Gap 1 of #584**: the LCS *alignment* chosen by `computeAtomLcs` was completely unpinned — inverting the DP backtracker tie-breaker in `packages/docx-compare/src/baselines/atomizer/atomLcs.ts` (`dp[i-1][j] > dp[i][j-1]` → `>=`) provably changes which runs receive `w:rPrChange` revisions and which text is deleted/re-inserted, yet the entire suite stayed green. Since the alignment feeds `comparisonUnitAtomBefore` (format-change detection, move detection, merged-output ordering), any LCS-internals refactor — including the Myers / prefix-trim work gated behind #583 Track B — could silently move real redline output.

**Scope: Gap 1 only.** Gaps 2 (`atomsEqual` attribute re-check) and 3 (date-pinned byte determinism) from #584 are NOT touched here. The issue's proposed fix 2 (corpus-wide differential alignment snapshot harness) is also left as follow-up; this PR is the characterization-test half (fix 1).

## What is pinned

One new file, `packages/docx-compare/src/baselines/atomizer/atomLcs-alignment-pinning.test.ts`, four tests pinning **current observable behavior** (documented in the file header as a characterization, not a claim of unique correctness):

1. **Unit, legacy-equality path** — swapped adjacent atoms `[Alpha, Beta]` vs `[Beta, Alpha]`: the Beta pair survives (`matches === [{originalIndex: 1, revisedIndex: 0}]`); Alpha is deleted + re-inserted.
2. **Unit, interned-identity path** — reversed sequence `[A,B,C,D]` vs `[D,C,B,A]`: the backtrack crosses three consecutive DP ties and D survives (`[{originalIndex: 3, revisedIndex: 0}]`).
3. **Pipeline (`compareDocuments`)** — swapped formatted runs inside one paragraph: the alignment decides that **Beta** (bold→plain) anchors the match, so the only `w:rPrChange` sits on the Beta run recording old `<w:b/>`, `stats.formatChanges === 1`, and Alpha — not Beta — is the deleted/re-inserted text.
4. **Pipeline, duplicate identity** (the issue's proposed fixture) — `[Same:bold][Same:italic]` collapsing to `[Same:bold]`: the surviving run pairs with the *italic* original (rPrChange old rPr `<w:i/>`) and the *bold* duplicate is deleted. This test is inversion-neutral by construction (the backtracker matches equal heads before consulting the tie) and is documented as pinning the eager last-occurrence choice that a first-occurrence-preferring rewrite (e.g. Myers) would flip.

## Executed evidence (both directions)

Gap reproduction — tie-breaker inverted, pre-existing suite only:

```
 Test Files  50 passed (50)
      Tests  699 passed | 73 skipped (772)   # exit 0 — the gap is real
```

New tests with the tie-breaker inverted (`>` → `>=`):

```
 ❯ atomLcs-alignment-pinning.test.ts (4 tests | 3 failed)
   × swapped adjacent atoms: tie resolves toward the later original / earlier revised match
   × reversed sequence: the single surviving atom is the LAST original element
   × swapped formatted runs: alignment decides which run gets the rPrChange revision
   ✓ duplicate identity: the LAST same-text original run anchors the match, not the first
 Tests  3 failed | 1 passed (4)   # exit 1
```

New tests with the tie-breaker as-is:

```
 ✓ atomLcs-alignment-pinning.test.ts (4 tests)
 Tests  4 passed (4)   # exit 0
```

The inversion was applied only transiently for verification; `git diff origin/main -- packages/docx-compare/src/baselines/atomizer/atomLcs.ts` is empty — **no production code is modified**.

## Notes for reviewers

- The pipeline fixtures carry shared context words (`Clause text before … after end`) because single-token paragraphs ("AlphaBeta" vs "BetaAlpha") fail the hierarchical paragraph-similarity gate (Jaccard 0) and degrade to whole-paragraph delete+insert before `computeAtomLcs` ever runs — discovered by probing, documented inline.
- No position is taken on whether the current tie-breaker is *right*; changing it would move real redline output and is a human decision. This PR only makes that change impossible to ship silently.

Full local gate chain passed: build, lint:workspaces, test:run, check:spec-coverage, check:conformance-citations, check:conformance-doc (exit 0).

Fixes: #584


stevenobiajulu:
## Codex peer review: APPROVE — adjudication

An independent Codex CLI review (workspace-write sandbox, dynamic execution) was run against this branch with instructions to re-verify the PR's central claim itself rather than trust the description.

**Verdict: APPROVE. Blocking: none. Should-fix: none. Nit: 1 (accepted, fixed in b70f07a).**

### What the reviewer independently executed (from its verification trace)

- Ran the new test file as-is: `4 passed`, exit 0.
- **Independently inverted the tie-breaker** (`atomLcs.ts` line 94, `>` → `>=`) and re-ran the file: `3 failed | 1 passed`, exit 1, with the exact pinned values flipped (`{originalIndex: 1, revisedIndex: 0}` → `{0, 1}`; `{3, 0}` → `{0, 3}`; `stats.formatChanges` 1 → 0) — confirming the characterization actually discriminates.
- Ran the pre-existing suite (excluding the new file) under the inversion: `699 passed | 73 skipped`, exit 0 — independently reproducing the #584 gap this PR closes.
- Restored production code and confirmed the full package baseline: `51 files / 703 passed`, exit 0; `tsc --noEmit` and eslint on the new file both exit 0.
- Verified production-file integrity by SHA-256: working-tree `atomLcs.ts` identical to `HEAD:atomLcs.ts` (`b3b3c830…`).

### Adjudicated findings

1. **Nit (accepted → fixed in b70f07a):** the file-header JSDoc claimed *any* alignment-changing refactor must fail here, which overstates what four fixtures guarantee — a refactor could preserve these cases while changing other ambiguous alignments. Wording now scopes the claim to the pinned choices and names the corpus-wide differential harness (issue #584's proposed fix 2) as follow-up.
2. **Exact `matches` assertions — reviewer sided with keeping them:** `computeAtomLcs` is publicly re-exported, and match identity/order feed downstream behavior; loosening to a pair-set comparison was considered and rejected.
3. **Grouping dependency of the pipeline fixtures — accepted as-is:** the fixtures genuinely traverse hierarchical grouping (TF-IDF degenerate → Jaccard fallback), which is acceptable for integration-level pins; the two unit tests isolate `computeAtomLcs` directly.
4. **Tie-breaker correctness:** reviewer found no evidence the current tie-breaker is wrong; not changing production behavior confirmed as the right call.

### Post-review integrity check (reviewer ran with write access)

Because the reviewer inverts the tie-breaker as part of verification, the production file was re-checked after the review: `gh pr diff 680 --name-only` returns exactly one path (the new test file), and `git diff origin/main -- packages/docx-compare/src/baselines/atomizer/atomLcs.ts` is empty — **`atomLcs.ts` is byte-identical to origin/main**. No reviewer-left inversion shipped.

Raw prompt and reviewer output retained locally under the worktree's `.peer-review/` (not committed).

PR #681, opened 2026-07-27
fix(docx-compare): split revision wrappers at partial bookmark boundaries

### LINKED ISSUE #643: docx-compare: a bookmark boundary partway inside a revision wrapper cannot be repositioned faithfully without splitting the wrapper

stevenobiajulu:
## Summary

A bookmark boundary sitting partway inside an inline revision wrapper (`w:ins` / `w:del` / `w:moveFrom` / `w:moveTo`) is now repositioned by **splitting the wrapper at the boundary**, copying its attributes onto the second half under a fresh `w:id`, so the range keeps its exact original span in both the Accept All and Reject All projections.

Before (pinned lossy behavior — the end was dropped in front of the wrapper, shrinking the range to `kept`):

```xml
<w:bookmarkStart w:id="50" w:name="Partial"/>
<w:r><w:t>kept</w:t></w:r>
<w:bookmarkEnd w:id="50"/>
<w:del w:id="1">
  <w:r><w:delText>inside-range</w:delText></w:r>
  <w:r><w:delText>outside-range</w:delText></w:r>
</w:del>
```

After (the range still covers `kept` + `inside-range`):

```xml
<w:bookmarkStart w:id="50" w:name="Partial"/>
<w:r><w:t>kept</w:t></w:r>
<w:del w:id="1"><w:r><w:delText>inside-range</w:delText></w:r></w:del>
<w:bookmarkEnd w:id="50"/>
<w:del w:id="900"><w:r><w:delText>outside-range</w:delText></w:r></w:del>
```

## Why two sites

Empirically (repro first, on a clean build), each reconstruction mode reaches the collapse through a different placement:

1. **`enforceConsumerCompatibility` / `liftMarkersAroundWrapper`** (shared by both modes; how the rebuild path and pre-existing tracked markup collapse). The pass now walks every marker out of the wrapper level by level. A marker partway through a container splits it — trailing content moves into a shallow clone (attributes copied, fresh `w:id` from the caller's existing collision-aware allocator) placed after the marker. A marker at a container edge steps directly outside that edge, which reproduces the exact pinned #641 shapes for wholly-enclosed ranges with no special case, and handles nested wrappers (`w:ins` around `w:del`) by construction. Tail clones emptied by adjacent boundaries are dropped.

2. **The inplace deletion/moveFrom emission** (how the NVCA corpus collapse actually happens: the markers never sit inside a wrapper by the time the pass runs). `cloneUnemittedSourceBookmarkMarkers` now reports which side of the source run each marker sat on; markers that FOLLOWED the run re-emit after the fragment's wrapper, and because word-level atomization fragments one source run into many wrappers, each later fragment re-anchors its run's trailing clones after itself so they come to rest after the run's final fragment. The next-atom insertion anchor advances past those clones so later runs cannot push them rightward.

## Postprocess passes cannot undo the split

`mergeAdjacentTrackChangeSiblings`, `coalesceDelInsPairChains`, and `mergeWhitespaceBridgedTrackChanges` all run BEFORE `enforceConsumerCompatibility` in the inplace pipeline (rebuild runs none of them), and their predicates require direct element adjacency or a whitespace-only `w:r` bridge — an interposed `w:bookmarkEnd` blocks all three. This is verified empirically, not by inspection: the new e2e tests assert the `w:del, w:bookmarkEnd, w:del` shape on the FINAL emitted document XML in both modes, plus both projections' spans.

## Real-corpus evidence

The two `#643` characterization pins in the real-corpus paragraph-deletion gate (`nvca-voting-agreement × inplace`, `nvca-rofr-co-sale-agreement × inplace`) flipped to passing against the SHA-256-verified corpus and are removed — exactly the signal #658 defined for this fix landing. Emitted NVCA voting output for the struck `Board Composition` clause is now:

```xml
<w:bookmarkStart w:name="_Ref444624639" w:id="30"/>
<w:bookmarkStart w:name="_Toc469383986" w:id="31"/>
<w:bookmarkStart w:name="_Ref145424752" w:id="32"/>
<w:del w:id="127">…Board Composition…</w:del>
<w:bookmarkEnd w:id="30"/><w:bookmarkEnd w:id="31"/>
<w:del w:id="140">…. Each Stockholder…</w:del>
```

The `#646` rebuild pins are untouched.

## The crossing-ranges wart

The issue's related wart — one range wholly inside a wrapper plus one partially inside coming out crossing — is normalized as a **side effect of the split**, not by a separate pass: every boundary keeps its position in the content stream, so ranges that were disjoint or nested in the input stay that way. Pinned by the new "come out in content order" test.

## Residuals (out of scope, verified pre-existing)

Both were surfaced by Codex peer review with executed fixtures and adjudicated by running the same fixtures against origin/main sources:

1. **Equal-content marker anchoring (inplace).** When a bookmarked source run is partly kept (e.g. `alpha removedOne beta removedTwo gamma` revised to `alpha inserted beta gamma`, bookmark around the whole original run), the cloned start re-anchors at the first deleted fragment rather than before the surviving equal prefix — the rejected projection covers `removedOne beta removedTwo gamma` but not `alpha`. On origin/main the SAME fixture collapses the range to zero length in BOTH projections (`accepted="" rejected=""`), so this PR strictly improves it. Faithful anchoring relative to equal/inserted fragments of the same source run is a marker-anchoring redesign, not a wrapper split.

2. **Revised-side partial `w:ins` boundary in rebuild mode.** A bookmark end that closes partway through newly inserted revised content is reconstructed past the insertion by the rebuild reconstructor before the consumer pass runs, so Accept All over-covers (`keptinsideoutside` instead of `keptinside`). Output is byte-identical on origin/main — untouched by this PR. The shared pass DOES split all four wrapper types when a marker reaches it inside a wrapper (unit-covered for `w:del`, `w:ins`, `w:moveFrom`, nested); this residual is upstream marker reconstruction, not the pass.

Not filing follow-up issues without discussion, per review protocol.

## Testing

- `packages/docx-compare/src/baselines/atomizer/consumerCompatibility-bookmark-ranges.test.ts`: the pinned lossy case now asserts the split shape (both `w:del` halves' content, attribute copy, fresh `w:id`); new cases for `w:moveFrom` splits, content-order (crossing) normalization, adjacent boundaries (no empty wrapper half), and three e2e cases asserting final emitted XML + both projections in both modes.
- Real-corpus matrix re-run against the verified NVCA cache: 15/15 pass with the pins removed.
- Full pre-submit chain green: build, lint:workspaces, test:run (all workspaces, corpus env set), check:spec-coverage, check:conformance-citations, check:conformance-doc.

Fixes: #643

### DIFF
--- a/packages/docx-compare/src/baselines/atomizer/consumerCompatibility-bookmark-ranges.test.ts
+++ b/packages/docx-compare/src/baselines/atomizer/consumerCompatibility-bookmark-ranges.test.ts
@@ -143,6 +143,21 @@ function rangeText(documentXml: string, id: string): string {
   return text;
 }
 
+const REVISION_WRAPPER_TAGS = ['w:ins', 'w:del', 'w:moveFrom', 'w:moveTo'] as const;
+
+/** Assert every emitted revision wrapper carries a unique `w:id`. */
+function expectUniqueWrapperIds(documentXml: string): void {
+  const body = bodyOf(documentXml);
+  const ids: string[] = [];
+  for (const tag of REVISION_WRAPPER_TAGS) {
+    for (const wrapper of Array.from(body.getElementsByTagName(tag)) as Element[]) {
+      const id = wrapper.getAttribute('w:id');
+      if (id !== null) ids.push(`${tag}#${id}`);
+    }
+  }
+  expect(new Set(ids).size).toBe(ids.length);
+}
+
 const ORIGINAL_WITH_BOOKMARKED_PARAGRAPH =
   textParagraph('Leading survivor') +
   bookmarkedParagraph('41', 'DeletedBoundary', 'Bookmarked deleted text', '<w:pPr><w:keepNext/></w:pPr>') +
@@ -372,6 +387,10 @@ describe('Bookmark ranges survive paragraph-level revisions', () => {
     await and('Accept All shrinks the range to the surviving text only', () => {
       expect(rangeText(acceptAllChanges(documentXml), '50')).toBe('kept');
     });
+
+    await and('every emitted revision wrapper keeps a unique w:id', () => {
+      expectUniqueWrapperIds(documentXml);
+    });
   });
 
   test('a boundary partway through newly deleted text closes between the deletions [inplace]', async (
@@ -409,6 +428,10 @@ describe('Bookmark ranges survive paragraph-level revisions', () => {
 
     await and('Accept All leaves the range without the deleted text', () => {
       expect(rangeText(acceptAllChanges(documentXml), '50')).toBe('');
+    });
+
+    await and('every emitted revision wrapper keeps a unique w:id', () => {
+      expectUniqueWrapperIds(documentXml);
     });
   });
 
@@ -458,6 +481,10 @@ describe('Bookmark ranges survive paragraph-level revisions', () => {
 
     await and('Accept All keeps the range over the surviving text only', () => {
       expect(rangeText(acceptAllChanges(documentXml), '50')).toBe('kept');
+    });
+
+    await and('the split half w:id does not collide with the pre-existing wrapper ids', () => {
+      expectUniqueWrapperIds(documentXml);
     });
   });
 });
--- a/packages/docx-compare/src/baselines/atomizer/consumerCompatibility-bookmark-ranges.test.ts
+++ b/packages/docx-compare/src/baselines/atomizer/consumerCompatibility-bookmark-ranges.test.ts
@@ -13,6 +13,11 @@
  * surviving content still gets that boundary anchored outside, because the
  * paragraph disappears in one of the two projections.
  *
+ * A boundary partway inside an inline revision wrapper splits the wrapper at
+ * the boundary (attributes copied, fresh `w:id` on the second half), so the
+ * range keeps its exact original span instead of shrinking or growing to the
+ * nearest wrapper edge (issue #643).
+ *
  * The second describe block drives the pass directly, because the boundary cases
  * it covers are hard to reach through a whole comparison: the atomizer is free
  * to align a fixture into a different revision shape than the one under test.
@@ -108,6 +113,34 @@ function paragraphChildTags(documentXml: string, text: string): string[] {
     }
   }
   throw new Error(`no paragraph containing ${JSON.stringify(text)}`);
+}
+
+/**
+ * Concatenated visible text (`w:t` / `w:delText`) lying between the
+ * `w:bookmarkStart` and `w:bookmarkEnd` of range `id`, in document order.
+ * This is the text a `REF` field aimed at the bookmark would resolve to.
+ */
+function rangeText(documentXml: string, id: string): string {
+  let inside = false;
+  let text = '';
+  const visit = (node: Element): void => {
+    for (const child of childElements(node)) {
+      if (child.tagName === 'w:bookmarkStart' && child.getAttribute('w:id') === id) {
+        inside = true;
+        continue;
+      }
+      if (child.tagName === 'w:bookmarkEnd' && child.getAttribute('w:id') === id) {
+        inside = false;
+        continue;
+      }
+      if (inside && (child.tagName === 'w:t' || child.tagName === 'w:delText')) {
+        text += child.textContent ?? '';
+      }
+      visit(child);
+    }
+  };
+  visit(bodyOf(documentXml));
+  return text;
 }
 
 const ORIGINAL_WITH_BOOKMARKED_PARAGRAPH =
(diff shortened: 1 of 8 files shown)


stevenobiajulu:

Codex CLI reviewed the diff dynamically (read all affected files, ran the cited suites — 17/17 bookmark-ranges, full `docx-compare` 705 passed, lint + both conformance checks clean — and executed its own fixtures through the pipeline). Verdict: **NEEDS CHANGES**, 2 high + 1 low. Adjudication below; each high finding was re-executed on this branch AND against origin/main sources to classify regression vs pre-existing.

### Finding 1 (high): trailing-clone forwarding with interleaved statuses in one source run — VALID OBSERVATION, PRE-EXISTING LIMITATION, STRICT IMPROVEMENT HERE

Codex fixture: bookmark around the whole original run `alpha removedOne beta removedTwo gamma`, revised to `alpha inserted beta gamma`. Deleted atoms of the run are NOT contiguous (del → ins → equal → del → equal → del), and the cloned start anchors at the first deleted fragment, so Reject All misses the leading `alpha`.

Executed both ways:

| | accepted | rejected |
|---|---|---|
| origin/main | `""` | `""` (range fully collapsed) |
| this PR | `"inserted beta "` | `"removedOne beta removedTwo gamma"` |

The PR strictly improves this shape; the residual is equal/inserted-fragment marker anchoring — a different mechanism than the wrapper split #643 asks for. Documented in the PR body under Residuals. Not reworked here: it is not a regression, and redesigning source-run marker ownership across all correlation statuses is its own change.

### Finding 2 (high): revised-side partial `w:ins` boundary in rebuild mode — PRE-EXISTING, BYTE-IDENTICAL ON MAIN

Codex fixture: revised bookmark closes between inserted `inside` and following `outside`; rebuild Accept All covers `keptinsideoutside`. Re-ran on origin/main sources: identical output (`accepted="keptinsideoutside" rejected="kept"`). The marker is repositioned by the rebuild reconstructor upstream, before `enforceConsumerCompatibility` — it never reaches the pass inside a wrapper, so the splitter is not the defective component. Per Codex's own alternative ("or explicitly narrow the PR's claims"), the PR body now states this residual explicitly. The pass itself splits all four wrapper types (unit-covered incl. `w:ins` and nested `w:ins`→`w:del`; Codex's own executed positive: partial `w:ins` inplace gave `accepted="keptinside" rejected="kept"`).

### Finding 3 (low): stale comment in `inPlaceModifier-bookmark-boundaries.test.ts` — VALID, FIXED

Comment updated to describe the split behavior instead of "still cannot be repositioned faithfully".

### Additional review asks

- **Wrapper-id uniqueness assertions**: added `expectUniqueWrapperIds` to all three e2e cases (both modes, incl. the pre-existing-id collision case where the split half allocated a fresh non-colliding id — Codex's own executed check: existing ids 1,2,3 → split id 4).
- **Real-corpus matrix with `SAFE_DOCX_REAL_CORPUS_REQUIRED=1`**: Codex could not run it (no `SAFE_DOCX_REAL_CORPUS_DIR` in its sandbox). Run here against the SHA-256-verified cache materialized by `scripts/prepare_real_comparison_corpus.mjs`: exit 0, **15/15 passed, 0 matrix cells skipped**, with the two #643 pins removed. CI's dedicated `Run real-corpus paragraph-deletion gate` job runs the same matrix with `SAFE_DOCX_REAL_CORPUS_REQUIRED=1`.

### Verified positives from the review (executed)

- Postprocess ordering as claimed; a same-author whitespace-bridged `w:ins` chain merged pre-pass then stayed split around the bookmark in final XML.
- Inplace id collision fixture: split id skips pre-existing ids.
- moveFrom fixture: bookmark stays between the correct fragments with one surviving outer move-range pair.

Issue #679, opened 2026-07-27
comparison: paragraph-property-only changes are not tracked, and reconstruction mode silently decides which formatting survives


stevenobiajulu:
## Summary

A change confined to paragraph properties is not reported by comparison, on **empty and non-empty paragraphs alike**. Restyling a paragraph Heading1 → Normal produces zero reported changes in both reconstruction modes. Worse, the reconstruction mode silently decides which formatting survives into the output — so the same input pair yields two different documents, neither of which records that anything changed.

## Reproduction

Two documents whose only difference is the `w:pStyle` of a paragraph, run through `compareDocuments` on `1caa9b1`:

**Empty paragraph, `Heading1` → `Normal`:**

```
inplace : insertions 0 / deletions 0 / modifications 0 / formatChanges 0
rebuild : insertions 0 / deletions 0 / modifications 0 / formatChanges 0
```

**Non-empty paragraph (same `w:pStyle` change, paragraph carries unchanged text):** also zero changes in both modes.

**Which formatting survives, for the empty-paragraph case:**

```
inplace : Heading1 absent, Normal present   → revised formatting wins
rebuild : Heading1 present, Normal absent   → original formatting wins
```

No `w:pPrChange` is emitted in either mode.

## Root cause

Format detection compares **run properties only**. `detectFormatChangesInAtomList` (`packages/docx-compare/src/format-detection.ts:281`) extracts `w:rPr` from the before/after atoms via `getRunPropertiesFromAtom`; there is no paragraph-property path. So `w:pStyle`, `w:numPr`, `w:jc`, `w:ind`, and `w:spacing` changes have nothing that detects them.

For empty paragraphs there is a second, independent reason the change is invisible: their atom identity hashes `w:pPr` shallowly, so paragraph-property children never reach the hash. That half is a distinct defect and is filed separately — fixing it alone would make empty-paragraph formatting produce whole-paragraph delete+insert markup while the identical change on a non-empty paragraph stays invisible, which is not a coherent behavior.

## Why this is a capability gap, not a small bug

The OpenSpec requirements describe format-change detection in terms of run properties. Extending to paragraph properties means deciding:

- **What markup represents the change.** `w:pPrChange` is the native OOXML mechanism and is what Word emits; the engine currently emits none. This is the correct target, but it is a new emission path, not a hash tweak.
[...]

PR #683, opened 2026-07-27
fix(docx-compare): exclude w:pPr from empty-paragraph atom identity

### LINKED ISSUE #678: atomizer: empty-paragraph identity treats bare w:pPr and namespace-declaration placement as content
## Summary

Empty-paragraph atom identity hashes `w:pPr` with `hashElement`, which is **shallow** — `elementIdentityString` (`packages/docx-compare/src/atomizer.ts:78`) reads an element's tag name, its own DOM attributes, and its leaf text, with no recursion into children. Applied to `w:pPr`, whose meaning lives entirely in its children, this makes identity sensitive to things that carry no formatting meaning and blind to the things that do.

Two concrete phantom-diff sources, both reproduced end-to-end:

1. **Bare `w:pPr` presence counts as content.** `<w:p/>` and `<w:p><w:pPr/></w:p>` are formatting-equivalent but compare as different.
2. **Namespace-declaration placement counts as content.** `xmlns:w` appears in the DOM `attributes` collection, so a paragraph that inherits the binding and one that redundantly redeclares it locally compare as different. This is legal XML — the earlier assumption that only non-conforming input could put an attribute on `w:pPr` is wrong.

## Where

`packages/docx-compare/src/atomizer.ts`, empty-paragraph atom construction (~`:695`):

```ts
const pPr = findChildByTagName(paragraphElement, 'w:pPr');
const pPrHash = pPr ? hashElement(pPr) : 'no-pPr';
const contextHash = state.lastContentHash || 'document-start';
const hashContent = `empty-paragraph:${contextHash}:${state.consecutiveEmptyIndex}:${pPrHash}`;
```

`hashElement` (`atomizer.ts:111`) is exactly `sha1(elementIdentityString(element))`. `getLeafText` (`packages/docx-core/src/primitives/dom-helpers.ts:34`) reads direct text nodes only. There is no downstream independent `w:pPr` comparison — the coarse empty-paragraph matcher consumes this identity strictly.

## Reproduction

Both cases, run through `compareDocuments` on `1caa9b1`:

**Bare vs absent `w:pPr`** — `<w:p/>` vs `<w:p><w:pPr/></w:p>`, following an identical text paragraph:

```
inplace : insertions 1 / deletions 1 / insertedAtoms 1 / deletedAtoms 1
rebuild : insertions 1 / deletions 1 / insertedAtoms 1 / deletedAtoms 1
```

**Redundant `xmlns:w` redeclaration only** (inherited binding vs locally redeclared on `w:pPr`):

```
inplace : insertions 1 / deletions 1  (paragraph-mark insert + delete)
rebuild : insertions 1 / deletions 1, plus a w:pPrChange
```

Both are pure serialization topology. Neither changes rendering, and both produce a paragraph-mark delete+insert pair in the redline.

A scan of the 54 checked-in `.docx` packages (706 XML parts, 9,498 `w:pPr` elements) found 752 `w:pPr` elements carrying an attribute — every one of them `xmlns:w`, none of them rsids. So the redundant-declaration shape is already present in this repository's own corpus.

## Conformance

- `CT_PPrBase` has **no attributes** in the vendored strict and transitional schemas (`spec-compliance/ecma-376/schemas/{strict,transitional}/wml.xsd`), and `CT_ParaRPr` has none either.
- `CT_P` owns `rsidRPr`, `rsidR`, `rsidDel`, `rsidP`, `rsidRDefault` — paragraph rsids live on `w:p`, not `w:pPr`.

So the only attributes reaching `elementIdentityString(pPr)` in practice are namespace declarations, which are serialization artifacts rather than content.

## Current behavior is pinned by a test

`packages/docx-compare/src/atomizer.test.ts:925`, `pins shallow paragraph property hashing for empty paragraphs`:

```ts
expect(noPPrHash).not.toBe(barePPrHash);         // presence matters
expect(barePPrHash).not.toBe(attributedPPrHash); // pPr attributes matter
expect(barePPrHash).toBe(childPPrHash);          // pPr children are ignored
```

The `attributedPPrHash` fixture is `el('w:pPr', { 'w:rsidR': '00112233' })` — a shape the schema does not permit, so the test pins discrimination on something that cannot legally occur while pinning *non*-discrimination on the children that actually distinguish two paragraphs. The test name says "pins", which reads as characterization of existing behavior rather than a designed invariant; this issue proposes replacing it with assertions that follow from a stated principle.

## Fix direction

The right precedent is **not** `hashElementDeep` in `premergeRuns.ts:45` — that includes all descendant attributes, so reusing it would make revision provenance participate in identity. Verified through its real `premergeAdjacentRuns` caller: two runs differing only in `w:id`/`w:date` inside valid `w:rPrChange` descendants failed to merge (`merges: 0`), which is exactly the phantom-diff class #675 removed.

The better precedent already in this package is the canonicalizer at `packages/docx-compare/src/baselines/atomizer/formattingFidelity.ts:99`, which hashes recursively while removing revision-property subtrees, explicitly ignoring `xmlns` declarations, and excluding `w:sectPr`. Extracting and reusing that canonicalization principle is the suggested route.

Note the revision path inside paragraph properties is `w:pPr/w:rPr/w:ins|w:del` — not every run-level `w:rPr` permits those children.

## Acceptance criteria

1. `<w:p/>` and `<w:p><w:pPr/></w:p>` compare equal.
2. Inherited versus redundantly redeclared namespace bindings compare equal.
3. Revision metadata under `w:pPr` (`w:pPrChange`, `w:pPr/w:rPr/w:ins|w:del`, and their `w:id` / `w:date` / `w:author`) does not affect identity.
4. Whether substantive `w:pPr` children belong in empty-paragraph identity is decided **explicitly** and written down, rather than falling out of the hash's shape. If they do not belong here, they belong to paragraph-format detection — see the companion issue.
5. The pin test is replaced by one whose assertions follow from the stated principle.
6. Real-document stats measured before and after (NVCA COI pair plus a template-heavy document), showing no new phantom markup.

## Provenance

Surfaced during review of #677 (fix for #675, rsid attributes in run-merge identity) and recorded there as a follow-up candidate. The original note described this as "`elementIdentityString` hashes `pPr` rsids into empty-paragraph identity"; that framing is inaccurate — `w:pPr` has no rsid attributes to hash — and the substantive defect is the shallow-hash inversion above. Codex peer review then corrected a second framing error: the attribute-sensitivity is live through legal namespace declarations, not merely hypothetical.

@see #675, #677, #409

stevenobiajulu:
## Summary

Empty-paragraph atom identity hashed `w:pPr` with the shallow `hashElement` (tag name + own DOM attributes + leaf text, no child recursion). Applied to `w:pPr` — whose meaning lives entirely in its children, and on which OOXML `CT_PPrBase` permits no attributes — this made identity sensitive to serialization topology and blind to content.

Reproduced phantom sources, before → after (`compareDocuments`, both `inplace` and `rebuild`):

| Case | Before (both modes) | After (both modes) |
|---|---|---|
| `<w:p/>` vs `<w:p><w:pPr/></w:p>` | insertions 1 / deletions 1 (+ `w:pPrChange` in rebuild) | 0 / 0, no `w:pPrChange` |
| Inherited vs locally redeclared `xmlns:w` on `w:pPr` | insertions 1 / deletions 1 (+ `w:pPrChange` in rebuild) | 0 / 0, no `w:pPrChange` |
| Revision-metadata-only delta (`w:pPr/w:rPr/w:ins` `w:id`/`w:author`/`w:date`) | 0 / 0 (children already ignored) | 0 / 0, now guarded by tests |

A fourth phantom source surfaced during verification: pretty-printed vs compact `w:pPr` serialization (whitespace text nodes reach `getLeafText`). In the checked-in `nvca-regression` pair this noise alone made **every** pPr-bearing empty paragraph compare unequal between source and filled — all were phantom delete+insert pairs.

## The fix (three coupled changes)

**1. `atomizer.ts` + `formattingFidelity.ts` — canonical positional identity.** Empty-paragraph identity is now the structural container chain (`w:tbl`/`w:tr`/`w:tc`/`w:txbxContent`), the preceding-content signature, the consecutive-empty index, and a **canonical projection of `w:pPr`** (`canonicalizeParagraphPropertiesForIdentity`, built on the `formattingFidelity.ts` canonicalizer the issue cites as precedent). The projection erases serialization topology — namespace declarations, inter-element whitespace, child order, attributes on `w:pPr` itself, bare-vs-absent `w:pPr` — and revision provenance (`w:pPrChange`, `w:pPr/w:rPr/w:ins|w:del`, plus the `w:rPr` shell left empty after their removal). Substantive children remain identity-distinguishing. `w:sectPr` participates as a **presence marker**: a section break's existence is document topology and must distinguish, but its content (page geometry, header/footer bindings via package-local `r:id`s) belongs to the section dimension — the same split `formattingFidelity` already draws — and equivalent breaks routinely differ in `r:id`s across packages, so content-sensitive identity would turn every both-sided section break into a delete+insert pair whose cloned original `r:id`s cannot resolve in revised-based output (verified: it broke the ILPA inplace pipeline through the ancillary binding gate, `sectpr_reference_wrong_relationship_type`).

**2. `atomLcs.ts` — unified-index remap for revised twins (latent bug, exposed).** For Equal empty-paragraph pairs, `createMergedAtomList` keeps the ORIGINAL atom, so the revised twin never flowed through the unified-index remapping loop and kept its raw revised-tree paragraph index. The inplace modifier keys revised paragraphs by unified index (`unifiedParaToElement`), so lookups for Equal empties either missed (stale insertion anchor) or collided with a *different* paragraph's unified index. Pre-fix this was masked on real documents because serialization noise prevented empty paragraphs from matching Equal at all; once they matched, the NVCA structural-regression pair mis-anchored a deleted `]]` paragraph after the wrong table and fell back to rebuild. Verified by instrumented runs: `unifiedParaToElement.get(pIdx)` returned UNDEFINED for all seven newly-Equal tranche-table empties.

**3. `inPlaceModifier-containers.ts` — body-anchor hoisting (latent bug, exposed).** A deleted body-level paragraph whose insertion anchor is a paragraph inside a table cell (an Equal cell empty processed just before it) was inserted as the anchor's sibling — i.e., INTO the cell. The anchor is now hoisted to the enclosing direct child of `w:body` (the table), keeping the deleted paragraph at body level in the same document-order slot (verified on the NVCA `]]` deletion: `insideTc=1` before, `insideTc=0` after).

## Acceptance criterion 4: substantive `w:pPr` children are IN identity (canonically)

Decided explicitly, and revised after Codex peer review of the first cut (which had excluded `w:pPr` entirely):

- **Serialization topology and revision provenance never distinguish** — that is the phantom class this issue was filed for (criteria 1–3).
- **Substantive property children always distinguish; `w:sectPr` distinguishes by presence.** Executed probes on the first cut showed why the alternative is unsafe: with `w:pPr` fully ignored, a section-break empty paragraph compared Equal to a plain empty — inplace dropped the section break while rebuild kept it (reconstruction mode silently deciding document topology, unrepresentable in the redline) — and an empty paragraph gaining `w:jc="center"` kept the revised formatting under inplace but lost the paragraph under rebuild, silently. As delete+insert markup the difference is representable: accept yields the revised properties, reject restores the original's, identically in both modes.
- The residual asymmetry — a pPr-only edit on an *empty* paragraph produces whole-paragraph delete+insert markup while the identical edit on a *non-empty* paragraph produces no markup — is the pre-existing paragraph-property change-detection gap. Representing such deltas as `w:pPrChange` for empty and non-empty paragraphs alike is #679's scope and is not expanded into here.

## Tests

- Replaced the characterization pin (`pins shallow paragraph property hashing for empty paragraphs`, whose `attributedPPrHash` fixture used a `w:rsidR` attribute on `w:pPr` that `CT_PPrBase` does not permit) with a principle-stated test: absent / bare / xmlns-redeclared / revision-marked `w:pPr` all hash identically, while a different content context, a table-cell container, a substantive child, and `w:sectPr` presence each distinguish (while differing section content does not — presence marker).
- E2e phantom regression tests (3 serialization pairs × both modes): zero changes, no comparison-authored paragraph revision markup, substantive formatting survives whichever side's serialization is kept, and accept/reject text projections round-trip to revised/original.
- E2e substantive-distinction tests (× both modes): section-break-vs-plain and absent-vs-`w:jc` empty pairs produce 1 insertion / 1 deletion; accept resolves to the revised properties and reject restores the original's.
- `atomLcs.test.ts`: unit regression asserting the revised twin of an Equal empty paragraph is remapped to the merged atom's unified index.
- E2e regression asserting a deleted body paragraph following a table of Equal empty cells is emitted at body level, not inside a `w:tc`.

## Real-document measurements (AC 6)

`tests/test_documents/nvca-regression` (pinned by the structural-regression suite):

| Comparison | Before (1caa9b1) | After |
|---|---|---|
| source vs filled, inplace | inplace, insertions 165 / deletions 352 (every pPr-bearing empty a phantom del+ins pair) | inplace, insertions 136 / deletions 323; `]]` deletion correctly anchored at body level before EXHIBIT A-1 |

`tests/test_documents/redline` ILPA pair (feature-rich, header/footer-heavy):

| Comparison | Before (1caa9b1) | After |
|---|---|---|
| WOF_v2 vs Deal-By-Deal_v1, inplace | inplace, no fallback | inplace, no fallback; insertions 93 / deletions 87 |

`tests/test_documents/nvca-coi-regression`:

| Comparison | Before (1caa9b1) | After |
|---|---|---|
| source vs filled, inplace | insertions 205 / deletions 375 / insertedAtoms 336 / deletedAtoms 13741 | insertions 203 / deletions 373 / insertedAtoms 334 / deletedAtoms 13739 |
| source vs filled, rebuild | throws `OpaquePassthroughError` (boundary count 135/98) — pre-existing | identical pre-existing throw, unchanged |
| source vs minimally-edited revision, inplace + rebuild | 1 / 1 | 1 / 1 (unchanged) |

Fixes: #678

@see #675, #677, #679


stevenobiajulu:
## Codex peer review: verdict + adjudication

An executed (dynamic) Codex review of the first cut of this branch returned **REQUEST-CHANGES** with two blocking findings, one should-fix, and one nit. All are addressed in the follow-up commit; both blockers were accepted as correct.

### Blocking 1 — `w:sectPr` section breaks compared Equal to plain empty paragraphs — ACCEPTED, fixed

With `w:pPr` fully excluded from identity (the first cut), a section-break-carrying empty paragraph matched a plain empty. The reviewer's probe showed inplace dropped the section break while rebuild kept it — reconstruction mode silently deciding document topology, with the divergence unrepresentable in the redline. Fixed by the canonical `w:pPr` projection, which keeps substantive children identity-distinguishing and `w:sectPr` distinguishing **by presence**. Presence (not content) is the right grain: including the section's content in identity re-broke the ILPA pair — both documents carry equivalent section breaks under different header/footer `r:id`s, so content-sensitive identity turned them into delete+insert pairs whose cloned original `r:id`s failed the ancillary binding gate (`sectpr_reference_wrong_relationship_type` at section ordinals 2/4/6) and forced a rebuild fallback. With the presence marker, ILPA compares inplace without fallback (insertions 93 / deletions 87), while the asymmetric section-break case stays visible markup. Re-executed probe now shows, in **both** modes: 1 insertion / 1 deletion, `w:sectPr` present in the redline, gone after accept-all, restored after reject-all:

```
{"case":"sectPr-to-plain","mode":"inplace","ins":1,"del":1,"out_sectPr":1,"acc_sectPr":0,"rej_sectPr":1}
{"case":"sectPr-to-plain","mode":"rebuild","ins":1,"del":1,"out_sectPr":1,"acc_sectPr":0,"rej_sectPr":1}
```

### Blocking 2 — substantive pPr additions silently lost in rebuild — ACCEPTED, fixed

The reviewer's probe showed an empty paragraph gaining `w:jc="center"` kept the revised formatting under inplace but lost the whole paragraph under rebuild, with `unrepresentedChanges: null` — contradicting the PR's preservation assumption. The acceptance-criterion-4 decision is therefore revised: substantive `w:pPr` children are IN identity (canonically), so genuine property differences pair as delete+insert and both modes represent them identically. Re-executed probe:

```
{"case":"absent-to-jc","mode":"inplace","ins":1,"del":1,"acc_jc":1,"rej_jc":0}
{"case":"absent-to-jc","mode":"rebuild","ins":1,"del":1,"acc_jc":1,"rej_jc":0}
```

Serialization topology (bare-vs-absent pPr, xmlns declarations, whitespace, child order) and revision provenance still never distinguish — the three phantom pairs this issue was filed for still compare 0 / 0 in both modes. Representing property deltas as `w:pPrChange` instead of delete+insert remains #679's change-detection scope.

### Should-fix — tests didn't assert which side's `w:pPr` survives — ACCEPTED, fixed

The phantom-pair e2e tests now assert surviving substantive formatting (`w:jc` present in the redline and in both projections), and that accept/reject text projections round-trip to the revised/original documents. New e2e tests cover the `w:sectPr` and absent-to-`w:jc` distinctions in both modes.

### Nit — stale test count and line number — ACCEPTED

PR body updated.

### Post-fix measurements (all re-executed after the changes and the merge of origin/main @ 5a4a314)

- Phantom pairs: 0 / 0 in both modes, no `w:pPrChange`, all three cases.
- `nvca-regression` source vs filled, inplace: no safety fallback; 165/352 → 136/323 insertions/deletions vs base; `]]` deletion at body level before EXHIBIT A-1.
- `nvca-coi-regression` source vs filled, inplace: 205/375 → 203/373.
- Full pre-submit gate chain green.

stevenobiajulu:
Note for anyone arriving here from `git log` or `git blame`: **the squash subject describes the first cut, not what shipped.**

`08240cc` reads `fix(docx-compare): exclude w:pPr from empty-paragraph atom identity`. What actually shipped *includes* `w:pPr` in identity, via a canonical projection. The squash body carries both commit messages, and the second one (`canonicalize w:pPr in empty-paragraph identity per review`) is the accurate account — but the subject is what shows in `--oneline`, so it is worth stating plainly:

**Shipped design.** `canonicalizeParagraphPropertiesForIdentity` (`packages/docx-compare/src/baselines/atomizer/formattingFidelity.ts`) projects `w:pPr` to a canonical string that empty-paragraph atom identity hashes.

- **Never distinguishes** — serialization topology (bare vs absent `w:pPr`, `xmlns` redeclarations, inter-child whitespace, child order, attributes on `w:pPr` itself, which `CT_PPrBase` permits none of) and revision provenance (`w:pPrChange`, `w:ins`/`w:del`, and the empty `w:rPr` shell their removal leaves behind).
- **Always distinguishes** — substantive property children, canonically. Genuine differences pair as delete+insert and both reconstruction modes represent them identically.
- **`w:sectPr` distinguishes by presence only.** Content-sensitive identity re-broke the ILPA pair: equivalent section breaks under different header/footer `r:id`s became delete+insert pairs whose cloned `r:id`s failed `sectpr_reference_wrong_relationship_type` and forced rebuild fallback. Presence is topology; section content belongs to the section dimension.

The first cut's rationale — "substantive `w:pPr` children stay OUT of identity by explicit decision" — was rejected by Codex peer review with two executed blocking findings (a section break comparing Equal to a plain empty; `w:jc="center"` surviving inplace but not rebuild). Both are fixed. The in-code JSDoc has always matched the shipped design; only the commit subject does not.

Paragraph-property *change detection* remains out of scope here and is #679.

Issue #684, opened 2026-07-27
test(docx-compare): formatting-loss D1 compares declared formatting, not formatting resolved through styles.xml


stevenobiajulu:
`scripts/check_docx_formatting_loss.mjs` (added in #682) projects each character
of a paragraph onto its `(bold, italic, underline, character style)` tuple, read
from direct `w:rPr` plus the `w:rStyle` reference. It does not resolve
formatting through `word/styles.xml`.

Including the `w:rStyle` reference closes the dangerous direction — dropping a
style that carried the emphasis changes the reference, so the loss is caught
rather than silent. Two gaps remain:

- Emphasis arriving through paragraph-style inheritance is invisible.
- Editing a style's *definition* while keeping the reference is invisible.
- Replacing a style with equivalent direct properties reports a change that is
  arguably a representation difference rather than a loss.

`extractEffectiveRunFormatting` in `packages/docx-core/src/primitives/styles.ts`
already resolves the `w:rStyle` chain and paragraph-style contribution. It is
not currently part of docx-core's public entry point, and a root-level
diagnostic script deep-importing built package internals would couple the script
to workspace build order.

Proposed: export the resolver from docx-core's public surface, then have the
detector consume it instead of reading direct properties, so there is one
effective-formatting implementation rather than two that can drift.

The limitation is documented in `docs/comparison-failure-diagnosis.md` under
"Detecting Formatting Loss" until this lands.

Refs #682

stevenobiajulu:
Heads-up on provenance: this issue describes `scripts/check_docx_formatting_loss.mjs` as "added in #682" and cites `docs/comparison-failure-diagnosis.md`, but neither exists on `main` (`dbdb6ef`) — `git ls-tree` finds no such script, and `docs/` contains only `comparisons.md`. #682 is still open, so this is written against unmerged work and will read as stale to anyone picking it up cold. Worth either landing #682's script first or restating this issue so it stands alone.

Separately: the effective-formatting resolution proposed here (`extractEffectiveRunFormatting`) is also the dependency for an edit-time convention check — filed as #687 — so it is worth exporting with both callers in mind rather than only the diagnostic script.

PR #685, opened 2026-07-27
test(docx-compare): add structural formatting-loss detectors and a stated pass criterion

### LINKED ISSUE #682: test(docx-compare): comparison diagnostics miss run-level formatting loss, state no pass criterion, and run without a control arm
## Summary

Our comparison-diagnosis procedure has three gaps that let a run produce a
confident but unfounded verdict. All three surfaced while diagnosing a real
contract pair rather than the synthetic corpus.

## 1. The failure taxonomy omits run-level formatting loss

Diagnostics currently classify failures as: a throw, a phantom, silent content
loss, or a degraded-but-valid redline. Missing is the class where **character
formatting is destroyed while text is preserved**. It is invisible to every
check we routinely run — extracted text is byte-identical, round-trip reports
success, and the document opens cleanly in Word.

Two observed shapes:

- A replacement whose span crosses a run boundary collapses the boundary and
  drops bold/italic from the affected span.
- Replacing a block leaves emptied paragraph shells. An empty body-text
  paragraph renders as a blank line; an empty list paragraph that retains its
  `w:numPr` renders an **orphan numbered label**, which a reader sees and a text
  diff does not.

Two cheap structural detectors catch both, and neither needs document text:

- **D1** — per paragraph, build a multiset of runs keyed on
  `(text, bold, italic, underline)`. If the concatenated text is unchanged but
  the multiset changed, formatting was flattened.
- **D2** — diff paragraphs by `w14:paraId`. Flag any that were non-empty before
  and empty after, plus any empty paragraph still carrying `w:numPr`.

Related but distinct: #679 covers paragraph-*property* changes not being
tracked. This is run-level character formatting.

## 2. No pass criterion is stated, so "it worked" is unfalsifiable

A diagnostic run should have to clear all of:

- round-trip identity of extracted text, with the one documented exception that
  text inside `w:txbxContent` is excluded when text-box comparison is disabled;
- `PAGEREF` field count preserved (this is how table-of-contents destruction
  shows up);
- zero `w:ins` / `w:del` markers in clean output;
- both detectors above reporting zero;
- `reconstructionModeUsed` reported, not merely requested — the library defaults
  to `rebuild` while the MCP tool hardcodes `inplace`
  (`packages/docx-mcp/src/cli/commands/compare.ts:24`), so the same input can
  succeed through one entry point and abort through the other.

## 3. Runs have no control arm

Exercising only the suspect input tells you what happened; it does not tell you
whether the instrumentation can detect anything. A known-bad input run through
the same harness converts "no failure reproduced" from ambiguous into
trustworthy. We have no standing known-bad fixture for the run-split abort
(`boundary 0 changed paragraph ownership`), so today that check cannot be run at
all.

## Adjacent process finding

Synthesise-from-scratch is currently the default route to a shareable repro, and
it has failed twice: four hand-built fixtures for the run-split defect all
passed, and a 20-control synthetic round-tripped cleanly on #660. Scrub-in-place
— lorem tokens with markup, fields, `w:sdt`, numbering, and run boundaries
preserved — is the method with a track record, because these defects live in
structure rather than in text. A synthetic that does not reproduce is not
evidence that there is no bug.

## Proposed

- Add D1/D2 as reusable structural checks and wire them into the diagnostic
  procedure.
- Codify the pass criterion above.
- Build and check in a minimal known-bad fixture for the run-split abort so a
  control arm is always available.

stevenobiajulu:
Closes #682.

## Problem

A comparison run that produces no visible error is not the same as one that
succeeded. Extracted text can come back byte-identical, round-trip can report
success, and the output can open cleanly in Word while a defect a reader
notices on the first page is still there.

The failure taxonomy had no class for formatting destroyed while every
character survives, so that class passed every check the diagnostic procedure
routinely runs, and shipped twice unnoticed. Two shapes: a replacement whose
span crosses a run boundary flattens bold off a defined term; replacing a block
leaves an emptied paragraph shell that renders a blank line, or an orphan
numbered label when it keeps its `w:numPr`.

There was also no stated pass criterion, which made "it worked" unfalsifiable,
and no control arm, which made "no failure reproduced" ambiguous.

## What this adds

`scripts/check_docx_formatting_loss.mjs` — two structural detectors over a
before/after pair, plus `docs/comparison-failure-diagnosis.md` stating the
five-part pass criterion, the control-arm requirement, and the scrub-in-place
repro policy.

- **D1** projects each character of a paragraph onto its
  `(bold, italic, underline, character style)` tuple and fires when the text is
  unchanged but the projection is not.
- **D2** matches paragraphs on `w14:paraId` and flags those emptied between the
  two sides, including emptied paragraphs that kept their numbering.
- `--self-test` proves the detectors fire on a known-bad pair and stay silent on
  a known-good one. It runs in CI. A detector never shown to fire cannot be told
  apart from one that did not run.

Detector reports emit counts, `w14:paraId` values, and element names only, and
the projection keeps a digest of each paragraph's text rather than the text, so
a run over material that cannot be shared still produces reportable output.

## Two deliberate deviations from the issue

**D1 is a per-character projection, not the multiset of runs the issue
sketches.** A multiset keyed on `(text, bold, italic, underline)` fires whenever
a run boundary moves, including moves that preserve every character's emphasis
— which this codebase produces routinely via atomizer token splitting and rsid
churn (#677). On a 1500-paragraph document that false-positive class would swamp
the signal. Regression test included.

**Duplicate `w14:paraId` values are excluded from matching, not resolved.** The
issue's sketch keys a dictionary on `paraId`, which resolves duplicates by last
write. Real documents carry them — a real contract pair measured 17 and 19 per
side — so those paragraphs are counted as uncompared instead.

## Peer review found a defect that would have made the tool useless

Codex executed the detectors against real engine output rather than reading the
diff, and demonstrated that **`reconstructionMode: 'rebuild'` emits output
carrying no `w14:paraId` at all**. Independently reproduced: 1 paragraph in, 0
`paraId` attributes out. Every paragraph then failed to match, every detector
reported zero, and the CLI exited 0 — a clean pass over a document it had never
read, which is exactly the failure mode this change exists to prevent. Pass
criterion item 5 requires running rebuild, so the criterion was self-defeating
as first written.

Coverage is now enforced rather than merely reported: below 95% the result is
`INCONCLUSIVE` and exits 2, with `--min-coverage` to move the floor
deliberately. The real pair measures 97% and stays conclusive.

The same review also caught: D1 missing emphasis dropped via `w:rStyle`
removal; D2 flagging unchanged empty numbered paragraphs, so an untouched
document compared to itself exited 1; image-only paragraphs read as empty; and
the tool silently accepting redlines, non-`w:document` parts, and unknown
flags. All fixed with tests against their reproductions.

## Known limits, stated rather than hidden

- D1 compares *declared* formatting — direct `w:rPr` plus the `w:rStyle`
  reference — not formatting resolved through `styles.xml`. Paragraph-style
  inheritance and edits to a style definition are invisible. #684 tracks
  consolidating onto docx-core's existing effective-formatting resolver rather
  than growing a second one here.
- D1 requires identical text, so loss co-occurring with a text edit in the same
  paragraph is out of reach.
- The `--self-test` is a synthetic control over the detector logic, not over the
  comparison engine. The guide's own warning about synthetic fixtures applies to
  it.

## Verification

```
node --test scripts/check_docx_formatting_loss.test.mjs   27 pass, 0 fail
npm run test:docx-formatting-loss                         exit 0
npm run build                                             exit 0
npm run lint:workspaces                                   exit 0
npm run check:control-bytes                               exit 0
```

Also exercised end-to-end against a real 1500-paragraph document pair (0.9s),
and against real `compareDocuments` output in both reconstruction modes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,6 +77,8 @@ jobs:
         run: npm run check:release-isolation
       - name: Test release-isolation guard
         run: npm run test:release-isolation
+      - name: Test formatting-loss detectors (unit tests + known-bad self-test)
+        run: npm run test:docx-formatting-loss
 
   spec-coverage:
     runs-on: ubuntu-latest
--- a/.gitignore
+++ b/.gitignore
@@ -72,3 +72,4 @@ integration-tests/OPENSPEC_TRACEABILITY.md
 *.mcpb
 *.dxt
 site/src/trust/system-card.md
+.peer-review/
--- a/docs/comparison-failure-diagnosis.md
+++ b/docs/comparison-failure-diagnosis.md
@@ -1,0 +1,70 @@
+# Comparison Failure Diagnosis
+
+A comparison run that produces no visible error is not the same as a comparison run that succeeded. Extracted text can come back byte-identical, the round-trip can report success, and the output can open cleanly in Word while a defect a reader would notice on the first page is still present. This page states the failure classes a diagnosis has to distinguish, the criterion a run has to clear before it counts as a pass, and the evidence a negative result needs to be believable.
+
+## Failure Classes
+
+| Class | What it looks like | How it is caught |
+|---|---|---|
+| Throw | `OpaquePassthroughError`, a field-structure validation error, or a round-trip safety abort | The stack trace, plus the offending node's ancestor chain |
+| Phantom markup | Unchanged text struck and reinserted — `w:delText` and `w:t` carrying the same characters | Scan the output for adjacent `w:delText` / `w:t` pairs with matching text |
+| Silent content loss | Content present in the input and absent from the output | Round-trip identity of extracted text |
+| Degraded redline | Comparison succeeds and the redline is unusable | Inspection |
+| Formatting loss | Character formatting destroyed while every character survives | [The structural detectors below](#detecting-formatting-loss) |
+
+The last class is the one that hides. It passes text-level checks by construction, because no text changed. Two shapes have been observed:
+
+- A replacement whose span crosses a run boundary collapses the boundary and drops bold or italic from the affected span, so a defined term loses its emphasis.
+- Replacing a block leaves paragraph shells behind. An empty body paragraph renders as a blank line. An empty list paragraph that kept its `w:numPr` renders an orphan numbered label, which a reader sees and a text diff does not.
+
+## Detecting Formatting Loss
+
+`scripts/check_docx_formatting_loss.mjs` compares a before/after `.docx` pair and reports two detectors:
+
+- **D1 — run-formatting flattening.** Each character of a paragraph is projected onto its `(bold, italic, underline)` triple. When a paragraph's text is unchanged but that projection changed, emphasis was flattened. The projection is compared rather than a multiset of runs so that a run boundary moving without changing any character's emphasis — which token splitting and rsid churn produce routinely — is not reported as loss.
+- **D2 — emptied-but-retained paragraphs.** Paragraphs are matched on `w14:paraId` and flagged when they carried text before and carry none after, plus any empty paragraph still carrying `w:numPr`.
+
+```bash
+npm run check:docx-formatting-loss -- before.docx after.docx   # exit 1 on findings
+npm run check:docx-formatting-loss -- --json before.docx after.docx
+npm run test:docx-formatting-loss                              # unit tests + self-test
+```
+
+The tool emits counts, `w14:paraId` values, and element names, and never document text, so its output can be reported from a run over material that cannot be shared. The projection holds a digest of each paragraph's text rather than the text.
+
+Two limits the output states rather than hides. Paragraphs without a `w14:paraId` cannot be matched at all; they are counted and named in the coverage line, and a zero finding does not cover them. Duplicate `paraId` values are dropped from the match set rather than resolved by last write, for the same reason. The emphasis key is deliberately narrow — bold, italic, underline — so losses confined to other run properties pass unreported.
+
+## Pass Criterion
+
+A diagnostic run is a pass only when all five hold. Anything short of all five is a fail and should be reported as one.
+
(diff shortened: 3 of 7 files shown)


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `dbdb6ef`
Built from: `main` @ `dbdb6ef` (clean throwaway worktree, fresh `npm install`)
Smoke: `npm run build && npm run lint:workspaces && npm run test:run` + real-document end-to-end

**Steps**
- ✅ build (exit 0)
- ✅ lint (exit 0)
- ✅ tests — 2476 passed across 6 packages (`121+96+10+19+2+2` test files, 2 expected-fail, 49 skipped)
- ✅ `npm run test:docx-formatting-loss` (27 unit tests + known-bad self-test)
- ✅ detectors driven over real `compareDocuments` output in both reconstruction modes

**Real-world fixture**
- `nvca-indemnification-agreement` — real cached source DOCX (70,578 bytes, 153 paragraphs), not a synthetic build
  - Real edit: 222 occurrences of a defined term replaced across the agreement, the cross-run shape that produced the original defect
  - `compareDocuments` → redline (222 `w:ins` / 222 `w:del`) → `acceptChanges()` → clean output, 0 revision markers remaining
  - **inplace**: coverage 153/153 (100.0%), D1 = 0, D2 = 0, orphan numbering = 0 → conclusive pass, exit 0
  - **rebuild**: 153/153 paragraphs emitted with no `w14:paraId` → coverage 0.0% → `INCONCLUSIVE`, exit 2

That rebuild line is the point of the whole review cycle. Before the coverage gate, that run printed three zeros and exited 0 — a clean pass over a real 153-paragraph legal agreement the tool had not read a single paragraph of. It now refuses to answer, on real engine output rather than only on the unit fixture.

**Advisory gate audit** — `Aggregate and post review`: **Overall PASS** (3 pass, 0 warn, 0 error, 14 skipped). No `Unable to verify` / provider-failure markers; the gate genuinely ran.

Visual evidence: skipped — this change adds a diagnostic script and documentation and alters nothing about how a document renders.
Prod: ✅ Production deployment for `dbdb6ef` reported `success`.
Log: `/tmp/automerge-smoke-685-1785191790.log` (local).

**Cleanup**
- ⚠️ remote branch already deleted — repo has auto-delete enabled
- ✅ removed worktree `.worktrees/682-formatting-loss-detectors-20260727` and the throwaway smoke worktree
- ✅ deleted local branch `682-formatting-loss-detectors-20260727`

Follow-up #684 remains open for resolving formatting through `styles.xml` rather than reading declared properties.

PR #690, opened 2026-07-27
test(docx-compare): profile corpus outcomes by runtime

stevenobiajulu:
## Summary

- separate the real-corpus failure fingerprints common to CI and local development from three Darwin/Node 26-only fingerprints
- pin required CI to the characterized Linux/Node 20 profile so runtime upgrades require deliberate re-characterization
- cover profile selection directly while preserving the two-way behavior gate

## Why

PR #671 merged while its newly added real-corpus job was still non-blocking. That job showed three cells pass on Linux/Node 20 but fail on Darwin/Node 26. A single expected-outcome map therefore made the gate non-portable even though both runtimes were stable.

## Validation

- full 14-cell SHA-256-pinned corpus matrix on Darwin/Node 26
- docx-compare TypeScript lint and no-cache skip path
- repository build and workspace lint
- docx-core: 1,238 passed, 2 expected failures, 12 skipped
- spec coverage and all conformance checks

Follow-up to #671.
Ref: #658

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -408,6 +408,43 @@ jobs:
             packages/docx-core/allure-results/
             packages/docx-mcp/allure-results/
 
+  real-corpus-comparison:
+    name: Real-corpus paragraph deletion
+    runs-on: ubuntu-latest
+    needs: workspace-lint
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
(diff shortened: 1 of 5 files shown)


stevenobiajulu:
Closing as superseded by the fixes and expectation retirement merged in #674/#681 after #671. After resolving against current main, the SHA-256-pinned 14-cell corpus matrix passes on Darwin/Node 26 with no runtime-specific expectations, so retaining this patch would reintroduce stale pins rather than fix CI. PR #671 remains the merged implementation for #658.

Issue #687, opened 2026-07-27
docx-mcp: no formatting-convention check for inserted runs — a new defined term can land off-convention and pass every gate


stevenobiajulu:
## Summary

When an edit introduces a new construct that the document already formats by convention — an inline defined term, a proviso, a cross-reference — nothing compares the inserted run's properties against how that construct is formatted everywhere else. The edit is well-formed, round-trips, and renders. It is simply inconsistent, and no gate can see that.

Depends on #686 (there is currently nowhere for such a warning to be returned).

## Concrete failure class

A replacement introduced an inline defined term of the form `(each, a "Term")` with plain run properties. Every other parenthetical definition in that document — 44 of 46 — is bold italic, including the one in the sentence being replaced and the two in adjacent paragraphs. The result opens in Word, extracts byte-identical text, and passes `validate_document`; it is just visibly wrong to a reader.

The same shape recurs elsewhere: a proviso emitted without the underline on `provided`/`however` that the surrounding agreement uses 19 times; a heading emitted roman where its siblings are italic.

## Why existing checks cannot catch it

- `validate_document.ts` covers well-formedness only: `ORPHANED_BOOKMARK_START/END`, `MALFORMED_TRACKED_CHANGE`, `EMPTY_TRACKED_CHANGE`, `UNMATCHED_FIELD_BEGIN/END`.
- `validate_ai_revisions.ts` checks revision-markup legality.
- Neither compares an inserted run against the document's own distribution of comparable constructs.

## Ask

After a replacement introduces a new run, sample the document's existing formatting for the matching construct class and warn when the inserted run departs from the dominant pattern. Minimum viable version, in rough order of value:

1. **Inline defined terms** — runs whose text is enclosed in typographic quotes inside a parenthetical. Compute the modal `(bold, italic, underline)` across all such runs; warn on divergence.
2. **Proviso keywords** — `provided` / `however` immediately following `; `.

Structural only: compare resolved run properties, not text. This should reuse the effective-formatting resolution proposed in #684 (`extractEffectiveRunFormatting`), since a run inheriting bold italic from a character style must not be reported as divergent.

## Acceptance

- An edit introducing a parenthetical defined term whose formatting differs from the document's modal pattern returns a warning naming the construct and the observed-vs-dominant properties.
- No warning when the document has no established convention (fewer than N comparable instances, or no clear mode).
- Warning only — never blocks an edit.

stevenobiajulu:
## Cost of the convention check, measured

[...]

PR #691, opened 2026-07-27
feat(docx-core): make the effective run-formatting resolver public, fix its per-property chain resolution, and resolve D1 through styles.xml

### LINKED ISSUE #684: test(docx-compare): formatting-loss D1 compares declared formatting, not formatting resolved through styles.xml

stevenobiajulu:
## Summary

Two things ship together here, because the second was found while doing the first:

1. **Declare docx-core's effective run-formatting resolver public and rewire the D1 formatting-loss detector onto it** (issue #684). Accuracy note on the issue's premise: the resolver was *already reachable* from the package entry point via the star-export chain (`index.ts` → `primitives/index.ts` → `styles.ts`) — #684's "not currently part of docx-core's public entry point" overstates it. What this PR changes is intent, not reachability: the names are now explicitly exported and documented as deliberate public surface with their consumers named, instead of being incidental barrel spillage nothing was allowed to rely on. The detector previously compared *declared* formatting — direct `w:rPr` plus the `w:rStyle` reference — so emphasis arriving through paragraph-style inheritance was invisible, editing a style's *definition* while keeping the reference was invisible, and swapping a style for equivalent direct properties was flagged as a loss a reader cannot see.
2. **Fix a correctness bug in the resolver itself** (folded in from #700, which is closed in favor of this PR): `extractEffectiveRunFormatting` picked *one* `rPr` container per style chain — the first chain member that had any `rPr` at all — and read every property off that container. A derived style specifying only color therefore **masked an ancestor's bold**: `Derived(color) basedOn Base(bold)` resolved to not-bold. The nearby `extractStyleRunFormatting` already resolved per property, so the same file held two resolvers that disagreed. Making the buggy one public surface and building a loss detector on it would have shipped the miss to every consumer.

## What changed

- **`packages/docx-core/src/index.ts`** — explicit named exports: `parseStylesXml`, `extractEffectiveRunFormatting`, and the types their signatures reference (`StylesModel`, `StyleDef`, `RunFormatting`). The names already reached the entry point via star exports; restating them records that they are public on purpose, with both known consumers named (this script, and #687's formatting-convention detector).
- **`packages/docx-core/src/primitives/styles.ts`** —
  - *The precedence fix*: the four tiers (direct `rPr` → `w:rStyle` `basedOn` chain → paragraph mark `rPr` → paragraph style `basedOn` chain) are flattened into one ordered list of `rPr` containers and **each property resolves independently down it**. A chain member that specifies only color no longer masks an ancestor's bold. This aligns `extractEffectiveRunFormatting` with `extractStyleRunFormatting`, removing the disagreement.
  - *Toggle handling — pinned and bounded*: `<w:b w:val="0"/>` is a specification, not an absence — `parseBoolProp` returns non-null `false` for `0`/`false`/`off`, so `firstNonNull` stops at the nearer tier and an explicit off still defeats a more distant on. Pinned by tests across four shapes (off inside a `basedOn` chain, direct-over-paragraph-style, character-style-over-paragraph-style, nearer-on-over-ancestor-off). **Stated limitation, not a conformance claim**: this is a nearest-declaration cascade, deliberately simpler than full OOXML toggle-property evaluation, in which a style-level true XORs against accumulated state and a style-level false leaves it unchanged — repeated toggle declarations across style levels can resolve differently here than in Word. The pre-fix resolver applied the same nearest-value rule, **but per-property flattening widens where the simplification is visible** (round-2 review finding, confirmed by execution): the container-level race could skip an ancestor's toggle declaration behind a nearer member's unrelated `rPr`, and on some of those shapes it happened to agree with Word where the cascade does not — e.g. paragraph-style bold + a character-chain ancestor's explicit off XORs to bold in Word, resolves off here. That exact shape is pinned as a characterization test (`documented divergence: ...` in `check_docx_formatting_loss.test.mjs`), and the limit is documented in the resolver JSDoc, the test comments, and `docs/comparison-failure-diagnosis.md`. Implementing true toggle evaluation is deliberately out of scope here.
  - `parseBoolProp` now also reads the transitional ST_OnOff spelling `off` as false (the old detector handled it; the library did not).
  - JSDoc documents the resolution order and what the resolver does NOT reach: `w:docDefaults`, table-style rPr, numbering-level rPr, theme font references (`w:asciiTheme`, which would need `theme1.xml`).
- **`scripts/check_docx_formatting_loss.mjs`** — D1's per-character tuple is now the full resolved formatting (bold, italic, underline, highlight, font, size, color) from the public resolver, fed by `word/styles.xml` from each package. Color hex compares case-insensitively (`ff0000` = `FF0000`). A present styles part whose root is not `w:styles` is rejected (an empty model would silently blind D1 while reading as a clean pass); a *missing* part degrades to direct-only with a stderr note. The docx-core import is dynamic so an unbuilt tree fails with build instructions instead of a bare `ERR_MODULE_NOT_FOUND`. The self-test proves the new claims: a style-definition de-bolding fires, a style-vs-direct representation swap does not.
- **`docs/comparison-failure-diagnosis.md`** — the "D1 compares declared formatting" limitation this PR removes is rewritten to state what the resolver reaches and still does not; the build-order dependency is documented next to the usage block.
- **Tests** — script suite grows from 27 to 37: style-definition edits, paragraph-style inheritance, `basedOn` chains, the per-property masking fixture (executed red on the old resolver — see PR comments), toggle precedence, representation swaps staying silent, color-case equivalence, wrong-root styles part rejection, transitional `off`, the underline on/off trade, and the public import path. `generation-styles-formatting.test.ts` imports the resolver from `../index.js` so the surface breaks loudly if it falls off the barrel.

## Behaviour change, characterized on real documents

Against the real 863-paragraph ILPA model LPA (`tests/test_documents/redline/...v1.docx`):

| Scenario | Old detector | New detector |
|---|---|---|
| Remove `<w:b/>` from the `BoldCenter24pt` style definition; `document.xml` byte-identical | exit 0, **0 findings** (blind) | exit 1, **flags the exact 3 paragraphs** using the style |
| Swap one `w:rStyle="Hyperlink"` for equivalent direct `color`+`u` | exit 1, **false positive** | exit 0, silent |
| Unmodified NVCA `source.docx` vs `filled.docx` pairs | baseline | **output identical line-for-line** — no new false positives |

Stated limitations (deliberate, documented in header/docs/tests): underline is reduced to on/off, so a single→dotted underline restyle is no longer a finding; theme font references are not resolved, so swapping `w:asciiTheme` for the explicit font the theme maps to reads as a font change.

## Verification

Full pre-submit gate green locally before each push (`build`, `lint:workspaces`, `test:run`, `check:spec-coverage`, `check:conformance-citations`, `check:conformance-doc`), plus `npm run test:docx-formatting-loss` (37/37 + self-test) and the real-document runs above. Codex peer review round 1 verdict and adjudication in comments; the masking bug's red/green demonstration and a second Codex review of the combined change follow there too.

Closes #684
Refs #682, #687, #700

https://claude.ai/code/session_016rzEcGP1SqigLDb1QomJDg


stevenobiajulu:
## Executable evidence: the rPr-container masking bug (`eed038f`), red then green

Fixture: character style `Derived` (`basedOn` `Base`); `Derived`'s `rPr` carries **only** `<w:color w:val="FF0000"/>`, `Base`'s `rPr` carries `<w:b/>`. A run referencing `Derived` must resolve bold. Probe script: `.peer-review/probe-masking-684.mjs` (worktree-local, gitignored) — it calls both public resolvers on the same styles model.

**RED — pre-fix resolver** (`git checkout f2af67d -- packages/docx-core/src/primitives/styles.ts` + rebuild of `@usejunior/docx-core`):

```
=== RED phase (f2af67d resolver, pre-fix) ===
extractEffectiveRunFormatting -> bold:false colorHex:FF0000
extractStyleRunFormatting     -> bold:true colorHex:FF0000
RESOLVER: container-masking (bug present)
```

The pinned test (`node --test --test-name-pattern "basedOn resolution is per property" scripts/check_docx_formatting_loss.test.mjs`) on the same build:

```
tests 1
pass 0
fail 1
```

Two things demonstrated by actual execution, not description: the old `extractEffectiveRunFormatting` returns `bold:false` for a color-only derived character style inheriting bold, and `extractStyleRunFormatting` — same file, same styles model — already returned `bold:true`, i.e. the two resolvers genuinely disagreed pre-fix.

**GREEN — fixed resolver** (restored `HEAD` `b29f055`, rebuilt):

```
=== GREEN re-verify ===
extractEffectiveRunFormatting -> bold:true colorHex:FF0000
extractStyleRunFormatting     -> bold:true colorHex:FF0000
RESOLVER: per-property (fixed)
```

```
tests 1
pass 1
fail 0
```

Tree state after the demonstration verified clean: `git status --short` empty, `HEAD` = `b29f055`, branch `684-export-effective-run-formatting-20260727`.

Reachability in the real corpus was independently confirmed during the supervisor's review: 54 tracked `.docx` scanned, 6 documents / 25 style chains affected, e.g. `tests/test_documents/nvca-coi-regression/source.docx` style `Subtitle1` masking `[i, sz]`.

stevenobiajulu:

Ran a dynamic Codex review (`codex exec --sandbox workspace-write`) on the worktree; full transcript kept locally under `.peer-review/`. Codex executed the suite (33/33 at the time), the self-test, the NVCA real-document run, the public-import check, build/lint, and `check:conformance-citations`, and built its own counter-fixtures.

**Verdict: REQUEST-CHANGES** — three findings, adjudicated as follows:

1. **High — `basedOn` resolved per `w:rPr` container, not per property — ACCEPTED, fixed.** The resolver picked the first chain member that had *any* `rPr` and read every property from it, so `Derived(color) basedOn Base(bold)` masked the bold (Codex proved it by execution: projection showed `bold:false, color:FF0000`). Real pre-existing docx-core bug, exposed by making the resolver public; the neighboring `extractStyleRunFormatting` already resolved per-property. Fixed in `extractEffectiveRunFormatting` — properties now resolve independently across the ordered rPr sources — and pinned with the reviewer's exact fixture (projection shows bold AND color; de-bolding the ancestor is a D1 finding).

2. **Medium — color-casing and theme-font false positives — ACCEPTED for color, REBUTTED-AS-SCOPED for fonts.** `ff0000` vs `FF0000` is the same ink; the D1 key now compares hex case-insensitively (test added). Theme fonts (`w:asciiTheme` swapped for the explicit font the theme maps to) are NOT resolved — that requires `theme1.xml`, which the resolver deliberately does not reach. Rather than drop font from the tuple (which would hide real font loss), the corner is now a stated limitation in the script header, the resolver JSDoc, and the docs page. If it bites in practice, theme resolution belongs in docx-core, not in a script-side second implementation.

3. **Medium — present-but-wrong-root `styles.xml` silently reads as an empty model — ACCEPTED, fixed.** `<foo/>` as a styles part degraded to direct-only resolution while reading as a clean pass. The part is now validated exactly like `word/document.xml` (namespace + `w:styles` root) and a wrong root exits 2 (test added).

Codex also explicitly verified: the export surface is coherent and sufficient for #687 with nothing missing; removing an *undefined* rStyle correctly stays silent; missing-styles degradation notes fire per side; CI builds before the detector test runs.

Fixes are in the follow-up commit; full pre-submit gate rerun green before push. Script suite now 36/36.

stevenobiajulu:
## Codex peer review, round 2 (combined PR) — verdict and adjudication

Second dynamic Codex review of the full five-commit state (`f2af67d`…`b29f055`), explicitly asked to keep three rules separate: per-property inheritance for ordinary properties, direct-formatting boolean precedence, and style-level toggle behavior. Transcript kept locally under `.peer-review/`.

**Verdict: REQUEST-CHANGES** — one substantive finding, now addressed in `5001b6c`.

**Finding (Medium, ACCEPTED): per-property flattening *widens* the toggle-cascade divergence, and this PR's wording denied that.** The reviewer executed a fixture against builds of both the merge-base and HEAD: paragraph style bold on; character-chain ancestor `Base` with explicit `<w:b w:val="0"/>`; nearer member `Derived` carrying only color. Old resolver: container race skips `Base`, falls through to the paragraph style → `bold:true` — which happens to match full OOXML toggle evaluation (XOR across levels). New resolver: `Base`'s off is the first toggle declaration in the flattened order → `bold:false`. So "stated limit, not a regression" (resolver JSDoc) and "the simplification predates this change" (test comment) were inaccurate: the nearest-value *rule* predates the change, but its visible surface expanded.

Resolution, per the maintainer's stated preference for documenting over scope expansion:
- Both flagged wordings now state the expansion explicitly, with the concrete shape.
- The reviewer's exact fixture is pinned as a characterization test (`documented divergence: a nearer style-level toggle off defeats a paragraph-style on, where Word would XOR to on`) asserting the current simplified answer, so any future change to toggle handling is a visible decision rather than drift. Script suite is now 38/38.
- The PR body's toggle paragraph is corrected the same way. Implementing true §17.7.3-style toggle evaluation remains deliberately out of scope for this PR.

**Everything else was confirmed by execution**, with the three rules kept distinct as requested:
1. Ordinary (non-toggle) per-property inheritance: correct in all executed cases; cross-tier order intact (direct → rStyle chain → paragraph mark → paragraph-style chain).
2. Direct-formatting boolean precedence: direct `<w:b w:val="0"/>` defeats all style contributions (executed).
3. Toggle cascade: documented in all three locations; tests pin without overclaiming (post-fix).

Also verified by the reviewer: the five-name public surface present in `dist/index.d.ts`; D1 color case-insensitivity; wrong-root `styles.xml` → exit 2; missing `styles.xml` → stderr note + continue; detector suite green; full docx-core suite green outside the sandbox (1291 passed, 2 expected-fail, 15 skipped — the in-sandbox failures were tsx IPC `EPERM` and LibreOffice probes, environmental); `check:conformance-citations` OK.

stevenobiajulu:
## ✅ Final dependent-chain smoke passed

PR #691 merged as `804c30e`; its follow-ups #749 (full toggle evaluation) and #750 (theme font/color resolution) are now also merged. Final combined `main` is `6b79faa`.

- ✅ clean build, lint, and all workspace suites
- ✅ real IPC and LibreOffice probes
- ✅ 41/41 D1 detector tests plus self-test
- ✅ real 863-paragraph ILPA unchanged comparison clean at 100% coverage
- ✅ changing only the real ILPA package’s `minorHAnsi` theme font produced the expected 23 D1 findings
- ✅ spec and ECMA conformance gates
- ✅ final production deployment successful; live deployment follows to HTTP 200

Log: `/tmp/automerge-smoke-750-20260729T1736Z.log`

**Cleanup**
- ✅ repository auto-deleted the remote branch
- ✅ removed the clean `issue-684` worktree
- ✅ deleted the local branch

PR #700, opened 2026-07-28
fix(docx-core): resolve run formatting per property down the style chain

stevenobiajulu:
Rescued from an uncommitted worktree — opened as a draft so the work is durable and reviewable, not because it is ready to merge unexamined.

## Relationship to #691 — read this first

This branch is **stacked on #691**: it contains that PR's commit `f2af67d` plus one new commit `eed038f`. The diff against `main` therefore shows both.

The new commit **fixes a correctness bug in the code #691 is about to merge.** `extractEffectiveRunFormatting` resolved one `rPr` container per style chain — the first chain member possessing any `rPr` — and then read every property off it. A style specifying only a color masked an ancestor's bold, because a nearer chain member had already won a container-level race that should never have been container-level.

**#691 is currently green with zero reviews. Merging it as-is ships that bug.** The clean resolution is to fold `eed038f` into #691 and close this draft. This branch exists so the fix is not lost while that call is made.

## What the fix does

Flattens the four precedence tiers — direct `rPr`, `rStyle` chain, paragraph-mark `rPr`, paragraph-style chain — into one ordered list of containers, and resolves each property independently across it. `extractStyleRunFormatting` in the same file already worked this way, so this also removes a disagreement between two resolvers that are supposed to agree.

`scripts/check_docx_formatting_loss.mjs` consumes this resolver for D1, so the masking bug was a false-negative source in the formatting-loss detector, not only a library-level wrong answer.

Also documents two limits found while testing: theme font references (`w:asciiTheme`) are unresolved, so swapping a theme reference for the font it maps to reads as a font change without `theme1.xml`; and color hex compares case-insensitively.

## What has NOT been verified

I committed and pushed this without running the pre-submit gate or reviewing the accompanying test additions in `scripts/check_docx_formatting_loss.test.mjs`. CI on this PR is the first signal. Treat every claim above as the commit author's, not as verified.

Ref: #684


stevenobiajulu:
Folded into #691, which now carries this branch's commit unchanged: `eed038f` ("fix(docx-core): resolve run formatting per property down the style chain") is on `684-export-effective-run-formatting-20260727` as a distinct, unsquashed commit, followed by `ccde3c3` (toggle-precedence tests) and `1e6e917` (the toggle-cascade limitation stated in JSDoc/tests/docs).

Verified before closing: `gh pr view 691 --json commits` lists `eed038f` on the PR head. The executable old-vs-new masking probe (old resolver `bold:false`, fixed resolver `bold:true` for a color-only derived character style inheriting bold) is posted as review evidence on #691.

This draft existed so the fix could not be lost while the fold decision was made; that purpose is served. Closing, deleting the branch.

PR #692, opened 2026-07-28
feat(docx-core): add tracked section page numbering

stevenobiajulu:
## Summary
- add deterministic DOCX section discovery through the core API and the internal get_sections tool
- add a bounded format_section mutation for w:pgNumType/@w:start with native w:sectPrChange tracking
- preserve section topology, page setup, numbering format/chapter settings, columns, and header/footer references
- add OpenSpec, ECMA-376 citations, generated docs, bundle metadata, and accept/reject regression coverage

## Scope boundary
This is the first #654 slice. It does not create or split sections, edit break type/page size/margins, mutate header/footer parts, or change page-number format.

## Validation
- exact mandatory pre-submit sequence passed: build, workspace lint, full workspace tests, strict spec coverage, conformance citations, and conformance-doc freshness
- OpenSpec strict validation passed
- tool-doc freshness and emitted-schema self-test passed
- real mutual-NDA DOCX smoke passed for tracked, clean, accepted, and rejected outputs; all packages were ZIP-valid, all four document.xml parts passed the Transitional WML schema, and LibreOffice rendered all variants without visual drift

Ref: #654

PR #701, opened 2026-07-28
feat(docx-mcp): plumb validator warnings through the mutation success path

### LINKED ISSUE #686: docx-mcp: replace_text drops validator warnings on success and has no warnings field
## Summary

`replace_text` returns no `warnings` field, and the only validator on its path structurally discards warnings whenever the edit succeeds. An edit can therefore be accepted, round-trip cleanly, and render correctly while a validator had something to say about it that nobody ever sees.

## Evidence

`preflightAiRevisionMutation` returns early on success, so `validation.warnings` is unreachable:

```ts
const validation = await previewDoc.validateAiRevisions(session.aiAuthor, touched);
if (validation.errors.length === 0) return null;
```

`packages/docx-mcp/src/tools/ai_revision_guard.ts:120` (the `warnings` array is only read at `:143`, inside the failure return). A `null` return means the edit proceeds and every warning is dropped — not returned, not logged.

`packages/docx-mcp/src/tools/replace_text.ts:307-315` returns `{ success, file_path, edit_count, target_paragraph_id, replacements_made, before_text, after_text }` — no `warnings`.

This is inconsistent rather than a house style: `batch_edit.ts:51,588,598`, `read_file.ts:589,618` and `format_layout.ts:357-422` all carry a `warnings` channel. `clear_formatting.ts:143-148` and `add_comment.ts:67-73,151-157` share `replace_text`'s pattern.

## Why it matters

Callers are agents. An agent that receives `success: true` with no other signal has no reason to re-inspect its own edit, so a warning that exists inside the engine cannot influence the next step. This is the plumbing that any future edit-time check would need in order to be useful at all.

## Ask

Plumb `warnings` through the success path for `replace_text`, `clear_formatting`, and `add_comment`: have `preflightAiRevisionMutation` return warnings on success rather than `null`, and add a `warnings?: string[]` field to the success responses, matching the shape already used by `batch_edit`.

## Acceptance

- A successful `replace_text` whose validation produced a warning returns that warning in the response.
- Existing callers that ignore the field are unaffected (additive, optional).

stevenobiajulu:
## Problem

`preflightAiRevisionMutation` returned `Promise<ToolResponse | null>`, where `null` meant "proceed". Because it returned bare `null` on every success path, `validation.warnings` — and the `demoted` pre-existing diagnostics on the introduced/demoted split — were structurally unreachable whenever an edit succeeded. A validator could have something to say about an accepted edit and nobody would ever see it: not returned, not logged. Callers here are agents; an agent that receives `success: true` with no other signal has no reason to re-inspect its own edit, so an engine-side warning could never influence the next step.

`replace_text`, `clear_formatting`, and `add_comment` also had no `warnings` field in their success responses at all — inconsistent with the warnings channels `batch_edit`, `read_file`, and `format_layout` already expose (`format_layout` and `read_file` carry plain `string[]`; `batch_edit` wraps the same human-readable strings with step metadata as `{ step_id, warning }` entries).

## Change

- `preflightAiRevisionMutation` now returns `Promise<AiRevisionPreflightResult>` where `AiRevisionPreflightResult = { blocked: ToolResponse | null; warnings: string[] }`. `blocked` non-null is the old failure contract; `warnings` carries formatted non-blocking diagnostics on the proceed path. Both success-ish return points carry warnings — the post-split one merges `demoted` diagnostics exactly as the failure path always has, so pre-existing anomalies are demoted-but-surfaced instead of demoted-and-silenced.
- New `formatAiRevisionWarning()` renders an `AiRevisionDiagnostic` as the plain-string shape the `format_layout`/`read_file` `warnings` channels use: `CODE: message (part, element, id=N, author=X)`.
- `replace_text`, `clear_formatting`, and `add_comment` (both reply and root modes) surface `warnings?: string[]` on success — present only when non-empty, so callers that ignore it are unaffected.
- All other call sites (`batch_edit`, `format_layout`, `format_numbering`, `format_section`, `insert_section_break`, `insert_paragraph`, `add_footnote`, `update_footnote`, `delete_footnote`, `delete_comment`) are updated to the explicit `.blocked` check but do not surface warnings — that is outside #686's scope. `format_section` (#692) and `insert_section_break` (#711) landed on main while this branch was in flight; each merge commit adapts them shape-only (stop-blocking semantics preserved, no warnings surfaced), deliberately keeping the warnings-surfacing scope to the three tools #686 names.
- No change to which conditions warn, and no warning blocks an edit. Plumbing only.

## Why the `{ blocked, warnings }` shape

The old `if (revisionPreflight) return revisionPreflight;` truthiness idiom is exactly the pattern that made warnings unreachable. Returning an always-object with an explicit `blocked` field makes every call site's "did preflight block me?" check impossible to conflate with "did preflight say anything?", and the compiler forces every stale call site to be revisited (a leftover truthiness check no longer typechecks as a `ToolResponse` return).

## Tests

`ai_revision_validation.test.ts`: successful `replace_text` returns a foreign-revision warning; clean success omits the `warnings` key entirely; `clear_formatting` and `add_comment` surface warnings; the pre-existing-anomaly (demoted) path now asserts the demoted diagnostic appears in the success response; unit coverage for `formatAiRevisionWarning`.

Full pre-submit gate (build, lint:workspaces, test:run, spec-coverage, conformance-citations, conformance-doc) passes locally at `ae75779`.

The branch includes two merges of current `main`; each merge commit adapts a concurrently-landed preflight call site (`format_section.ts` from #692, then `insert_section_break.ts` from #711) to the reshaped return type — the compile errors CI reported between merges are exactly the fail-closed behavior the `{ blocked, warnings }` shape was chosen for: a call site written against the old `ToolResponse | null` contract fails the build instead of silently returning the wrapper object.

#687 (convention detector) builds on the warnings channel shape landed here.

Closes #686

https://claude.ai/code/session_016rzEcGP1SqigLDb1QomJDg


stevenobiajulu:

**Verdict: APPROVE-WITH-NITS** (Codex CLI, dynamic review with execution: it ran the test file — exit 0, 15/15 at the time — built `@usejunior/docx-mcp` — exit 0 — swept all `preflightAiRevisionMutation` call sites, and exercised a one-off reply-mode fixture).

### Findings and how each was handled

1. **Nit: the `add_comment` reply-mode success branch was implemented but untested.** Accepted — real gap. The reply branch builds its success response separately from the root branch, so a regression in its warnings plumbing would have been invisible. Fixed in `6d17305`: new test adds a root comment, replies to it in a document carrying a foreign-revision warning, and asserts the reply succeeds with `mode: 'reply'` and surfaces the `REVISION_METADATA_MISSING` warning. Suite now 16/16.

2. **Nit: `formatAiRevisionWarning` JSDoc wrongly claimed `batch_edit` exposes plain-string warnings.** Accepted — factually wrong prose (the runtime shape landed here was already correct). `batch_edit` wraps its human-readable warning strings with step metadata as `{ step_id, warning }` entries; the plain `string[]` channels are `format_layout` and `read_file`. JSDoc corrected in `6d17305`, and the same imprecision was fixed in the PR body. Worth being exact because #687 builds a convention detector on this channel.

No REQUEST-CHANGES findings. The reviewer confirmed all 13 call sites use the new `.blocked` contract and that no truthiness check on the result object remains.

### codecov/patch on `ae75779`

The 63.33%-of-diff failure on the pre-nit head does not reproduce locally against the real diff: cross-referencing `git diff <merge-base> HEAD -- packages/docx-mcp/src` against fresh lcov shows **0 uncovered added lines** across all 14 touched files (30/30 instrumented lines hit). The docx-mcp package is also above the ratchet baseline on all four metrics (lines 86.04 vs 83.66 baseline, branches 76.23 vs 74.30). Expecting the recompute on `6d17305` to clear; holding automerge until it does.

Full pre-submit gate (build, lint:workspaces, test:run, spec-coverage, conformance-citations, conformance-doc) passed at `6d17305` with explicit exit code 0.

stevenobiajulu:
## Post-merge smoke — PASS

Merged as `e85428c` on `main`. Smoke run by the supervisor against a real corpus document
(`tests/test_documents/open-agreements/common-paper-mutual-nda.docx`), driving the real MCP tool
functions, from a tree whose `packages/docx-mcp/src` is byte-identical to merged `main`.

| Case | Expectation | Result |
|---|---|---|
| A | `replace_text` on the clean NDA → success, **no** `warnings` field | pass |
| B | `replace_text` on the NDA carrying a foreign tracked deletion with no `w:date` → success **with** `warnings[]` | pass |
| C | `clear_formatting` on the clean NDA → success, field optional | pass |
| D | `add_comment` on the clean NDA → success, field optional | pass |

Case B is the behaviour this issue was about — warnings that were previously dropped on the success
path now reach the caller, without blocking the edit:

```json
"success": true,
"replacements_made": 1,
"warnings": [
  "REVISION_METADATA_MISSING: <w:del> missing required w:date (word/document.xml, w:del, id=9001, author=Opposing Counsel)",
  "REVISION_DATE_INVALID: <w:del> has invalid w:date (word/document.xml, w:del, id=9001, author=Opposing Counsel)"
]
```

### Note on the final CI cycle

The `codecov/patch` failure at 64.51% was a real coverage gap, not a metric artifact: the
`{ blocked, warnings }` reshape rewrote the rejection check in every guarded tool, and eleven of
them had no test proving the tool still refuses to mutate when validation fails. `da6df9b` added a
blocked-path test per uncovered tool, taking diff-line branch coverage to 31/31 full. The branch was
then re-synced onto current `main` (which had moved through #691, #749, #750) and pushed as
`13a02f6`; full build and 966 `docx-mcp` tests green locally, all checks green including the Vercel
commit status.

### Open follow-up (not a blocker for this PR)

`batch_edit` still ignores `revisionPreflight.warnings`. Its response exposes a different shape —
`{ step_id, warning }[]` — so plumbing the preflight warnings through it is a separate decision
about which shape wins, not a mechanical extension of this change.

Issue #718, opened 2026-07-29
docx-compare: private competitive comparison corpus gate


stevenobiajulu:
Goal: a leak-resistant, hash-pinned acceptance gate over large Word-authored document pairs that fails when comparison quality regresses against a pinned licensed baseline, without committing confidential text, filenames, party names, metadata, links or proprietary terms.

Done so far: the private corpus harness and manifest (`packages/docx-compare/src/integration/private-comparison-corpus.corpus.test.ts`, #911, #930) and the required real-corpus gate on main (#931). Quality-contract item 8 (a passing compiled-verifier certificate) no longer applies: the Lean verifier was retired in #826 and #971, and its dependencies #714, #715, #745, #746, #747 and #760 were closed not planned.

Remaining:

- [ ] #688
- [ ] #754
- [x] #582 (closed in this triage: the rebuild path no longer exists)
- [x] #713 (closed in this triage as fixed)
- [ ] Add the remaining pairs to `private-comparison-corpus-manifest.json` (it lists one pair; the gate was designed for three).
- [ ] Fail in both directions: regressions fail, and an expected-failure characterization that starts passing must be deliberately removed.
- [ ] Quality contract, per pair: comparison succeeds on the supported production path.
- [ ] Quality contract, per pair: output opens without repair.
- [ ] Quality contract, per pair: accept/reject projections recover revised/original.
- [ ] Quality contract, per pair: no unchanged numeric or text span is marked as changed.
- [ ] Quality contract, per pair: TOC/field results have no false-positive noise.
- [ ] Quality contract, per pair: formatting, sections, tables, content controls, text boxes and side stories are preserved or truthfully certified.
- [ ] Quality contract, per pair: Safe-DOCX has no more false-positive or false-negative revisions than the pinned baseline.

Done: #716, #717, #720, #724, #648.

---
<details><summary>Original report (preserved from before the 2026-09-23 triage)</summary>

## Objective

Create a leak-resistant, structurally faithful acceptance corpus and differential gate for three large Word-authored document pairs. The gate must establish comparison quality against a pinned licensed baseline without committing confidential text, filenames, party names, addresses, metadata, links, or proprietary terms.

## Privacy requirements

- Generic package names only: `file-1-original.docx`, `file-1-revised.docx`, etc.
- Deterministic pseudonymization across each original/revised pair so equality and edit relationships survive.
[...]

PR #723, opened 2026-07-29
fix(docx-comparison): ignore volatile TOC page caches

### LINKED ISSUE #716: docx-compare: cached TOC/PAGEREF results create widespread false-positive revisions
## Summary

On a Word-authored agreement with a cached table-of-contents field, current in-place comparison marks most cached TOC entries and page numbers as deleted/reinserted even though the competing baseline leaves unchanged page references clean and marks only the substantive changed entry.

Aggregate observation from a confidential pair, reported without document text or identifiers:

- Safe-DOCX: 11 TOC paragraphs carry revisions; 27 insertion/deletion containers occur inside them.
- Baseline comparator: no spurious cached page-number revisions; the substantive amended-title edit remains visible.

The confidential file is not an acceptable public fixture. A structurally equivalent scrubbed fixture or minimal public fixture must reproduce the behavior before implementation.

## Likely boundary to define

TOCs combine nested `TOC`/`PAGEREF` fields, cached results, hyperlinks, bookmarks, tabs, leaders, and page numbers. The comparison contract must distinguish:

- changed field instructions or entry text, which are substantive;
- changed cached page numbers caused only by pagination/field refresh;
- genuinely added or removed TOC entries;
- formatting-only changes;
- user-authored text that happens to use a TOC style.

## Acceptance

- [ ] Unchanged cached `PAGEREF` results do not create phantom revisions.
- [ ] A genuinely added/removed/renamed TOC entry is still represented.
- [ ] Accept/reject projections recover the intended revised/original field structures.
- [ ] Field marker nesting and hyperlink/bookmark ownership remain valid.
- [ ] A synthetic control and a safely scrubbed Word-authored fixture both pass.
- [ ] A differential assertion establishes that Safe-DOCX produces no more false-positive TOC changes than the pinned baseline on the acceptance fixture.

## Related

- #645 / #659: nested field serialization.
- #646 / #665: field ownership across paragraph deletion.
- #675 / #677: phantom revisions from run identity.

stevenobiajulu:
## Summary
- suppress tracked changes caused solely by refreshed `PAGEREF` cached results in TOC-styled paragraphs
- require accepted/rejected structural fingerprints to match after masking only the cache before cleaning a revision
- share the cache-insensitive projection between runtime round-trip safety and corpus invariants
- add generic ECMA-376 regression coverage for cache-only refreshes plus substantive renamed/added TOC entries

## Confidential-oracle result
- TOC revision paragraphs: 11 -> 6 genuine changes
- TOC revision containers: 27 -> 12
- accept projection: all 12 revised semantic TOC paragraphs recovered
- reject projection: all 10 original semantic TOC paragraphs recovered
- unchanged currency delete/reinsert artifacts remain 0
- no confidential filenames, text, parties, or paths are included in this PR

## Validation
- `npm run build && npm run lint:workspaces && npm run test:run && npm run check:spec-coverage && npm run check:conformance-citations && npm run check:conformance-doc`
- confidential aggregate oracle benchmark

Fixes #716
Ref #718

### DIFF
--- a/packages/docx-compare/src/baselines/atomizer/pipeline-pageref-cache.test.ts
+++ b/packages/docx-compare/src/baselines/atomizer/pipeline-pageref-cache.test.ts
@@ -1,0 +1,182 @@
+/**
+ * Cached PAGEREF results are layout artifacts, not authored TOC edits.
+ *
+ * @conformance ECMA-376 edition 5, Part 1 § 17.16.18
+ * @conformance ECMA-376 edition 5, Part 1 § 17.16.5.45
+ * @see https://github.com/UseJunior/safe-docx/issues/716
+ */
+
+import { DocxArchive, parseXml } from '@usejunior/docx-core';
+import { describe, expect } from 'vitest';
+import {
+  buildDocxFromBodyXml,
+  completeField,
+  fldChar,
+  instrText,
+  resultText,
+} from '../../testing/ooxml-fixtures.js';
+import { testAllure, type AllureBddContext } from '../../testing/allure-test.js';
+import { compareDocumentsAtomizer } from './pipeline.js';
+import {
+  acceptAllChanges,
+  extractTextWithParagraphs,
+  rejectAllChanges,
+} from './trackChangesAcceptorAst.js';
+
+const test = testAllure
+  .epic('Document Comparison')
+  .withLabels({
+    feature: 'In-Place Reconstruction',
+    story: 'TOC PAGEREF Cached Results',
+    severity: 'critical',
+  })
+  .conformance(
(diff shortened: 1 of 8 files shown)


stevenobiajulu:
Post-merge smoke passed on merged `main` (`2cbca117`).

- built `@usejunior/docx-mcp` and the comparison package
- ran the built `safe-docx compare` CLI end-to-end on the public NVCA COI source/filled document pair in `inplace` mode
- output DOCX ZIP integrity passed
- accept-all semantic projection matched the filled document
- reject-all semantic projection matched the source document

No confidential corpus paths, filenames, or content were used in this comment.

PR #749, opened 2026-07-29
fix(docx-core): evaluate style toggle properties by parity

### LINKED ISSUE #737: docx-core: run-formatting resolver implements a nearest-declaration cascade, not OOXML toggle-property evaluation
## Summary

`extractEffectiveRunFormatting` (`packages/docx-core/src/primitives/styles.ts`) resolves each run property by taking the nearest declaration across four tiers: direct `rPr`, the `w:rStyle` chain, the paragraph-mark `rPr`, and the paragraph-style chain. That is a nearest-declaration cascade. It is not OOXML toggle-property evaluation, and for toggle properties the two rules disagree.

## What the spec requires

Toggle properties — `w:b`, `w:i`, `w:caps`, `w:smallCaps`, `w:strike`, `w:emboss`, `w:imprint`, `w:outline`, `w:shadow`, `w:vanish` — do not follow ordinary nearest-wins inheritance:

- At **style level**, a `true` toggles the currently-effective state rather than setting it, and a `false` leaves the current state unchanged.
- At **direct-formatting level**, the value is absolute.
- When the same toggle is declared at multiple style levels, the effective value depends on the parity of the declarations, not on the nearest one.

See ECMA-376 Part 1 §17.7.3 and the corresponding [MS-OI29500] implementation notes.

`parseBoolProp` treats `w:val="0" | "false" | "off"` as a plain absolute false at every tier, so none of the above is implemented.

## Evidence this is observable

PR #691 changed container-level resolution to per-property resolution, which was a correctness fix for ordinary properties (see the masking bug it closed). During review of that change, a fixture was constructed where the behaviours diverge:

- paragraph style declares bold
- a character-style ancestor declares an explicit off
- a nearer character-style member declares only a colour

The pre-#691 container race produced `bold:true`, matching Word. The per-property cascade produces `bold:false`. The fixture is pinned in the test suite as a documented divergence rather than as expected behaviour.

This is one confirmed shape. The parity rule for repeated cross-level declarations is untested and is expected to produce further divergence.

## Ask

Implement toggle-property evaluation for the toggle set, distinguishing style-level from direct-formatting declarations, and replace the pinned divergence fixture with a conformance assertion.

Scope notes:

- The resolver is consumed by `document_view`, `footnotes`, `document_view-headings`, and the D1 formatting-loss detector (`scripts/check_docx_formatting_loss.mjs`). A precedence change affects all of them, so this needs its own characterisation baseline before the change, not after.
- Ordinary (non-toggle) properties must keep per-property nearest-wins resolution. The two rules coexist; this is not a revert of #691.
- Word's observed behaviour governs where it diverges from a strict reading of the spec. A Word differential across repeated cross-level toggle declarations is the acceptance evidence, not a reading of the text alone.

## Acceptance

- A document with a toggle declared at two style levels resolves to the parity-correct value, verified against Word.
- Style-level `false` leaves the inherited state unchanged; direct-formatting `false` forces off.
- The divergence fixture added in #691 becomes a passing conformance test.
- The JSDoc limitation note added in #691 is removed rather than reworded.

Refs #684, #691.

stevenobiajulu:
## What changed

- evaluates all supported OOXML toggle properties (`b`, `i`, `caps`, `smallCaps`, `strike`, `emboss`, `imprint`, `outline`, `shadow`, and `vanish`) using style-level parity
- keeps direct formatting absolute
- extends the D1 formatting tuple to the complete supported toggle set
- converts PR #691's pinned Word differential from a documented divergence into a passing assertion
- adds ECMA-376 §17.7.3 conformance evidence and generated documentation

## Why

PR #691 correctly made ordinary properties resolve independently down style chains, but its nearest-declaration treatment of toggle properties made one Word-observed fixture resolve incorrectly. Style-level enabled declarations invert the accumulated state, disabled declarations are no-ops, and direct formatting sets the absolute result.

## Validation

- full build and workspace lint
- full test suite; sandbox-only IPC and LibreOffice failures rerun successfully with normal process access
- D1 detector suite and self-test
- spec coverage
- conformance citation and generated-document gates

Fixes #737
Depends on and is based on merged PR #691.


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `1f38334ebaaaa9f0cdbcc4338c09e14170850939`
Verified in final combined `main`: `6b79faa`

- ✅ clean build and workspace lint
- ✅ all workspace suites, including IPC and LibreOffice probes
- ✅ Word-differential toggle assertion and complete 10-toggle D1 tuple: 41/41 detector tests green in the final combined state
- ✅ real 863-paragraph ILPA self-comparison: zero findings at 100% coverage
- ✅ spec/conformance gates
- ✅ production deployment for final combined main is successful and returns HTTP 200

Central LLM gate verdict for this PR: **PASS** (8 pass, 0 warn, 0 error).

Log: `/tmp/automerge-smoke-750-20260729T1736Z.log`

**Cleanup**
- ✅ remote branch already deleted by repository auto-delete
- ✅ removed dedicated worktree `/private/tmp/safe-docx-737`
- ✅ deleted local feature branch

PR #776, opened 2026-07-30
feat(docx-mcp): add verified CLI comparisons

### LINKED ISSUE #775: safe-docx CLI does not expose Lean-verified comparison certificates
## Problem

`safe-docx compare` does not currently expose the optional compiled Lean verifier, so CLI users receive TypeScript reconstruction safety checks but no document-integrity certificate. The verifier also rejects harmless ZIP directory entries produced by valid tooling, and its public timeout documentation disagrees with the implementation.

## Scope

- Add an explicit CLI verified-comparison mode.
- Return and optionally write the JSON document-integrity certificate.
- Fail closed when verification does not pass.
- Accept harmless ZIP directory entries without weakening archive ambiguity checks.
- Make the verifier timeout consistently 10 seconds.
- Add a public NVCA-derived end-to-end performance gate requiring verified comparison within 10 seconds on the supported test environment.

## Privacy

Use only committed public/sanitized fixtures. Do not reference confidential corpus names, filenames, entities, addresses, or other PII.

stevenobiajulu:
## Summary

- add `safe-docx compare --verify` and `--certificate <path>`
- fail closed before publishing a redline when the Lean certificate does not pass
- use a 10-second default verifier timeout and exercise a public NVCA-derived document pair
- admit only inert ZIP directory placeholders while proving they cannot be selected as XML stories

## Evidence

- full mandatory pre-submit sequence passes
- full Lean build passes (6,628 jobs)
- axiom audit matches the reviewed production boundary: `Classical.choice`, `Quot.sound`, `propext`
- no `sorry` in Lean sources
- strict OpenSpec validation passes
- public verified CLI performance gate: 6.65 seconds under full-suite load (10-second ceiling)

## Notes

Ordinary CLI comparisons remain unverified unless explicitly requested. Certificate and redline files are published only after verification passes. No confidential corpus material or identifying metadata is included.

Fixes #775

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -37,6 +37,8 @@ Running `safe-docx` with no command starts the MCP server. The CLI can also invo
 ```bash
 safe-docx grep "governing law" contract.docx
 safe-docx compare original.docx revised.docx comparison.docx
+safe-docx compare original.docx revised.docx comparison.docx --verify \
+  --certificate comparison.certificate.json
 safe-docx export --file-path contract.docx --format markdown
 ```
 
--- a/docs/trust-and-conformance.md
+++ b/docs/trust-and-conformance.md
@@ -60,6 +60,19 @@ verification/lean/.lake/build/bin/leanDocxChecker
 
 Set `SAFE_DOCX_LEAN_XML_CHECKER` to use another executable. Normal package usage does not require Lean or Lake.
 
+The installed CLI exposes the same opt-in boundary:
+
+```bash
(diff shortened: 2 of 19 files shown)


stevenobiajulu:
Automerge stalled — CI failed: `allure-labels`.

Local reproduction: `packages/docx-mcp/src/cli/commands/compare.test.ts` contains OpenSpec traceability tests but does not declare the required deterministic `const TEST_FEATURE = ...`. Other checks were still running when this gate failed. Automerge remains armed while the focused metadata fix is prepared.

Issue #794, opened 2026-08-11
docx-compare: mc:Fallback subtrees are walked as real content, so one text box counts as two stories and its text is projected four times


stevenobiajulu:
## Summary

**Severity is wrong counts and wrong diagnostics, not wrong redline.** See "What I could NOT reproduce" below — I tried to produce a corrupted output document and could not.

Word stores a modern text box **twice** inside a single `mc:AlternateContent`: an `mc:Choice Requires="wps"` holding the DrawingML spelling and an `mc:Fallback` holding the VML spelling. Exactly one of them renders. Both contain a full `w:txbxContent` with real `w:p`/`w:r`/`w:t`.

Nothing in the repository skips `mc:Fallback`. The strings `mc:Choice` and `mc:Fallback` do not appear in any `.ts`, `.md` or `.lean` file. The consequence is that every unfiltered `getElementsByTagNameNS` scan counts one visual text box as two, and projects its text more than once.

## Where

- `packages/docx-compare/src/baselines/atomizer/textBoxRevisionSafety.ts:99-104` — `textBoxes()` is an unfiltered `root.getElementsByTagNameNS(OOXML.W_NS, 'txbxContent')`. The same scan appears at `:149`, `:158`, `:227`, `:245`, `:320`, `:328`, `:911`, `:914`, `:1060`, `:1142`, `:1156`.
- Same file `:1058-1062` — the splice-back step re-resolves its target by the same doubled ordinal.
- Same file `:53` — the user-facing locator is built from that ordinal:
  ``const locator = `${partPath ?? 'word/document.xml'}#w:txbxContent[${index}]` ``.
- `packages/docx-compare/src/fieldComparisonSemantics.ts:46-48` — `getElementsByTagNameNS(OOXML.W_NS, 'p')` over the whole document, so both branches' paragraphs enter the round-trip projection.

The atomizer is not affected: `mc:AlternateContent` is in `LEAF_NODE_TAGS` (`packages/docx-compare/src/atomizer.ts:452`), so `atomizeTreeInternal` stops there.

## Reproduction

Fully synthetic. One text box, authored the way Word writes one. Save as `repro-mc.mjs` in the repo root and run with `node` after `npm run build`.

```js
// Fully synthetic. Run from the safe-docx repo root after `npm run build`:
//   node repro-mc.mjs
import { createRequire } from 'node:module';
const require = createRequire(process.cwd() + '/package.json');
const { extractRoundTripComparisonText } = require(
[...]

PR #800, opened 2026-08-11
fix(docx-compare): enforce the DrawingML text-box exclusion for standalone boxes

### LINKED ISSUE #795: docx-compare: the documented "does not cover DrawingML text boxes" exclusion is not enforced — the guard compares undefined !== undefined
## Summary

**This is a conformance-claim defect, not a live output defect.** In every case I ran, the produced redline was correct. See "Honest note on severity" below.

`spec-compliance/CONFORMANCE.md:155-157` states that the `w:txbxContent` comparison claim "does not cover DrawingML text boxes, ancillary-part text boxes, inserted/deleted text-box topology, or changes to the containing VML scaffold or relationship closure." The code does not enforce the DrawingML half of that.

`scaffoldFingerprint()` (`packages/docx-compare/src/baselines/atomizer/textBoxRevisionSafety.ts:144-154`) calls `nearestShape()` (`:132-142`), which walks ancestors looking **only for a VML `v:shape`**:

```ts
function nearestShape(textBox: Element): Element | undefined {
  let node: Node | null = textBox.parentNode;
  while (node?.nodeType === 1) {
    const element = node as Element;
    if (element.namespaceURI === VML_NS && element.localName === 'shape') {
      return element;
    }
    node = node.parentNode;
  }
  return undefined;
}
```

A DrawingML text box has no `v:shape` ancestor, so `scaffoldFingerprint` returns `undefined` on both sides. The guard at `:948-951` is then:

```ts
(scaffoldFingerprint(originalTextBox) !==
scaffoldFingerprint(revisedTextBox)
  ? 'the containing VML shape scaffold changed or could not be paired'
  : undefined)
```

`undefined !== undefined` is `false`, so the guard does not fire and the box is admitted to the supported subset.

Contrast the sibling guard three lines above (`:943-947`), which *does* fail closed on `undefined` by testing `=== undefined ||` explicitly:

```ts
(originalRelationshipClosure === undefined ||
revisedRelationshipClosure === undefined ||
originalRelationshipClosure !== revisedRelationshipClosure
  ? 'the text-box relationship closure changed or could not be resolved'
  : undefined)
```

The scaffold guard was written to the same shape but omits that test. **The ancillary-part copy at `:767-769` has the same hole**, spelled the other way round — `scaffoldFingerprint(a) === scaffoldFingerprint(b) ? undefined : '…'` — and again treats two `undefined`s as a successful pairing.

## Reproduction

Fully synthetic. The document's only text box is DrawingML (`wps:txbx/w:txbxContent`), with no `v:shape` anywhere. Save as `repro-drawingml.mjs` in the repo root and run with `node` after `npm run build`.

```js
// Fully synthetic. Run from the safe-docx repo root after `npm run build`:
//   node repro-drawingml.mjs
//
// The document's ONLY text box is DrawingML (wps:txbx/w:txbxContent), with no
// VML v:shape anywhere — the exact case CONFORMANCE.md:155-157 excludes.
const JSZip = require('jszip');
const compare = require(
  './packages/docx-compare/dist/baselines/atomizer/pipeline.js');

const W = 'http://schemas.openxmlformats.org/wordprocessingml/2006/main';
const NS =
  ` xmlns:wp="http://schemas.openxmlformats.org/drawingml/2006/wordprocessingDrawing"` +
  ` xmlns:a="http://schemas.openxmlformats.org/drawingml/2006/main"` +
  ` xmlns:wps="http://schemas.microsoft.com/office/word/2010/wordprocessingShape"`;

const doc = (body) =>
  `<?xml version="1.0" encoding="UTF-8" standalone="yes"?>` +
  `<w:document xmlns:w="${W}"` +
  ` xmlns:w14="http://schemas.microsoft.com/office/word/2010/wordml"` +
  ` xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"` +
  NS + ` mc:Ignorable="w14"><w:body>${body}<w:sectPr/></w:body></w:document>`;

// A pure DrawingML text box. No mc:AlternateContent, no v:shape, no VML at all.
const drawingMlBox = (inner) =>
  `<w:p><w:r><w:drawing><wp:inline><a:graphic><a:graphicData>` +
  `<wps:wsp><wps:txbx><w:txbxContent><w:p><w:r><w:t>${inner}</w:t></w:r></w:p>` +
  `</w:txbxContent></wps:txbx></wps:wsp>` +
  `</a:graphicData></a:graphic></wp:inline></w:drawing></w:r></w:p>`;

const zipDocx = async (documentXml) => {
  const zip = new JSZip();
  zip.file('[Content_Types].xml',
    `<?xml version="1.0" encoding="UTF-8" standalone="yes"?>` +
    `<Types xmlns="http://schemas.openxmlformats.org/package/2006/content-types">` +
    `<Default Extension="rels" ContentType="application/vnd.openxmlformats-package.relationships+xml"/>` +
    `<Default Extension="xml" ContentType="application/xml"/>` +
    `<Override PartName="/word/document.xml" ContentType="application/vnd.openxmlformats-officedocument.wordprocessingml.document.main+xml"/>` +
    `</Types>`, { createFolders: false });
  zip.file('_rels/.rels',
    `<?xml version="1.0" encoding="UTF-8" standalone="yes"?>` +
    `<Relationships xmlns="http://schemas.openxmlformats.org/package/2006/relationships">` +
    `<Relationship Id="rId1" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/officeDocument" Target="word/document.xml"/>` +
    `</Relationships>`, { createFolders: false });
  zip.file('word/document.xml', documentXml, { createFolders: false });
  zip.file('word/_rels/document.xml.rels',
    `<?xml version="1.0" encoding="UTF-8" standalone="yes"?>` +
    `<Relationships xmlns="http://schemas.openxmlformats.org/package/2006/relationships"/>`,
    { createFolders: false });
  return zip.generateAsync({ type: 'nodebuffer' });
};

const fn = compare.compareDocumentsAtomizer ?? compare.default?.compareDocumentsAtomizer;

const before = doc(drawingMlBox('Charlie'));
const after = doc(drawingMlBox('Delta'));

console.log('  v:shape elements in the input :', (before.match(/<v:shape/g) || []).length);
console.log('  w:txbxContent elements        :', (before.match(/<w:txbxContent/g) || []).length);
console.log('');

try {
  const res = await fn(await zipDocx(before), await zipDocx(after),
    { reconstructionMode: 'inplace' });
  const d = res?.document;
  const xml = typeof d === 'string'
    ? d
    : (await (await JSZip.loadAsync(d)).file('word/document.xml').async('string'));
  const s = xml.slice(xml.indexOf('<w:txbxContent'), xml.indexOf('</w:txbxContent>'));
  const c = (re) => (s.match(re) || []).length;
  console.log('  compareDocumentsAtomizer     : returned a result (did NOT fail closed)');
  console.log(`  reconstructionModeRequested  : ${res?.reconstructionModeRequested}`);
  console.log(`  reconstructionModeUsed       : ${res?.reconstructionModeUsed}`);
  console.log(`  fallbackReason               : ${res?.fallbackReason ?? 'none'}`);
  console.log(`  unrepresentedChanges         : ${JSON.stringify(res?.unrepresentedChanges ?? null)}`);
  console.log(`  inside the DrawingML box     : w:ins=${c(/<w:ins[ >]/g)} w:del=${c(/<w:del[ >]/g)} Charlie=${c(/Charlie/g)} Delta=${c(/Delta/g)}`);
} catch (err) {
  console.log(`  threw: ${err?.constructor?.name}: ${err?.message}`);
}
```

## Actual output

Run at `8017621` (`[DEBUG]` lines from the atomizer elided):

```
  v:shape elements in the input : 0
  w:txbxContent elements        : 1

  compareDocumentsAtomizer     : returned a result (did NOT fail closed)
  reconstructionModeRequested  : inplace
  reconstructionModeUsed       : inplace
  fallbackReason               : none
  unrepresentedChanges         : null
  inside the DrawingML box     : w:ins=2 w:del=2 Charlie=1 Delta=1
```

A result is returned, in `inplace` mode, with no fallback, no `unrepresentedChanges`, and no diagnostic — for the exact document class the conformance registry says is out of scope.

## Expected

One of:

- **Fail closed.** `scaffoldFingerprint` returning `undefined` on either side should raise `UnsupportedTextBoxRevisionError` with a typed reason, matching the sibling relationship-closure guard at `:943-947`. Both call sites (`:948-951` and `:767-769`) need it; or
- **Widen the claim.** If DrawingML text boxes are in fact handled correctly — and in the cases I ran the output *was* correct — then `CONFORMANCE.md:155-157` is the thing that is wrong, and the exclusion should be dropped along with the scenarios that justify dropping it.

Either is fine. What is not fine is a published conformance claim whose enforcement depends on an accidental `undefined !== undefined`.

## Honest note on severity

In every case I ran, the produced redline was correct: revisions landed in the right branch, including when only the `mc:Choice` branch of a twin box changed (see #794, section C). So this is **not currently shipping a wrong document.** It is shipping a **conformance claim that is unbacked** — the exclusion is real in the docs, absent from the code, and untested either way. The failure mode is that the next change to the text-box path assumes the exclusion holds.

`openspec/changes/compare-vml-text-box-stories/tasks.md` item **7.10** is also still unchecked ("Run mandatory pre-submit checks, publish through a focused PR, and perform a real-document smoke") even though #730 merged, which may be related.

## Existing issues checked

No issue in the repo (283, all states) names DrawingML text boxes or `scaffoldFingerprint`. #713 (open, tracked revisions inside `w:txbxContent`) and #688 (open, unrepresentable container aborts the comparison) are the parents of this area, but neither covers an unenforced exclusion. #754 mentions DrawingML only incidentally, about header/footer section deletion. Not a duplicate.

## Bug or consumer lesson?

**Bug in safe-docx**, and specifically a spec-compliance bug rather than an output bug: the repository publishes a conformance registry, and this entry does not hold.

stevenobiajulu:
Fixes #795.

## What was wrong

`spec-compliance/CONFORMANCE.md` (ECMA-PART4-14-9-1-1) publishes that the `w:txbxContent` comparison claim *does not cover DrawingML text boxes*. The code did not enforce it. `scaffoldFingerprint()` returned `undefined` for a DrawingML box on **both** sides, so the guard compared `undefined !== undefined`, read `false`, and admitted the box. The sibling relationship-closure guard three lines above tests `=== undefined ||` explicitly — that contrast is the tell. The ancillary-part copy had the same hole spelled the other way round.

**This is an unbacked conformance claim, not a live output defect.** In every case tried the produced redline was correct. What was shipping was a published document asserting a protection the code did not provide.

## The gap this PR shipped with is now closed

The first version of this PR said, honestly, that **no real-world document exercised the changed guard**, and declined to hand-write XML on the grounds that it tests the scaffolding rather than the code. That was accurate: no `.docx` in the corpus contained a text box at all, so every text-box test in the repository was written against hand-built XML.

Two Word-authored fixtures now close it:

- `tests/test_documents/text-box/source.docx`
- `tests/test_documents/text-box/revised.docx`

A before/after pair authored in Word for Mac. Two text boxes whose text changes (`Text box 1` → `1st Text box is first. . .`, `Text box 2` → `This is text box number two.`) plus one body paragraph that does not — the control that catches a guard which has become over-broad and refuses documents wholesale. Provenance and storage shape are documented in `tests/test_documents/text-box/README.md`, following the convention set by `tests/test_documents/redline/README.md`. Content is fully synthetic; the only personal string in either package is the repo owner's name in `dc:creator`, which is on every commit here.

**Word-authored vs Word-derived — stated plainly, because it matters.** Two of the new cases run the untouched files: the counting case, and the twin admission case, which is what Word actually produces. The standalone-DrawingML and non-`v:shape` cases run markup *derived* from those files by keeping one `mc:AlternateContent` branch and discarding the twin. That is stronger than hand-writing the XML — every byte inside the surviving branch is Word's — but it is **not** a claim that some producer emits a bare `wps:txbx` with no wrapper. No such fixture is in the corpus. A fixture emitted naturally in that shape would be a strict improvement.

## What #803 does on a real document

#803 (merged, `d013a15`) changed text-box counting to count by *rendered branch* rather than stored copy, and merged without ever having been run against a Word-authored document, because none existed. These fixtures are that document. Answer:

```
raw w:txbxContent elements       : 4
groupElementsByTagNameNS         : 2   <- visual count
selectedElementsByTagNameNS      : 2
```

**Two, correctly** — 2 authored boxes, 4 stored copies. And the locators now number the way a reader does. On the pre-#803 base (`8035dce`) the same pair reports four changed containers with duplicated paragraph ids:

```
#w:txbxContent[0] (paragraph 4E88D6C6), #w:txbxContent[1] (paragraph 4E88D6C6),
#w:txbxContent[2] (paragraph 6BD23BC2), #w:txbxContent[3] (paragraph 6BD23BC2)
```

On `main` it reports two, one per authored box:

```
#w:txbxContent[0] (paragraph 4E88D6C6), #w:txbxContent[1] (paragraph 6BD23BC2)
```

**#803's merged fix is correct on a real Word document.** A new test pins it from the other direction: strip the twin from the *second* box only and the refusal names `#w:txbxContent[1]` — where a walk counting stored copies would say `[2]`, the second copy of the *first* box, sending whoever is chasing the diagnostic somewhere that does not exist.

## The two collateral regressions, verified on real markup

**1. Non-`v:shape` VML hosts.** `v:textbox` belongs to `EG_ShapeElements`, which `CT_Shape` shares with `CT_Rect`, `CT_RoundRect`, `CT_Oval` and others. Recognising the host by the literal name `v:shape` reported "no scaffold" for a schema-valid `v:rect` box and would have refused VML the engine handles fine. Rehosting Word's own `mc:Fallback` branch on each shape type: `v:shape`, `v:rect`, `v:roundrect`, `v:oval` — all admitted in place.

**2. `mc:AlternateContent` twins.** The commonest text box Word produces. Not merely admitted — **exactly reversible**, which is the claim that actually justifies admitting it. On the real pair: accept-all reproduces the revised text, reject-all the original, both stored branches of each visual box agree, and the DrawingML scaffold is untouched.

## Tests, and proof they can go red

Five new real-document cases on top of the existing synthetic ones. Every check is isolated by mutation rather than asserted:

| Mutation | Reds |
|---|---|
| revert the `=== undefined` fail-closed test (restore the #795 bug) | standalone refusal, visual locator (+2 synthetic) |
| remove the `mc:AlternateContent` carve-out | twin round trip, body control, visual locator (+1) |
| match the host by literal `v:shape` | non-`v:shape` host control (+1) |
| disable `mc:AlternateContent` grouping in `markupCompatibility.ts` | counting case, visual locator |
| splice only the rendered copy, leaving the twin stale | twin round trip, body control (+3) |
| none | 27 passed |

The last mutation is what gives the branch-synchronization assertion teeth: a redline that updated only the rendered copy would leave the document saying two different things depending on which Word opened it.

## Peer review

`/peer-review --codex` (`--code` lane), re-run after the fixtures landed because the PR had materially changed. Stored at `.peer-review/conformance-guards/`. Verdict **REQUEST CHANGES**, with a full execution trace: Codex independently re-ran all four mutations, decoded every `o:gfxdata` blob in both fixtures, and confirmed the two headline claims —

- the real-document test **does** exercise the changed guard, not merely pass alongside it. Neutralising only the `undefined` condition leaves `failure === undefined`, proving no earlier guard (relationship closure, unsupported story, reconstruction mode) is doing the refusing;
- `main` reports **two visual boxes, four stored copies**, so the fixture genuinely covers #803.

Its three blocking findings: assert the round trip rather than the presence of revisions; correct the "real document" language for the derived cases; and reconcile the `mc:AlternateContent` carve-out with the published exclusion. The first two are fixed in `5bdced6`. Advisories on the exact refusal reason, the body-test label, and real-fixture locator evidence are also fixed. The advisory to strip `dc:creator` is declined — it is the repo owner's name, already on every commit in this public repo.

## Known residual, not fixed here — needs a decision

Codex's third blocking finding is legitimate and this PR does **not** resolve it. `CONFORMANCE.md` says the claim "does not cover DrawingML text boxes", full stop. But `scaffoldPairingReason()` carves out every copy under `mc:AlternateContent`, *including the DrawingML `mc:Choice` Word renders* — and the new fixture proves that is the ordinary Word representation. So what this PR enforces is the exclusion for a **standalone** DrawingML box; the twinned case is admitted exactly as before. The title has been narrowed to say so.

Three ways out, none of them taken here:

1. **Validate the visual group using its VML twin as the scaffold.** This is the `mc:AlternateContent`-aware story walk's job, which is #794's subject; pre-empting it is what this PR set out not to do.
2. **Refuse the selected DrawingML case.** Rejected on evidence: the twinned case round-trips exactly on a real document, so refusing it would refuse something demonstrably safe and break the commonest text box Word makes.
3. **Narrow the published exclusion** to say that twinned DrawingML/VML Word shapes are covered and standalone DrawingML is not.

(3) is the accurate description of the code as it now stands, and the round-trip evidence above is what would support it. But `CONFORMANCE.md` is generated from `spec-compliance/registry/*.md`, and re-scoping a published compliance claim is a deliberate decision rather than a side effect of adding tests — so it is flagged here rather than made unilaterally. Tracked against #794.

**This PR is a strict improvement regardless:** before it, the guard could never fire at all; after it, it fires for standalone DrawingML, and the twinned case behaves exactly as it does on `main`.

## Verification

- `npm run test:run -w @usejunior/docx-compare` — 60 files, 808 passed, 106 skipped
- `npm run lint -w @usejunior/docx-compare` — clean
- `npm run check:conformance-citations` — OK (125 registry entries, 51 XSDs)
- `npm run check:conformance-doc` — OK (no hand edits to the generated doc)
- Merged up to `main` at `d013a15`, so #803's walk is in and the guards are exercised together.


stevenobiajulu:
## Smoke on the branch — PASS (5/5 real documents), plus one honest gap

CI is green (25 checks pass, 2 skipped, `mergeState=CLEAN`). **Automerge is deliberately NOT armed** and this PR is not merged — see the gap below and the #794 dependency in the description.

### Real-document smoke

Five real NVCA standard forms (pulled by the open-agreements downloader), each given a realistic edit and redlined end-to-end through `compareDocumentsAtomizer`, then accept/reject round-tripped:

| document | bytes | mode | w:ins / w:del | accept == revised | reject == original |
|---|---|---|---|---|---|
| indemnification agreement | 70,578 | inplace | 1 / 0 | yes | yes |
| voting agreement | 91,066 | inplace | 1 / 0 | yes | yes |
| management rights letter | 31,450 | inplace | 2 / 2 | yes | yes |
| stock purchase agreement | 140,221 | inplace | 1 / 1 | yes | yes |
| investors' rights agreement | 135,159 | inplace | 1 / 0 | yes | yes |

No regression on real documents: every one still redlines in `inplace` mode with no fallback.

### The gap, stated plainly

**None of those five documents contains a text box** (`w:txbxContent: 0`, `mc:AlternateContent: 0`). Neither does any other real `.docx` in the repo corpus — I scanned them all. So the smoke proves this PR does not *regress* real documents, but it does **not** exercise the changed guard on a real-world document, because no such document exists to hand.

I tried to author one rather than leave the hole: LibreOffice would not export a text-frame flat-ODT, and Word's AppleScript surface refused every route into a text frame's content. Rather than hand-write the XML — which would be testing my own scaffolding, exactly what the smoke rules forbid — I am reporting it as a gap.

What the guard path *is* backed by: the two reproductions in the description, 21 unit tests, and three mutation-verified isolations (each check reds exactly one test and no other). What it is not backed by is a Word-authored text-box document. If someone has one, it is worth running before this merges — particularly an `mc:AlternateContent` twin, since that is the case the carve-out protects and the case #794 must ultimately answer.

stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `8fb4906fccd85cb48ea376cd17643f34db98707b`
Built from: `main` @ `8fb4906`
Smoke: clean rebuild of `docx-core` + `docx-compare` from merged `main`, full suite, then a real-document end-to-end drive of the changed guard.

**Steps**
- ✅ clean build (both packages, `dist/` wiped first)
- ✅ lint (`tsc --noEmit`)
- ✅ tests — **894 passed**, 27 skipped, 61 files
- ✅ real-document end-to-end (below)

**Real-world fixtures**

- `nvca-stock-purchase-agreement` (via `ensureSourceDocx` cache, `~/.open-agreements/cache/`)
  - Real edit applied to the first substantive run, compared in place
  - Compared without refusal (`mode=inplace`), 1× `w:ins` produced; accept-all restores the edit, reject-all removes it
- `ILPA-Model-Limited-Partnership-Agreement-WOF_v2.docx` (167KB real published model agreement)
  - Compared in place, tracked insertion produced, reject-all restores the original
  - Text-box story count preserved — **nothing spuriously refused** by the newly fail-closed guard
- `ILPA-Model-Limited-Parnership-Agreement-Deal-By-Deal_v1.docx` (171KB) — same, all four checks pass
- `tests/test_documents/text-box/{source,revised}.docx` (Word-authored, added by this PR — the only document in the corpus that reaches the changed guard)
  - Word stored 2 authored boxes as **4** `w:txbxContent` across 2 `mc:AlternateContent` twins
  - Twin **admitted**, 8× `w:ins` / 6× `w:del` inside the stories
  - reject-all == source and accept-all == revised **in every stored copy**, unrendered `mc:Fallback` included
  - Unchanged body paragraph carries no revision
  - Standalone DrawingML box (twin stripped) **refused** — `the containing VML shape scaffold changed or could not be paired`. This is the published exclusion, actually enforced.
  - Refusal on the second box names `#w:txbxContent[1]`, not `[2]` — #803's visual ordinal, on a real document

**What #803 does on a real document**: reports **2** text boxes (4 stored copies), and the locators number the way a reader does. Pre-#803 the same pair reported four containers with duplicated paragraph ids. The merged fix is correct.

Visual evidence: rendered via LibreOffice and opened in-session — `/tmp/automerge-smoke-800-1786467152/textbox-redline-1.png`. Both text boxes render with their scaffolds intact, insertions underlined and deletions struck *inside* the boxes, and `Body text` sitting above them in plain black with no markup. First text-box redline this repo has been able to look at.

⚠️ **Advisory gate returned WARN, not PASS.** `usejunior-llm-gate` reported `Overall: WARN (4 pass, 1 warn, 0 error, 12 skipped)`. The warn is legitimate and worth a follow-up: `hasAlternateContentAncestor` (`textBoxRevisionSafety.ts:274`) re-implements the `mc:AlternateContent` tag check against a local `MC_NS` constant instead of reusing the canonical `isAlternateContent` helper and `MC_NAMESPACE` from `markupCompatibility.ts`. That duplication exists because this branch was written before #803 landed that module. Not a correctness bug — the two constants are the same URI — but it is exactly the "re-derived fact vs canonical source" drift the gate exists to catch, and it should be collapsed onto the canonical helper.

**Known residual, unchanged by this smoke**: the `mc:AlternateContent` carve-out means the published exclusion is enforced for *standalone* DrawingML only, while `CONFORMANCE.md` still says the claim "does not cover DrawingML text boxes" without qualification. See the PR body — the round-trip evidence above is what would support narrowing that wording, but re-scoping a published compliance claim was left as a deliberate decision rather than a side effect. Tracked against #794.

**Cleanup**
- ⚠️ remote branch already deleted — auto-delete is enabled
- ✅ removed worktree `.worktrees/795-drawingml-scaffold-20260811`
- ✅ deleted local branch `795-enforce-drawingml-exclusion-20260811`

Log: `/tmp/automerge-smoke-800-1786467152.log` (local).

PR #801, opened 2026-08-11
fix(docx-core): reject failed conversions instead of inferring success from an output file

### LINKED ISSUE #796: docx-core: runConvert() swallows the converter exit status and declares success on output-file existence
## Summary

**Low severity — a latent weak check, not a live defect.** See "Honest note on severity" below.

This is the `git bundle verify`-shaped check: a verification helper whose pass condition is "the thing did not obviously explode" rather than "the content is right."

`packages/docx-core/src/integration/generation-probes.ts:36-79`. `runConvert()` shells out to LibreOffice, **catches and discards the subprocess error** (`:63-66` — `catch (err) { diag = String(...) }`, no rethrow), and then decides success purely on whether an output file exists:

```ts
// generation-probes.ts:59-75
let diag = '';
try {
  const r = await execFileAsync(soffice, args, { timeout: 45_000, killSignal: 'SIGKILL', maxBuffer: 8 * 1024 * 1024 });
  diag = String(r.stderr ?? '') || String(r.stdout ?? '');
} catch (err) {
  const e = err as { stdout?: unknown; stderr?: unknown; message?: string };
  diag = String(e.stderr ?? e.stdout ?? e.message ?? '');
}

const outPath = path.join(outDir, `probe.${outExt}`);
if (!existsSync(outPath)) {
  throw new Error(
    `LibreOffice could not convert the generated package to ${outExt} — likely a load failure ` +
      `(the recovery-dialog class of bug).\nsoffice output:\n${diag.trim() || '(no output)'}`,
  );
}
return readFileSync(outPath);
```

A `soffice` run that exits non-zero while still emitting a truncated or partially-recovered `probe.docx` is reported as a passing probe. `diag` is attached to the error on the file-*missing* path, so it is not lost in general — but on exactly the path that matters here, where the file exists despite the failure, both the non-zero status and the captured stderr are dropped on the floor.

## Reproduction

Stand in for LibreOffice with a script that writes a truncated output file and then fails. Save as `fake-soffice.sh` and `repro-probe.mjs` in the repo root; run after `npm run build`.

```sh
#!/bin/sh
# fake-soffice.sh — emits a truncated output file, then FAILS.
outdir=""; prev=""
for a in "$@"; do
  if [ "$prev" = "--outdir" ]; then outdir="$a"; fi
  prev="$a"
done
printf 'PK\003\004TRUNCATED' > "$outdir/probe.docx"
echo "Error: source file could not be loaded" >&2
exit 1
```

```js
// repro-probe.mjs — node repro-probe.mjs ./fake-soffice.sh
const { probeDocxIdentity } = require('./packages/docx-core/dist/integration/generation-probes.js');

const fakeSoffice = process.argv[2];
try {
  const probe = await probeDocxIdentity(Buffer.from('not a real docx'), fakeSoffice);
  console.log('  probeDocxIdentity RETURNED (no error raised)');
  console.log('  savedPackage bytes :', probe.savedPackage.length);
  console.log('  savedPackage       :', JSON.stringify(probe.savedPackage.toString('latin1')));
} catch (err) {
  console.log(`  threw: ${err?.constructor?.name}: ${err?.message}`);
}
```

## Actual output

Run at `8017621`:

```
  probeDocxIdentity RETURNED (no error raised)
  savedPackage bytes : 13
  savedPackage       : "PK\u0003\u0004TRUNCATED"
```

The converter exited 1 and wrote `Error: source file could not be loaded` to stderr. The probe returned a 13-byte truncated "package" as a success, and neither the exit status nor the message reached the caller.

## Expected

- The non-zero exit status and the captured stderr are surfaced rather than swallowed, on both paths.
- Success is a statement about content, not about a file existing at a path.

## Honest note on severity

**Low.** The two current call sites are not badly exposed:

- `probeDocxIdentity` (`:90`) — `generation-package-structure.test.ts:158-168` asserts `savedPackage.length > 0`, reloads the package, and asserts the extracted text contains the expected sentence. Genuinely a content check.
- `probeDocxToPdf` (`:100`) — `generation-package-structure.test.ts:184-191` asserts only `pdf.length > 0` and that the first five bytes are `%PDF-`. That is a magic-byte check, not a content check; a truncated PDF emitted before an abort would pass it.

So the weakness is not currently shipping a wrong result. It is inherited by the next caller, which is the reason to fix it while it is cheap.

## What I checked and did NOT find

The rest of the verification tree does not have this problem. Grepping `exitCode === 0`, `status === 0`, `code === 0`, `.not.toThrow`, `resolves.toBeDefined`, `toBeTruthy()` across non-test production code returns two other hits, both sound:

- `packages/docx-compare/src/baselines/atomizer/leanXmlVerifier.ts:1518` — exit 0 is only the *entry* condition; the code then enforces fatal-UTF-8 decoding (`:1520-1526`), a trailing-newline/size/byte-count protocol-v6 envelope check (`:1527-1531`), and separate rejects for timeout and output-limit breach. Then it parses and validates the payload.
- `scripts/check_emitted_document_schema.mjs:84` — an xmllint loop, where exit 0 genuinely *is* the validation result.

There is no backup/bundle/archive verification helper in the tree at all; `npm run customer-signals:backup` and `:integrity` are SQLite signal-DB commands, unrelated to DOCX round-trip.

**The strongest weak check in this repository is not an exit code.** It is the round-trip integrity gate at `pipeline.ts:490-496`, which asserts that a `w:t` + `w:delText`-only text projection is equal on both sides. Any content class outside that set can be silently lost with the gate reporting success — structurally identical to `git bundle verify` returning 0. Filed separately as #793.

## Existing issues checked

No issue mentions `runConvert`. #627 (open) concerns LibreOffice headless startup crashes and parallel probe memory amplification on macOS — the same file's neighbourhood, but about the converter dying, not about its exit status being swallowed when it half-succeeds. Not a duplicate.

stevenobiajulu:
Fixes #796.

## What was wrong

`runConvert()` caught the LibreOffice subprocess error, **discarded it**, and then decided the probe had passed purely on whether a file existed at the output path. A converter that fails *after* dropping a partial file satisfies that. The captured diagnostic was already attached on the file-*missing* path — it was lost on exactly the path that mattered.

**Low severity and latent.** The two current call sites are not badly exposed; `probeDocxIdentity`'s test does check content. But `probeDocxToPdf`'s asserts only `length > 0` and the `%PDF-` magic bytes, which a truncated render passes, and the weakness is inherited by whatever calls these next.

## Reproduction, at `8035dce`

A stand-in `soffice` that writes a truncated file and exits 1:

```
$ node repro-probe.mjs ./fake-soffice.sh
  probeDocxIdentity RETURNED (no error raised)
  savedPackage bytes : 13
  savedPackage       : "PKTRUNCATED"
```

The converter exited 1 and wrote `Error: source file could not be loaded` to stderr. The probe returned a 13-byte truncated "package" as a success, and neither the exit status nor the message reached the caller. After the fix:

```
  threw: ConvertProbeError: LibreOffice failed while converting the generated package to docx.
soffice status: exit 1
soffice output:
Error: source file could not be loaded
```

## The measurement that shaped the fix

The obvious reading of the issue title is "check the exit status". On its own that would catch almost nothing. Measured against real LibreOffice 25.8:

| case | exit | stderr | output file |
|---|---|---|---|
| valid docx to docx | **0** | (macOS noise only) | written |
| valid docx to pdf | **0** | (macOS noise only) | written |
| **unloadable** source | **0** | `Error: source file could not be loaded` | **none** |

LibreOffice exits 0 even when it cannot load the source. So the status check does **not** subsume the output checks, and the output checks do not subsume it. Both are kept, and the module docstring records the measurement so a future reader does not collapse them.

The same table answers the false-failure question: no observed conversion exits non-zero on success, so adding the status check adds a rejection without removing one. The contrasting `runSoffice()` in `libreoffice-oracle.ts`, which deliberately tolerates a non-zero exit, is a different process contract — its macro terminates the desktop. Untouched.

## What changed

Success is now a statement about the artifact, not about a file existing at a path:

- exit status and terminating signal are recorded rather than discarded, and a failed run is rejected **before** the output path is consulted;
- a zero-length output is rejected;
- the re-saved `.docx` must be a readable package whose main part resolves through the package-level `officeDocument` relationship — `word/document.xml` is a convention, not an OPC rule, so the relationship is followed instead of assumed;
- the rendered `.pdf` must carry a `%%EOF` trailer, not just a `%PDF-` header;
- a new `ConvertProbeError` carries the exit status and captured output on **every** rejection path, and both probe results expose the same diagnostics.

The real-LibreOffice probes are now gated on `probeSofficeUsable()`, so a binary that is present but cannot launch (sandbox SIGABRT) skips rather than failing with exit 134.

## Tests, and proof they can go red

The probes are gated on a real LibreOffice, which CI does not install — so the new tests drive them against a **stub converter**, which both runs in CI and can script failure modes a real binary will not reproduce on demand. Every rejection is paired with a green control differing only in the property under test.

One test exists because of a negative control that came back green when it should not have: neutralising only the exit-status predicate left the whole suite passing, because the truncated artifact also failed the package check. The status check was covered only *incidentally*. `rejects a failed run even when its output would have passed inspection` — a converter that exits 77 while writing a perfectly readable package — isolates it:

| Mutation | Result |
|---|---|
| disable only the exit-status predicate | the isolating test fails |
| disable only the `%%EOF` predicate | the truncated-PDF test fails |
| none | `8 passed` |

Both bugs in this batch are instances of one class: a check that passes by construction. `undefined !== undefined` and "the file exists so it worked" are two spellings of the same mistake — and a test that passes by construction is that mistake one level up, which is what the control above caught.

## Verification

- `npm run test:run -w @usejunior/docx-core` — 130 files passed, 1 skipped; 1333 passed
- Real-LibreOffice gated probes run and pass locally with the new content checks — no false failures
- Peer-reviewed with `/peer-review --codex` (`--code` lane, full execution trace). Verdict APPROVE-WITH-CHANGES; both must-fix items (resolve the main part through `_rels/.rels`; gate on `probeSofficeUsable()`) are addressed here.

Fixtures are fully synthetic.

### DIFF
--- a/packages/docx-core/src/integration/generation-package-structure.test.ts
+++ b/packages/docx-core/src/integration/generation-package-structure.test.ts
@@ -1,5 +1,5 @@
 import JSZip from 'jszip';
-import { describe, expect } from 'vitest';
+import { beforeAll, describe, expect } from 'vitest';
 import { testAllure, type AllureBddContext } from '../testing/allure-test.js';
 import { DocxDocument } from '../primitives/document.js';
 import { parseXml } from '../primitives/xml.js';
@@ -7,7 +7,7 @@ import { childElements } from '../primitives/dom-helpers.js';
 import { generateDocx } from '../generation/compile.js';
 import { checkGeneratedPackage } from '../generation/structural-checks.js';
 import type { DocumentSpec } from '../generation/types.js';
-import { resolveSoffice } from './libreoffice-oracle.js';
+import { probeSofficeUsable, resolveSoffice } from './libreoffice-oracle.js';
 import { probeDocxIdentity, probeDocxToPdf } from './generation-probes.js';
 import { writeIntegrationArtifact, getIntegrationOutputModeLabel } from './output-artifacts.js';
 
@@ -145,9 +145,26 @@ if (!soffice) {
 }
 
 describeProbes('Traceability: LibreOffice probes over generated packages', () => {
+  // `resolveSoffice()` proves the binary EXISTS, not that it can launch: under
+  // a restricted shell it dies with SIGABRT during init, which would FAIL these
+  // tests rather than skip them (observed: exit 134, "Abort trap: 6"). Probe
+  // launchability once and skip the body when it is unusable.
+  let sofficeUsable = false;
+  beforeAll(async () => {
+    sofficeUsable = soffice ? await probeSofficeUsable(soffice) : false;
+    if (!sofficeUsable) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        '[generation-package-structure] SKIP: soffice present but not launchable ' +
+          '(sandbox abort); cross-reader probes skipped.',
+      );
+    }
+  }, 60_000); // probeSofficeUsable launches soffice; the 10s default hook timeout is too short.
+
   test.openspec('[SDX-GEN-090] LibreOffice identity round-trip succeeds')(
     'Scenario: LibreOffice identity round-trip succeeds',
     async ({ given, when, then, attachPrettyJson }: AllureBddContext) => {
+      if (!sofficeUsable) return;
       let generated!: Buffer;
       await given('a generated full-package document', async () => {
         generated = await generateDocx(phase1Spec());
@@ -174,6 +191,7 @@ describeProbes('Traceability: LibreOffice probes over generated packages', () =>
   test.openspec('[SDX-GEN-091] headless PDF conversion succeeds')(
     'Scenario: headless PDF conversion succeeds',
     async ({ given, when, then }: AllureBddContext) => {
+      if (!sofficeUsable) return;
       let generated!: Buffer;
       await given('a generated full-package document', async () => {
         generated = await generateDocx(phase1Spec());
--- a/packages/docx-core/src/integration/generation-probes.test.ts
+++ b/packages/docx-core/src/integration/generation-probes.test.ts
@@ -256,6 +256,136 @@ describe('Traceability: generation probes reject failed conversions', () => {
     });
   });
 
+  test('rejects a docx with no package relationships part', async ({
+    given,
+    when,
+    then,
+  }: AllureBddContext) => {
+    const soffice = await given('a converter that writes a zip with no _rels/.rels', async () =>
+      makeStubConverter({
+        name: 'writes-no-package-rels',
+        ext: 'docx',
+        writes: await createZipBuffer({ 'word/document.xml': '<w:document/>' }),
+      }),
+    );
+
+    let failure: unknown;
+    await when('the identity probe runs', async () => {
+      try {
+        await probeDocxIdentity(GENERATED, soffice);
+      } catch (error) {
+        failure = error;
+      }
+    });
+
+    await then('a readable zip is still not a package', () => {
+      expect(failure).toBeInstanceOf(ConvertProbeError);
+      expect((failure as Error).message).toContain('_rels/.rels');
+    });
+  });
+
+  test('rejects a docx whose relationships declare no main document', async ({
+    given,
+    when,
+    then,
+  }: AllureBddContext) => {
+    const soffice = await given('a converter that writes rels without an officeDocument type', async () =>
+      makeStubConverter({
+        name: 'writes-no-office-document-rel',
+        ext: 'docx',
+        writes: await createZipBuffer({
+          '_rels/.rels':
+            `<?xml version="1.0"?><Relationships ` +
+            `xmlns="http://schemas.openxmlformats.org/package/2006/relationships">` +
+            `<Relationship Id="rId1" ` +
+            `Type="http://schemas.openxmlformats.org/package/2006/relationships/metadata/core-properties" ` +
+            `Target="docProps/core.xml"/></Relationships>`,
+          'word/document.xml': '<w:document/>',
+        }),
+      }),
+    );
+
+    let failure: unknown;
+    await when('the identity probe runs', async () => {
+      try {
+        await probeDocxIdentity(GENERATED, soffice);
+      } catch (error) {
+        failure = error;
+      }
+    });
+
+    await then('the missing officeDocument relationship is named', () => {
+      expect(failure).toBeInstanceOf(ConvertProbeError);
+      expect((failure as Error).message).toContain('officeDocument relationship');
+    });
+  });
+
+  test('rejects a docx whose main-document relationship dangles', async ({
+    given,
+    when,
+    then,
+  }: AllureBddContext) => {
+    const soffice = await given('a converter that writes rels pointing at an absent part', async () =>
+      makeStubConverter({
+        name: 'writes-dangling-main-part',
+        ext: 'docx',
+        writes: await createZipBuffer({
+          '_rels/.rels':
+            `<?xml version="1.0"?><Relationships ` +
+            `xmlns="http://schemas.openxmlformats.org/package/2006/relationships">` +
+            `<Relationship Id="rId1" ` +
+            `Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/officeDocument" ` +
+            `Target="word/absent.xml"/></Relationships>`,
+          'word/document.xml': '<w:document/>',
+        }),
+      }),
+    );
+
+    let failure: unknown;
+    await when('the identity probe runs', async () => {
+      try {
+        await probeDocxIdentity(GENERATED, soffice);
+      } catch (error) {
+        failure = error;
+      }
+    });
+
+    await then('the probe follows the relationship rather than guessing the path', () => {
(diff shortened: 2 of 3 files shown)


stevenobiajulu:
## Post-merge smoke — PASS (2/2)

Run against freshly-pulled `main` at `9c85a2a`, driving the merged probe code through a **real document** (an NVCA standard form, 70,578 bytes) and a real LibreOffice 25.8 — not a synthetic fixture.

```
soffice: /opt/homebrew/bin/soffice
real NVCA form: 70578 bytes

PASS  identity probe accepts the real NVCA form
        savedPackage=53196B exit=0 signal=null
PASS  truncated real NVCA is REJECTED (#796 guard)
        ConvertProbeError: LibreOffice could not convert the generated package to docx —
        likely a load failure (the recovery-dialog class of bug). | soffice status: exit 0

===== POST-MERGE SMOKE #801: 2/2 passed =====
```

The negative control is the interesting one, and it confirms the measurement this PR was built on: truncating the real document made LibreOffice **exit 0** and produce no output. The exit-status check alone would not have caught it — the output check is what fired. Both checks are load-bearing on real data, which is why neither was allowed to subsume the other.

### A field observation worth recording

Mid-smoke, LibreOffice on this machine became wedged (unrelated concurrent `soffice` instances), and every conversion hit the probe's 45s timeout. Same real document, same binary, run under both versions:

| | reported |
|---|---|
| pre-merge (`8035dce`) | `Error: LibreOffice could not convert the generated package to docx — likely a load failure (the recovery-dialog class of bug)` with an **empty** `soffice output:` |
| post-merge (`9c85a2a`) | `ConvertProbeError: LibreOffice failed while converting the generated package to docx. soffice status: killed by SIGKILL` |

The old code misdiagnosed a timeout as a load failure and had nothing to show for it. That was not a designed test case — the environment supplied it — but it is the clearest evidence that surfacing the status is worth having.

Full `@usejunior/docx-core` suite on main: 130 files passed, 1 skipped; 1337 passed.

stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `9c85a2a2f9d6169f2be5483e1dc340f15c934a0a` (squash)
Built from: `main` @ `9c85a2a`
Smoke: `npm run build && npm run lint:workspaces && npm run test:run` + a real-document drive of `probeDocxIdentity` / `probeDocxToPdf` through a real LibreOffice 25.8

**Steps**
- ✅ build (all workspaces)
- ✅ lint (`lint:workspaces`, exit 0)
- ✅ tests — 1350 passed, 1 failed, 2 expected-fail, 1 skipped (131 files). The single failure is **pre-existing**: `nvca-structural-regression.test.ts > returns structured results for selected comments and nested-definition mutations` reproduces identically on pre-merge `main` @ `8035dce`.
- ✅ **This PR fixed two pre-merge failures.** On `8035dce` the two `generation-package-structure` LibreOffice probes FAILED with sandbox `soffice` aborts (`Killed: 9`). With `probeSofficeUsable` they now skip cleanly instead of reporting a false regression.
- ✅ real-document probe drive (below)

**Real-world fixtures**
- `tests/test_documents/nvca-regression/source.docx` — real NVCA form, 140,221 bytes
  - `probeDocxIdentity` → re-saved package 93,004 bytes, soffice exit 0; `_rels/.rels` resolved and the officeDocument-relationship target was present in the package.
  - `probeDocxToPdf` → 1,095,500-byte PDF with `%PDF-` header **and** a `%%EOF` trailer in the last 1 KB, i.e. the completeness check the PR added actually ran on real render output.
- `tests/test_documents/open-agreements/common-paper-mutual-nda.docx` — real Common Paper Mutual NDA, 23,473 bytes
  - `probeDocxIdentity` → re-saved package 18,352 bytes, soffice exit 0.
- **Negative control (the #796 shape):** the same real NVCA form truncated to its first 4,096 bytes (real PK header, incomplete zip) was **rejected** with `ConvertProbeError`, not read back as a success.
  - Notable: LibreOffice exited **0** on that run and simply produced no output — exactly the case the PR's comment documents, and proof the status check does not subsume the file checks. The pre-merge `existsSync` pass condition is what this guards.

Visual evidence: skipped — the change is converter-verdict logic and is structurally invisible on a rendered page.
Log: `/tmp/automerge-smoke-801-20260811-113946.log` (local).

**Advisory gates audited (not just their green checkmarks)**
- `Aggregate and post review` (usejunior-llm-gate): **Overall: PASS** (6 pass, 0 warn, 0 error, 11 skipped, 17 total). No `Unable to verify` / provider-failure markers.
- `codecov/patch`: 85.11%, 7 partials in `generation-probes.ts` — advisory, not a required context.

**Cleanup**
- ✅ remote branch `796-surface-converter-exit-status-20260811` already deleted (auto-delete enabled); `git remote prune origin` pruned the stale ref
- ✅ worktree `.worktrees/796-convert-exit-status-20260811` already removed
- ✅ no local branch remained

Issue #799, opened 2026-08-11
docx-compare: formatting fidelity check misses a changed symbol font on w:sym


stevenobiajulu:
## Summary
If comparison output rewrote `<w:sym w:font="Wingdings" w:char="F0A8"/>` into `<w:sym w:font="Symbol" w:char="F0A8"/>`, the page would show a different glyph and every safety gate would still pass. The text projection deliberately ignores `@w:font` (since #793, so the two legal spellings of one glyph compare equal), and the formatting fidelity check does not see `w:sym` at all. The comparison engine itself handles a real font change correctly: it emits a `w:del`/`w:ins` pair of `w:sym` runs. The gap is in the verifier. It is lower priority than #793 and #798 because the glyph is still present with the same codepoint.

The same blindness has a second effect. The literal spelling (the codepoint in `w:t` under a symbol font) contributes a character with `w:rFonts` in its key, while the `w:sym` spelling contributes none. A purely notational re-spelling therefore looks to `formattingFidelity` like a character appearing or disappearing.

## Reproduction
Checked on main @ fb0bd3af: still reproduces (`scratch-B/sym-fidelity.mjs` in the triage workspace). Run from the repo root with `node` after `npm run build`:
```js
// node repro-symbol-font-fidelity.mjs
import { createRequire } from 'node:module';
const require = createRequire(process.cwd() + '/package.json');
const { compareFormattingFidelity } = require('./packages/docx-compare/dist/tagged/formattingFidelity.js');

const W = 'http://schemas.openxmlformats.org/wordprocessingml/2006/main';
const sym = (font) =>
  `<w:document xmlns:w="${W}"><w:body><w:p>` +
  `<w:r><w:t xml:space="preserve">Alpha </w:t></w:r>` +
  `<w:r><w:sym w:font="${font}" w:char="F0A8"/></w:r>` +
  `<w:r><w:t xml:space="preserve"> Bravo</w:t></w:r>` +
  `</w:p></w:body></w:document>`;
const lit = (font) =>
  `<w:document xmlns:w="${W}"><w:body><w:p>` +
  `<w:r><w:t xml:space="preserve">Alpha </w:t></w:r>` +
  `<w:r><w:rPr><w:rFonts w:ascii="${font}" w:hAnsi="${font}"/></w:rPr><w:t>\uF0A8</w:t></w:r>` +
  `<w:r><w:t xml:space="preserve"> Bravo</w:t></w:r>` +
[...]

Issue #798, opened 2026-08-11
docx-compare: publication safety gate cannot see tabs, breaks, hyphen elements or empty paragraphs


stevenobiajulu:
The tagged publication gate proves accept/reject parity with a text projection that leaves out tabs, breaks, carriage returns, hyphen elements and empty-paragraph structure. A dropped construct of any of those kinds passes. This epic makes each one a checked claim and keeps unsupported drawings and objects explicitly outside the claim.

- [ ] #1026
- [ ] #1027
- [ ] #799

Related: #793, #831, #832, #1017

---
<details><summary>Original report (preserved from before the 2026-09-23 triage)</summary>

## Problem

The live tagged publication gate compares Accept/Reject projections using `extractRoundTripComparisonText` followed by text normalization. That projection intentionally includes `w:t`, `w:delText`, and `w:sym`, but excludes `w:tab`, `w:br`, `w:cr`, `w:noBreakHyphen`, empty-paragraph topology, drawings, and other non-text structures.

Normalization also collapses blank paragraph entries and trims leading/trailing whitespace. Consequently, a fault that drops some of these constructs can still receive a passing text-safety result.

This is a verifier-boundary defect, not evidence that the current tagged serializer drops every listed construct. The old issue's inplace/rebuild framing was removed because those implementations no longer exist.

## Required behavior

- Project stable authored characters for tabs, breaks, carriage returns, soft/no-break hyphens, and other supported text-bearing leaves.
- Add structural parity checks for empty paragraphs and topology that text normalization intentionally erases.
- Keep ordinary reader normalization from creating widespread false failures.
- Use fault injection to prove each new check can fail.
- Measure the before/after failure rate on the real comparison corpus.
- Keep unsupported drawings/objects explicitly outside the claim unless a structural inventory check covers them.

## Acceptance criteria

- A dropped supported construct cannot pass the tagged publication gate.
- Legitimate equivalent spellings do not make the gate reject all comparisons.
- The public safety report distinguishes text projection from structural parity.
- Focused synthetic fixtures cover each protected construct.

Re-triaged against `origin/main` `99109405` on 2026-08-21.

</details>

stevenobiajulu:
Audit re-triage after the tagged-only migration: the old inplace path is gone, but the live tagged publication gate still calls the same `extractRoundTripComparisonText`/normalization boundary. Its documentation explicitly excludes tabs, breaks, no-break hyphens, empty-paragraph structure, and other non-text constructs.

[...]

Issue #797, opened 2026-08-11
docx-core: w:sym glyphs are invisible to paragraph text, and a tracked edit spanning one deletes it untracked


stevenobiajulu:
## Goal
Make `w:sym` glyphs (checkboxes, Wingdings bullets and similar) a first-class part of the paragraph text coordinate space, without desynchronising the walkers that compute offsets, and stop tracked edits from deleting them without a `w:del`.

Checked on main @ fb0bd3af: the run-list repro still reproduces, and the untracked deletion that this report called latent is already reachable (see child 1). Correction to the original: the shared helper added by #793 is `projectSymbolRun()` (`symbol_run_content.ts:77`), not `symbolRunCharacter()`, and `getParagraphRuns` now filters `buildParagraphIndex` runs (`text.ts:28-37`).

- [ ] #1044
- [ ] #1045
- [ ] #1046

Related: #793, #799, #802

---
<details><summary>Original report (preserved from before the 2026-09-23 triage)</summary>

## Summary

`getParagraphRuns` drops a run whose only content is a `<w:sym/>`, so the glyph is invisible to `getParagraphText`, to every offset expressed in that text, and to `replace_text` on the MCP surface.

`packages/docx-core/src/primitives/text.ts` accumulates `runText` from `w:t` (plus `w:tab` → `\t` and `w:br` → `\n`) and never from `w:sym`. A run containing only `<w:sym w:font="Wingdings" w:char="F0A8"/>` therefore yields `runText === ''` and is discarded by the `if (runText)` guard before the run list is returned.

This is the second half of #793. #793 fixed the *comparison* projections — `extractRoundTripComparisonText` and `extractTextWithParagraphs` — which build a throwaway string for an equality test and own no coordinate space. This half is a different animal, and was deliberately deferred out of that PR: `getParagraphText` is the **definition** of the raw-offset coordinate space that paragraph editing, comment anchoring, footnote insertion and locator spans are all expressed in.

`packages/docx-core/src/primitives/symbol_run_content.ts` already exports `symbolRunCharacter()`, added by #793. The projection change itself is one line. Everything below is why it is not a one-line PR.

## What has to move in lockstep

Making a `w:sym` count as one character shifts every subsequent offset in that paragraph by one, and at least five walkers compute that same length independently:

[...]

PR #802, opened 2026-08-11
fix(docx-compare): project w:sym glyphs into the round-trip comparison text

### LINKED ISSUE #793: docx-compare: the round-trip integrity gate projects w:t + w:delText only, so losing a w:sym glyph passes verification
## Summary

`extractRoundTripComparisonText` — the projection the accept-all / reject-all safety gate compares — collects only `w:t` and `w:delText`. A `<w:sym w:font="Wingdings" w:char="F0A8"/>` contributes nothing to it.

Two consequences, in opposite directions, from the same cause:

1. **False negative (the serious one).** A document that has *lost* a symbol glyph and a document that still *has* it project to the same string. The gate at `pipeline.ts:490-496` compares that string, reports `normalizedIdentical`, and publishes the comparison as round-trip-safe. `INV-RT-001` is stated over the same element set (`verification/ROADMAP.md:46`: the projection "counts both `w:t` and `w:delText`"), so the formal invariant is satisfied while real content is gone.
2. **False positive.** OOXML has two legal spellings of one glyph — a literal private-use codepoint inside `w:t`, and `w:sym` — and Word renders them identically. A comparison engine may rewrite one into the other. The projection sees the two spellings as *different*, so a purely notational rewrite reads as a mass character loss.

The second is not hypothetical. On a real document, a `w:t`-only walker reported 36 "dropped checkbox" defects across a redline in which nothing had been dropped at all; the producing engine had rewritten literal U+F0A8 into the canonical `w:sym` spelling.

## Where

- `packages/docx-compare/src/fieldComparisonSemantics.ts:96-105` — the only content branch in the walk is `localName === 't' || localName === 'delText'`.
- `packages/docx-compare/src/baselines/atomizer/pipeline.ts:490-496` — the gate that compares that projection.
- `packages/docx-core/src/primitives/text.ts:103-109` — the same blindness on the MCP read/edit surface. Worse there: a run whose only content is a `w:sym` yields `runText === ''` and is dropped from the run list entirely by the guard at `:112`, so the glyph is invisible to `getParagraphText`, to offset arithmetic, and to `replace_text`.
- `packages/docx-compare/src/baselines/atomizer/formattingFidelity.ts:268` documents its own projection as "Visible text: concatenated `w:t` + `w:delText` character content."

The atomizer itself is fine — `w:sym` is in `LEAF_NODE_TAGS` (`packages/docx-compare/src/atomizer.ts:421`) and `premergeRuns` correctly refuses to merge across it (`premergeRuns.ts:28-39`, where `w:sym` is absent from `SAFE_RUN_CHILD_TAGS`). The gap is only in the *text projection*.

There is no code anywhere in the repo that maps a literal U+F000–U+F8FF codepoint to `w:sym` or back. The only Unicode normalization present is NFKC, used for fingerprints and style-name matching (`docx-core/src/primitives/content_fingerprint.ts:58`), and NFKC does not touch the private-use area.

## Reproduction

Fully synthetic; no real document involved. Save as `repro-sym.mjs` in the repo root and run with `node` after `npm run build`.

```js
// Fully synthetic. Run from the safe-docx repo root after `npm run build`:
//   node repro-sym.mjs
const { extractRoundTripComparisonText } = require(
  './packages/docx-compare/dist/fieldComparisonSemantics.js',
);

const W = 'http://schemas.openxmlformats.org/wordprocessingml/2006/main';
const doc = (body) =>
  `<?xml version="1.0" encoding="UTF-8" standalone="yes"?>` +
  `<w:document xmlns:w="${W}"` +
  ` xmlns:w14="http://schemas.microsoft.com/office/word/2010/wordml"` +
  ` xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"` +
  ` mc:Ignorable="w14"><w:body>${body}<w:sectPr/></w:body></w:document>`;

const wingdings = `<w:rPr><w:rFonts w:ascii="Wingdings" w:hAnsi="Wingdings"/></w:rPr>`;
const PUA = String.fromCharCode(0xf0a8); // the same glyph as a literal codepoint

// a) the glyph, spelled as w:sym
const asSym = doc(
  `<w:p><w:r><w:t xml:space="preserve">Alpha </w:t></w:r>` +
  `<w:r>${wingdings}<w:sym w:font="Wingdings" w:char="F0A8"/></w:r>` +
  `<w:r><w:t xml:space="preserve"> Bravo</w:t></w:r></w:p>`);

// b) the SAME glyph, spelled as a literal private-use character inside w:t
const asLiteral = doc(
  `<w:p><w:r><w:t xml:space="preserve">Alpha </w:t></w:r>` +
  `<w:r>${wingdings}<w:t>${PUA}</w:t></w:r>` +
  `<w:r><w:t xml:space="preserve"> Bravo</w:t></w:r></w:p>`);

// c) the glyph genuinely deleted
const deleted = doc(
  `<w:p><w:r><w:t xml:space="preserve">Alpha </w:t></w:r>` +
  `<w:r><w:t xml:space="preserve"> Bravo</w:t></w:r></w:p>`);

const esc = (s) => JSON.stringify(s).replace(/[\u007f-\uffff]/g,
  (c) => '\\u' + c.charCodeAt(0).toString(16).toUpperCase().padStart(4, '0'));

const a = extractRoundTripComparisonText(asSym);
const b = extractRoundTripComparisonText(asLiteral);
const c = extractRoundTripComparisonText(deleted);

console.log(`  projection of <w:sym w:char="F0A8"/>   : ${esc(a)}`);
console.log(`  projection of literal U+F0A8 in <w:t>  : ${esc(b)}`);
console.log(`  projection with the glyph DELETED      : ${esc(c)}`);
console.log('');
console.log(`  w:sym document == glyph-deleted document ?  ${a === c}   <-- expected false`);
console.log(`  w:sym document == literal-PUA document   ?  ${a === b}    <-- expected true`);
```

## Actual output

Run at `8017621`:

```
  projection of <w:sym w:char="F0A8"/>   : "Alpha  Bravo"
  projection of literal U+F0A8 in <w:t>  : "Alpha \uF0A8 Bravo"
  projection with the glyph DELETED      : "Alpha  Bravo"

  w:sym document == glyph-deleted document ?  true   <-- expected false
  w:sym document == literal-PUA document   ?  false    <-- expected true
```

A document that lost the glyph and a document that kept it project to the same string; the two legal spellings of the same glyph project to different strings. The gate is blind in the direction that matters and noisy in the direction that does not.

## Expected

- The projection resolves `<w:sym w:font="F" w:char="XXXX"/>` to `String.fromCharCode(0xXXXX)`, so the `w:sym` and literal-PUA spellings of one glyph compare equal and a genuine deletion compares unequal.
- `INV-RT-001` and `verification/ROADMAP.md:46` are restated over the widened element set, or state explicitly which element classes the invariant does *not* cover, so a green round trip means what it appears to mean.

## Notes on scope

`w:sym` is the case with a proven wrong answer, but it is one member of a class. The same projection also contributes nothing for `w:br`, `w:tab`, `w:cr`, `w:noBreakHyphen`, `w:pgNum`, `w:drawing`, `w:pict` and `w:object`. If the fix is a widened projection rather than a `w:sym` special case, it would be worth enumerating the element set in one place and testing it.

This is structurally the same failure as #647 and #648 — success reported while content is silently wrong — but at the verification layer rather than the output layer, which makes it the more consequential member of the family: it is the check that is supposed to catch the others.

## Why this is a library bug, not a consumer lesson

safe-docx **exports** `extractRoundTripComparisonText`, **gates its own safety check on it**, and makes a **formal claim** about it — `INV-RT-001`, with a Lean model in `verification/lean/LeanSpike/Spec.lean`. A consumer who reads the ROADMAP and trusts the invariant is behaving correctly. The invariant is what is too narrow.

## Existing issues checked

No issue in the repo (283, all states) mentions `w:sym`, symbol fonts, Wingdings, or private-use characters. #347 (closed) is adjacent but concerns *which projection* — accept vs reject — not which elements it contains. Not a duplicate.

stevenobiajulu:
## The bug

`extractRoundTripComparisonText` — the projection the accept-all / reject-all reconstruction safety gate compares — collected only `w:t` and `w:delText`. A `<w:sym w:font="Wingdings" w:char="F0A8"/>` contributed nothing to it, and that single gap failed in **both directions at once**.

**False negative, the serious one.** A document that *lost* a symbol and a document that still *had* it projected to the same string. The gate in `packages/docx-compare/src/baselines/atomizer/pipeline.ts` compares that string, reports `normalizedIdentical`, and the inplace comparison ships. `INV-RT-001` is stated over the same element set, so the formal invariant was satisfied by a document missing a glyph. This is the check that is supposed to catch the others.

**False positive.** OOXML has two legal spellings of one glyph — a literal private-use codepoint inside `w:t`, and `w:sym` — and Word renders them identically. A comparison engine may rewrite one into the other. The projection read the two spellings as different, so a purely notational rewrite read as mass character loss. Not hypothetical: a `w:t`-only walker reported 36 dropped checkbox glyphs across a redline where nothing had been dropped.

This is a library bug rather than a consumer's problem because safe-docx *exports* the projection, *gates its own safety check* on it, and makes a *formal claim* about it. A consumer who reads the ROADMAP and trusts `INV-RT-001` is behaving correctly. The invariant was what was too narrow.

## Reproduction

The issue's synthetic repro, run against `dist` at `origin/main` (`8035dce`), before the fix:

```
  projection of <w:sym w:char="F0A8"/>   : "Alpha  Bravo"
  projection of literal U+F0A8 in <w:t>  : "Alpha \uF0A8 Bravo"
  projection with the glyph DELETED      : "Alpha  Bravo"

  w:sym document == glyph-deleted document ?  true   <-- expected false
  w:sym document == literal-PUA document   ?  false    <-- expected true
```

After this change, same script, same fixtures:

```
  projection of <w:sym w:char="F0A8"/>   : "Alpha \uF0A8 Bravo"
  projection of literal U+F0A8 in <w:t>  : "Alpha \uF0A8 Bravo"
  projection with the glyph DELETED      : "Alpha  Bravo"

  w:sym document == glyph-deleted document ?  false   <-- expected false
  w:sym document == literal-PUA document   ?  true    <-- expected true
```

## What changed

- **`packages/docx-core/src/primitives/symbol_run_content.ts`** (new). `projectSymbolRun(element)` resolves a `w:sym` to `String.fromCharCode(@w:char)`. One place, so the element set is enumerated once. `@w:char` is `ST_ShortHexNumber` — `xsd:hexBinary` restricted to `length="2"` — so exactly four hex digits resolve and nothing else does. Both `CT_Sym` attributes are optional in the vendored schema, so a `w:sym` with no usable `@w:char` is not schema-invalid, just semantically unresolvable; it is still *content*, so it projects a framed identity rather than nothing.
- **`extractRoundTripComparisonText`** adds `sym` to its content branch, inside the existing TOC-PAGEREF suppression guard so a glyph in a suppressed page-number cache stays suppressed.
- **`extractTextContent` / `extractTextWithParagraphs`** get the same resolution through a shared document-order walk over `w:t` + `w:sym`. These are exported, are what the benchmark gates in `packages/docx-compare/src/benchmark/gates.ts` use, and are the surface the Lean model mirrors — leaving them blind would have left the same hole in a public API. The pre-existing "all live text first, then all deleted text" grouping is untouched, and there is no `w:delSym`, so output is byte-identical for any document containing no `w:sym`.
- **The Lean model moves with it.** `Tier2.OoxmlModel.Atom` gains a `sym` constructor carrying the resolved character, so `Tier2.RoundTripText.extractText` still mirrors the production projection definitionally instead of under-approximating it. It is its own constructor rather than a `text` atom precisely because it has no deleted-content counterpart: `reject`'s rename pass must leave it alone, which `renameAtom`'s catch-all already does. Both round-trip lemmas close unchanged.
- **Conformance registry** gains `[ECMA-PART1-17-3-3-30]` for `w:sym`, with the exclusions stated. **`INV-RT-001`** gains caveats naming the element set the projection counts and the classes it still does not.

## The font question

`@w:font` is deliberately **not** part of the projected identity, and this was the closest call in the change.

The argument for including it is real: `w:char="F0A8"` under Wingdings and under Symbol are different glyphs on the page, so a comparison engine that rewrote one to the other passes this gate.

Three reasons it is still excluded:

1. **This is a text projection, and it already excludes every other font signal.** A run's font lives on `w:rFonts`, which the projection does not see. Folding `w:sym/@w:font` in would make the same font signal visible for one spelling of a glyph and invisible for the other — incoherent, and worse than either consistent choice.
2. **Including it would defeat half the fix.** The literal-codepoint spelling carries its font on `w:rFonts`, not on the character, so there is nothing to compare it against. `w:sym` and literal spellings of one glyph would be unequal *by construction* — reintroducing exactly the false difference that produced the 36 phantom dropped checkboxes.
3. **A font-only change is a formatting change.** The repo checks formatting on a different surface with different semantics (`formattingFidelity.ts`, keyed per character on `w:rPr`).

The residual is filed, not buried: a comparison output that changed only `w:sym/@w:font` passes this gate, and it is not caught on the formatting side either — `formattingFidelity.ts` projects `w:t`/`w:delText` only, so a `w:sym` contributes neither a character nor a font key. That is #799, with its own reproduction.

### The sentinel for an unresolvable symbol

Worth flagging, because the obvious answer is wrong. A `w:sym` with no usable `@w:char` must still project *something* or the blindness returns. The first version used `U+FFFD` and called it "out of band." It is not: a document may author `U+FFFD` literally, and the gate then reads a broken `w:sym` and that text as the same content. It is now a framed token following the `__safe_docx_pageref__|` convention this file already uses for stable comparison identities, with a test pinning that the two do not compare equal.

## What is deferred, and why

**The run-list half — `packages/docx-core/src/primitives/text.ts` — is not in this PR.** It is #797.

`getParagraphRuns` drops a run whose only content is a `w:sym`, so the glyph is invisible to `getParagraphText`, to offset arithmetic, and to `replace_text`. The projection change there is one line; the rest is not:

- `getParagraphText` **defines** the raw-offset coordinate space that `replaceParagraphTextRange`, comment anchoring, footnote insertion, `buildCleanToRawOffsetMap` and every locator span are expressed in, and at least five sibling walkers compute that same length independently and would not move with it (`visibleLengthForEl`, the footnote-marker walk in `document_view.ts`, the comment-marker walk in `comments.ts`, `getParagraphRunVisibleLengths` in `docx-mcp/read_file.ts`, `splitRunAndInsertReference` in `footnotes.ts`).
- Two of the failure modes are the silent kind: an off-by-one split that corrupts text rather than throwing, and a symbol removed without a `w:del` record during a tracked edit.
- Two published identities are seeded from paragraph text and would change — the persisted `_bk_*` anchor names (which hash the paragraph's text *and its neighbours'*) and the client-held `content_fingerprint` / `portable_paragraph_ref`.
- **There is no `w:sym` anywhere in the corpus.** All ten fixtures under `tests/test_documents/` were checked: zero occurrences. CI would be green either way. That is a reason to isolate the change so it is bisectable, not a reason to fold it in.

**The rest of the gate's blindness is #798.** The same gate is still blind to a lost `w:tab`, `w:br`, `w:noBreakHyphen`, empty spacer paragraph, or leading indent. Two of those cannot be fixed by widening this projection at all — `normalizeText` erases them downstream — so they need a structural parity check rather than more elements, and the set as a whole is a behaviour change (the gate can only get stricter, i.e. more rebuild fallbacks) rather than a correctness fix. Different blast radius, different PR. The `w:sym` case is the one where the projection returns a *wrong answer in both directions*, which is why it goes alone.

## Tests

New: `packages/docx-compare/src/symbolCharacterProjection.test.ts` (6), `packages/docx-core/src/primitives/symbol_run_content.test.ts` (4). Fixtures are entirely synthetic and stay in the issue's Alpha/Bravo/Charlie vocabulary.

Both directions of the bug are pinned, plus the font decision, the sentinel collision, PAGEREF suppression, the four-hex-digit lexical space, and `extractTextWithParagraphs`.

**Negative controls, because the absence of them is why this survived:**

- `legacyTextOnlyProjection` — a `w:t`/`w:delText`-only walker kept *in the test*, asserted to still report the lost glyph and the kept glyph as equal. The test demonstrates the cause, not just the cure.
- `unchangedGate` — the same `compareTexts` comparison over a pair that really is unchanged must still report a match, so a green run is a result rather than a stuck assertion.
- A different `@w:char` must still compare unequal, so resolution is not returning a constant.
- A symbol-free mixed-revision document must project byte-identically, pinning that `extractTextWithParagraphs` did not move.

**Proof the tests can go red.** Reverting the two source files to `origin/main` and re-running:

```
 Test Files  1 failed (1)
      Tests  5 failed (5)
```

All of them, including `AssertionError: expected 'Alpha  Bravo' not to be 'Alpha  Bravo'`.

**Full suites:**

| suite | result |
|---|---|
| `@usejunior/docx-compare` | 59 files, 851 passed, 27 skipped |
| `@usejunior/docx-mcp` | 100 files, 981 passed |
| `test:baseline` | 43 files, 588 passed |
| `@usejunior/docx-core` | 129/131 files pass. Three failures, all reproduced unchanged at `origin/main` (`8035dce`): two LibreOffice `soffice` aborts, and `nvca-structural-regression.test.ts` asserting on captured atomizer console output. |

**Lean:** `lake build` completes 6621 jobs, no errors, no `sorry`. The axiom audit is set-identical to `expected-axioms.txt` — the `sym` constructor introduced no new axiom:

```
Classical.choice
LeanSpike.compareDocumentXml
LeanSpike.compareDocumentXml_output_preservation_friendly
LeanSpike.compareDocumentXml_output_text_roundtrip
Quot.sound
propext
```

`lint:workspaces` (0 errors), `check:conformance-citations`, `check:conformance-doc`, `check:conformance-explorer`, `check:invariants-doc`, `check:control-bytes`, `check:allure-labels`, `check:allure-filenames`, `check:allure-quality`, `check:cycles`, `check:ecma-376-coverage`, `check:capability-projection`, `check:lean-xml-checker-coverage`, `check:lean-comment-memory`, `check:tool-docs`, `check:mcpb-manifest`, `check:gemini-extension-manifest` all pass. (`check:spec-coverage-generation` fails identically on `origin/main` — pre-existing, untouched here.)

## Peer review

Reviewed by Codex under the dynamic-verification discipline: it built the repo, ran the focused and full suites, wrote and executed its own edge probes, and ran the mutation test. Verdict was **changes requested**. All four substantive findings are fixed in the second commit:

1. **The Lean model no longer described the shipped projection.** The most important finding, and correct: a caveat documents a mismatch without making the theorem describe the engine. `Atom` gained a `sym` constructor; both round-trip lemmas closed unchanged; the axiom inventory is identical.
2. **`@w:char` is `ST_ShortHexNumber`, not `ST_UcharHexNumber`.** Verified against the vendored schema. The resolver had cited the wrong type and accepted one to four digits, so `w:char="A"` was read as a character.
3. **The `U+FFFD` sentinel collided with authored `U+FFFD`.** Demonstrated with an executed probe. Replaced with a framed identity and pinned by a test.
4. **`verifiedBy` named a test that does not exercise `w:sym`.** Now names the two that do.

Its Lean recipe named `atomText`, `renameAtom` and `atomText_renameAtom` but not `FieldStructure.stepAtom` or the `ToJson Atom` encoder in `DifferentialHelpers.lean`, both of which match `Atom` exhaustively and needed a case; `XmlTripleChecker.lean` matches a different type and was unaffected. Everything it confirmed positively — no false differences under `w:del`/`w:moveFrom`/`w:ins`/`w:moveTo`, fields, or `premergeRuns`; traversal equivalence for `w:sym`-free nesting including text boxes; the tests being load-bearing — was reproduced here.

Fixes #793


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `8a7935bb1a5aaebc72473976803667b93587c4f5` (squash)
Built from: `main` @ `8a7935b`
Smoke: `npm run build && npm run lint:workspaces && npm run test:run` + a real-document drive of `projectSymbolRun` / `extractRoundTripComparisonText` / `extractTextWithParagraphs` / `compareDocuments`

**Steps**
- ✅ build (all workspaces)
- ✅ lint (`lint:workspaces`, exit 0)
- ✅ tests — 1354 passed, 1 failed, 2 expected-fail, 1 skipped (132 files); `docx-compare` 883 passed / 27 skipped (61 files). The one failure is **pre-existing** (details below).
- ✅ CI on `main` for `8a7935b`: **success**
- ✅ real-document drive: **8/8**

**Real-world fixture**
`~/.open-agreements/cache/nvca-investors-rights-agreement/source.docx` — the **NVCA Model Investors' Rights Agreement (10-1-2025)**, pulled by the open-agreements recipe downloader from `nvca.org` and SHA-256 verified against the recipe's declared `source_sha256` (`be8ad13f…bd08a`). Its `word/document.xml` carries **33 real `<w:sym w:font="Wingdings" w:char="F0A8"/>` ballot boxes** — the element under test, authored by NVCA.

**Before/after on those exact bytes** — this is the bug, measured:

| | pre-merge `d013a15` | merged `8a7935b` |
|---|---|---|
| `projectSymbolRun` exported from docx-core | ✗ `undefined` | ✓ `function`, resolves to U+F0A8 |
| ballot boxes visible to `extractRoundTripComparisonText` | **0 of 33** | **33 of 33** |
| projection length | 123,827 | 123,860 |
| drop **one** real `w:sym` → projection | **byte-identical** (123,827) | **123,859, differs** (delta 1) |
| same for `extractTextWithParagraphs` | identical | differs |

Before this change, 33 glyphs in a real NVCA form were invisible to the round-trip gate and losing one was undetectable. That is the whole of #793, on a real document.

**Other real-document checks**
- ✅ `w:sym` and the same glyph spelled as a literal private-use character inside `w:t` project **identically** (123,860 = 123,860) — the equivalence claim, not just a length change.
- ✅ End-to-end `compareDocuments` on the repo's real NVCA source/filled pair (`tests/test_documents/nvca-regression/`), `reconstructionMode: 'inplace'`: 3s, 147,116 bytes, `mode=inplace`, 107 `w:ins` / 292 `w:del`.
- ✅ INV-RT-001 over the **widened** projection on the real IRA: accept-all ≡ reject-all ≡ original, with all 33 glyphs present in every projection (123,860 chars each).

**Findings that are NOT this PR's fault, but should not be swallowed**

1. **`lean-build` is red — for a pre-existing reason.** It is not a required context, so it did not block. Its Lean steps all **passed** on this PR: `Build LeanSpike` ✅, `Audit Lean axiom allowlist` ✅, `Audit all Lean modules are zero-sorry` ✅, `Audit Lean checker coverage and stack-safety wiring` ✅ — the Lean model changes here compile and carry no `sorry`. What failed is step 15, `Complete NVCA selected-comment fixed-stack acceptance`, which is the same `nvca-structural-regression.test.ts > returns structured results for selected comments and nested-definition mutations` assertion at line 311 that fails locally and **reproduces on `8035dce`, `9c85a2a` and `d013a15`** — i.e. before any of #801/#802/#803.
2. **Consequence worth acting on:** because step 15 fails, steps 16–17 — **`Lean↔TS LCS differential harness (exhaustive)`** and **`Lean↔TS Tier 2-helper differential harness (extended)`** — were **skipped**. On a PR that edits `Tier2/RoundTripText.lean`, `Tier2/OoxmlModel.lean` and `DifferentialHelpers.lean` alongside their TS counterparts, the harness that checks Lean and TS agree never ran. The same skip has been happening on `main` since the 2026-08-10 Lean Build. Fixing that one assertion unblocks the differential harnesses.
3. `compareDocuments` on the IRA itself throws `OpaquePassthroughError: unsupported REF field instruction shape` — verified **identical on pre-merge `d013a15`**, so pre-existing, not a regression. The end-to-end leg was run on the NVCA source/filled pair instead.

Visual evidence: skipped — the change is a text-projection widening and is structurally invisible on a rendered page.
Log: `/tmp/automerge-smoke-802-20260811-122055.log` (local).

**Advisory gates audited**
- `Aggregate and post review` (usejunior-llm-gate): **Overall: PASS** (9 pass, 0 warn, 0 error, 8 skipped, 17 total), including the Lean-predicate-drift and canonical-emission-surface checks. No provider-failure markers.
- `codecov/patch`: 81.82%, 4 partials — advisory, not a required context. `2af8aba` closed the one genuinely reachable branch (`if (text)` in `collectLiveCharacterContent`); the rest are type-required `?? ''` guards a real parse cannot hit.
- `workspace-lint` failed once on `2af8aba` with `actions/checkout`: `server certificate verification failed` — a runner TLS fault, not a lint error. It also **skipped `workspace-test`**. Re-run, both green.

**Cleanup**
- ✅ remote branch `793-project-sym-glyphs-20260811` already deleted (auto-delete enabled); stale ref pruned
- ✅ removed worktree `.worktrees/793-sym-projection-20260811` (was clean)
- ✅ deleted local branch `793-project-sym-glyphs-20260811` (`-D`, squash-merge; PR confirmed `MERGED` first)

Issue #805, opened 2026-08-11
verification: the Tier 2-helper differential harness cannot exercise Atom.sym — #802's constructor is uncrossed by Lean↔TS


stevenobiajulu:
## Summary

The Tier 2-helper differential harness cannot exercise the `sym` constructor that #802 added to `Tier2.OoxmlModel.Atom`. Both sides of the wire block it, so re-enabling the harness (#804) verifies everything about the Lean↔TS correspondence **except** the part #802 changed.

This is a coverage gap, not a divergence. Nothing is known to be wrong. That is precisely the problem: nothing is known to be *right* either.

## Evidence

**1. The Lean encoder gained `sym`; the decoder did not.**

`verification/lean/DifferentialHelpers.lean` — `ToJson Atom` handles all six constructors, including the one #802 added:

```
| .sym s => Json.mkObj [("sym", toJson s)]
```

`FromJson Atom`, in the same file, matches only `text`, `delText`, `instrText`, `delInstrText` and `fldChar`, falling through to `.error s!"unknown Atom: {j.compress}"`. There is no `sym` case.

Feeding the compiled executable a `sym` atom confirms it:

```console
$ printf '%s' '{"cases":[{"doc":[{"body":[{"run":{"content":[{"text":"a"},{"sym":"✓"}]}}]}]}]}' \
    | ./verification/lean/.lake/build/bin/leanHelperDifferential
uncaught exception: case parse error: unknown Atom: {"sym":"✓"}
```

**2. The TypeScript harness has no `sym` variant either.**

In `packages/docx-core/src/integration/lean-differential-helpers.test.ts`, `WireAtom` is a five-way union — `text`, `delText`, `instrText`, `delInstrText`, `fldChar`. There is no `sym` member, `atomToXml` never emits a `<w:sym>` element, `atomTokens` has no `sym` branch (a `sym` atom would fall through its final `return` and be tokenized as `fc:undefined`), and the fast-check generator emits only `textAtom` (`'a' | 'b'`) and `delTextAtom` (`'x' | 'y'`).

So the harness cannot construct a `sym` case, and the Lean executable would reject one if it did.

## What is consequently unverified

#802's own commit message states the claims the harness would otherwise check:

- `atomText` projects `sym`.
- `reject`'s rename pass leaves `sym` untouched — the stated reason it is a distinct constructor rather than a `text` atom, since OOXML has no `w:delSym` and a symbol inside `w:del` is still spelled `w:sym`.
[...]

PR #785, opened 2026-08-04
fix(docx-compare): preserve machine-readable stdout

stevenobiajulu:
## Summary

- remove the unconditional `[DEBUG] atomizeTree` write to stdout
- add a regression test that holds `atomizeTree` silent on stdout
- preserve the existing consumer-compatibility warnings on stderr

## Root cause

`atomizeTree` called `console.log` unconditionally, so the diagnostic was interleaved with the `safe-docx compare` JSON result and broke machine consumers such as `jq`.

## Impact

Stdout once again belongs exclusively to the CLI's machine-readable comparison result. This is intentionally a narrow fix and does not introduce a logging framework or alter the result schema.

## Validation

- repository build and workspace lint
- full workspace test suite; five environment-restricted subprocess/LibreOffice tests rerun successfully with required permissions
- spec coverage and ECMA-376 conformance checks
- `safe-docx compare ... --certificate-format llm | jq -e .`
- `safe-docx compare ... --certificate-format full | jq -e .`
- focused atomizer regression: 64/64 passing

Fixes #783

Issue #783, opened 2026-08-04
fix(docx-compare): ungated [DEBUG] atomizeTree line on stdout corrupts machine-readable CLI output


stevenobiajulu:
## Problem

`packages/docx-compare/src/atomizer.ts:912` writes a debug line to **stdout** on every atomization, with no environment gate, log level, or opt-in:

```ts
console.log(
  `[DEBUG] atomizeTree: created ${rawAtoms.length} atoms, field-collapsed to ${fieldCollapsedAtoms.length}, merged to ${mergedAtoms.length}, word-split to ${wordSplitAtoms.length}, punct-merged to ${atoms.length}, ${state.emptyParagraphCount} empty paragraphs`
);
```

`safe-docx compare` prints its result as JSON on stdout, so the two are interleaved and the command's output is not parseable:

```
$ safe-docx compare original.docx revised.docx redline.docx --certificate cert.json | jq .
[DEBUG] atomizeTree: created 1734 atoms, field-collapsed to 1462, merged to 1332, word-split to 41621, punct-merged to 41621, 36 empty paragraphs
[DEBUG] atomizeTree: created 1696 atoms, field-collapsed to 1422, merged to 1115, word-split to 39824, punct-merged to 39824, 36 empty paragraphs
parse error: Invalid numeric literal at line 1, column 6
```

This is pre-existing, but it works directly against the goal #780 / #781 just shipped: the compare CLI now defaults to an LLM-oriented certificate projection precisely so a machine consumer can read the result. A consumer that pipes the command anywhere has to know to skip past leading noise to the first `{` — which is what the post-merge smoke for #781 had to do.

Note the contrast with the consumer-compatibility warnings emitted on the same code path (`[Consumer Compatibility] Warning: Orphaned bookmarkStart …`), which correctly go to stderr and cause no problem. `atomizer.ts:912` is the only ungated `console.log` in `packages/docx-compare/src`.

## Proposed capability

- Remove the line, or route it through stderr behind an explicit opt-in (env var or an existing diagnostics option) so it is off by default.
- Establish the invariant that library code never writes to stdout: stdout belongs to the CLI's machine-readable result, and diagnostics belong on stderr.

## Acceptance criteria

- `safe-docx compare … | jq .` parses with no preprocessing, for both `--certificate-format llm` and `full`.
- Nothing in `packages/docx-compare/src` writes to stdout on the default path; a test or lint rule holds the line.
[...]

PR #806, opened 2026-08-11
ci(verification): stop an unrelated failing test from silencing the Lean↔TS harnesses

### LINKED ISSUE #804: ci(verification): a failing acceptance test silently skips both Lean↔TS differential harnesses — unchecked since 2026-08-10
## Summary

In the `Lean Build` workflow, one failing acceptance test silently disables the two Lean↔TS differential harnesses — the only checks that verify the Lean model still describes the TypeScript engine.

The harnesses have not run on `main` since **2026-08-10**. The Lean steps themselves are green (`Build LeanSpike`, the axiom allowlist, and zero-sorry all pass), so the model compiles and is internally sound. Nothing has verified it still *describes the engine* for that entire window.

## Evidence

Run [31511870128](https://github.com/UseJunior/safe-docx/actions/runs/31511870128) (`main`, `8a7935b`), job `lean-build`, step conclusions read individually:

| # | Conclusion | Step |
|---|---|---|
| 13 | success | Near-limit comment payload memory acceptance |
| 14 | success | Real NVCA compared-only relationship-story mutations |
| 15 | **failure** | Complete NVCA selected-comment fixed-stack acceptance |
| 16 | **skipped** | Lean↔TS LCS differential harness (exhaustive) |
| 17 | **skipped** | Lean↔TS Tier 2-helper differential harness (extended) |

`lean-build` runs on `main` since the last green one (`30793720711`, 2026-08-03):

- `31363212774` — 2026-08-10, `8035dce`, failure — same step 15, same 16/17 skipped
- `31511870128` — 2026-08-11, `8a7935b`, failure — as above

## Why this matters more than a red check

`safe-docx`'s distinguishing claim is a machine-checked invariant (`INV-RT-001`, with a Lean model). The differential harnesses are what tie that model to the shipped engine. With them skipped, the Lean steps stay green while the correspondence goes unchecked.

**The sharp edge:** PR #802 merged on 2026-08-11 and edited `Tier2/RoundTripText.lean`, `Tier2/OoxmlModel.lean`, `Tier2/FieldStructure.lean` and `DifferentialHelpers.lean` **alongside their TypeScript counterparts** — adding a `sym` constructor to the model and widening the round-trip projection. The harness whose entire job is to check that Lean and TS agree never ran on it.

### (a) The wiring — a failing acceptance test can suppress a verification harness

Steps run with the default `if: success()`, so any earlier failure skips everything after it. Step 15 tests document-comparison behavior on an NVCA fixture; it has no bearing on whether the Lean↔TS correspondence holds. It should not be able to decide whether that correspondence gets checked.

**A naive `if: always()` on steps 16–17 would make this worse, not better.** Both harnesses gate themselves on the compiled Lean executable being present:

```ts
const exeExists = existsSync(LEAN_EXE);
const describeMaybe = exeExists ? describe : describe.skip;
```

An all-skipped vitest file exits `0`. So under `if: always()`, a run where `Build LeanSpike` failed would execute the harness steps, skip every case inside them, and report them **green** — a step that verifies nothing while claiming success. That is strictly harder to notice than the current skip.

The property to secure is therefore two-sided:

1. A failing *unrelated* step must not suppress a verification harness.
2. A harness must not be able to report success without actually running.

Proposed mechanism: gate steps 16–17 on `!cancelled()` **and** the explicit success of their genuine prerequisites (the Lean build and `npm ci`) via step `id`s, plus a strictness env var that makes the harnesses hard-fail rather than `describe.skip` when the executable is missing in CI. Reordering alone only inverts the problem; a separate job would re-pay the ~22-minute cold Lean build.

### (b) Step 15 — an assertion on a debug log that was deliberately deleted

`packages/docx-core/src/integration/nvca-structural-regression.test.ts` asserts:

```ts
expect(atomizerLogs.some((message) =>
  message.includes('word-split to ') && message.includes('punct-merged to '),
)).toBe(true);
```

Those strings no longer exist anywhere in the source tree. `8035dce` ("fix(docx-compare): preserve machine-readable stdout", #785, fixing #783) removed the `[DEBUG] atomizeTree: ...` line so CLI stdout stays valid machine-readable JSON — and added a regression test in `atomizer.test.ts` asserting `console.log` is **not** called during atomization.

The two tests are in direct contradiction: one requires the log, the other forbids it. That is why step 15 began failing on exactly the commit that removed the line.

The assertion's intent — confirming the word-split and punct-merge passes actually ran — is already covered more strongly and directly a few lines above by the exact atom-count fingerprint `expect(originalAtomCount).toBe(41_621)`, which is computed through the full pipeline (`splitTextIntoWords` defaults to `true`) and passes today.

## Not in scope here: branch protection

`lean-build` is not a required status context, which is why none of this blocked a merge. That is arguably the deeper problem, but making it required affects every open PR and all org seats, and interacts with the pending merge-queue decision. **Flagging only — deliberately not changing branch protection.**

## Follow-up already identified

The Tier 2-helper harness's `WireAtom` union is `text | delText | instrText | delInstrText | fldChar` — it has **no `sym` variant**, and its generator emits only `text`/`delText` atoms. So re-enabling the harness will not exercise the `sym` constructor #802 added. That is a coverage gap rather than a divergence, and will be filed separately.

stevenobiajulu:
Fixes #804.

One failing acceptance test was silently disabling the two Lean↔TS differential harnesses — the only checks that verify the Lean model still *describes* the TypeScript engine. They had not run on `main` since **2026-08-10**, a window that includes the merge of #802, which edited `Tier2/RoundTripText.lean`, `Tier2/OoxmlModel.lean`, `Tier2/FieldStructure.lean` and `DifferentialHelpers.lean` **alongside their TypeScript counterparts**. The harness whose entire job is to check those two agree never ran on it.

The Lean steps stayed green throughout — `Build LeanSpike`, the axiom allowlist and zero-sorry all pass — because they prove the model compiles and is internally sound. That property remains true no matter how far the engine drifts.

### (a) The wiring

Steps ran under the default `if: success()`, so step 15 (`Complete NVCA selected-comment fixed-stack acceptance`) skipped steps 16–17 after it. Step 15 tests document-comparison behavior on an NVCA fixture and has no bearing on the Lean↔TS correspondence.

- `id: lean_build` and `id: npm_ci` on the two genuine prerequisites.
- Both harness steps gated on `!cancelled() && steps.lean_build.conclusion == 'success' && steps.npm_ci.conclusion == 'success'`.
- `SAFE_DOCX_REQUIRE_LEAN_DIFFERENTIAL: '1'` on both, with a matching guard in each harness file.

**Why not `if: always()`.** Both harnesses self-gate on the compiled Lean executable (`describe.skip` when absent), and an all-skipped vitest file exits `0`. Under `always()`, a run whose `lake build` failed would execute these steps, skip every case, and report them **green** — the same defect wearing a passing checkmark. Demonstrated by moving the executable aside:

```console
$ npm run test:run -w @usejunior/docx-core -- src/integration/lean-differential-helpers.test.ts
 Test Files  1 passed (1)
      Tests  3 passed | 9 skipped (12)
# exit 0  ← a job summary shows this as success

$ SAFE_DOCX_REQUIRE_LEAN_DIFFERENTIAL=1 npm run test:run -w @usejunior/docx-core -- src/integration/lean-differential-helpers.test.ts
Error: [lean-differential-helpers] SAFE_DOCX_REQUIRE_LEAN_DIFFERENTIAL=1 but .../leanHelperDifferential is missing.
Refusing to skip: a skipped differential harness reports success without verifying anything.
 Test Files  1 failed (1)
# exit 1
```

The guard follows the pattern already used and CI-proven in this repo for `SAFE_DOCX_REQUIRE_LEAN_CHECKER` (`leanXmlVerifier.test.ts:102`).

`!cancelled()` rather than `always()` because it respects cancellation instead of launching a ten-minute exhaustive sweep on an abandoned or superseded run; the prerequisite clauses because a harness whose inputs were never built cannot be meaningful. Reordering ahead of the acceptance steps would merely invert the problem; a separate job would re-pay the ~22-minute cold Lean build.

### (b) Step 15 itself — fixed here, because the cause was unambiguous

`nvca-structural-regression.test.ts` asserted that the `[DEBUG] atomizeTree: ... word-split to N, punct-merged to M` line was emitted. Those strings exist nowhere in the source tree. `8035dce` (#785, fixing #783) deleted that line so CLI stdout stays valid machine-readable JSON — and added a unit regression in `atomizer.test.ts` asserting `console.log` is **not** called during atomization.

The two tests were in direct contradiction: one required the log, the other forbade it. That is why step 15 began failing on exactly the commit that removed the line.

Inverted rather than deleted, so the step now pins #785's stdout-silence property at full-document scale. The original intent is carried by the exact atom-count fingerprint (`expect(originalAtomCount).toBe(41_621)`) asserted just above, which is computed through the full pipeline.

## What the harnesses said once turned back on

Both pass, and both genuinely ran — not skipped:

| Harness | Result |
|---|---|
| `lean-differential-lcs` (`LEAN_DIFF_EXHAUSTIVE=1`, ~1.19M pairs) | 2/2 passed, 23.4s |
| `lean-differential-helpers` (extended sweep) | 12/12 passed, 26.9s |
| `nvca-structural-regression` (step 15, with Lean checker required) | 2/2 passed |

**No Lean↔TS divergence surfaced.** #802's Lean and TypeScript changes do not break the correspondence the harnesses check.

## ⚠️ But that pass does not cover #802 — filed as #805

The harnesses cannot exercise the `sym` constructor #802 added. Both sides of the wire block it:

- `WireAtom` in `lean-differential-helpers.test.ts` is a five-way union (`text`, `delText`, `instrText`, `delInstrText`, `fldChar`) with no `sym` member, and the generator emits only `text`/`delText`.
- `DifferentialHelpers.lean` encodes `.sym` in `ToJson Atom` but **never decodes it** in `FromJson Atom`.

```console
$ printf '%s' '{"cases":[{"doc":[{"body":[{"run":{"content":[{"text":"a"},{"sym":"✓"}]}}]}]}]}' \
    | ./verification/lean/.lake/build/bin/leanHelperDifferential
uncaught exception: case parse error: unknown Atom: {"sym":"✓"}
```

So the green sweep above says nothing about #802's constructor. This is a **coverage gap, not a divergence** — the harness rejects the case before either model is compared. Nothing is known to be wrong; nothing is known to be right either. Tracked in **#805** with a concrete fix plan. Nothing was loosened to make anything pass.

## Peer review

Codex, `--code` lane, dynamic with full execution trace — stored under `.peer-review/lean-harness-skip/`. Verdict **NEEDS-CHANGES**, three findings:

1. **High** — the harness is vacuous w.r.t. `sym`. Codex independently reproduced the `WireAtom` union, the missing `FromJson` case and the `unknown Atom` rejection, and argued it should be fixed *in this PR*.
2. **Medium** — the `always()` rationale comment was stale, since the guard added in the same change means `always()` would now *fail* rather than report green. **Correct catch; fixed in `797e90e`.**
3. Defect (b) — no blocking issue, but the comment overclaimed that the atom-count fingerprint covers each pass "more strongly and directly". **Fixed in `797e90e`.**

Codex also produced a failure-mode table confirming the condition behaves correctly across job failure, timeout, cancellation, prerequisite failure and runner loss, and found **no normal failure mode that silently yields a green job**. It confirmed GitHub documents `!cancelled()` as the recommended form here.

**Adjudication on finding 1 — deferred to #805, against Codex's preference.** Extending the harness means adding a Lean decoder case, a TS wire variant, real `<w:sym>` XML rendering matching the production `@w:char` contract, two token projections, seeded cases across five wrapper types, and generator changes. That is a decision about what the harness's documented "faithful subset" should contain, not a wiring fix — and it may itself surface a genuine divergence needing its own investigation. Blocking a narrow, provable wiring fix that protects *every future change* on an open-ended investigation would leave the wiring defect in place longer for no gain. The gap is pre-existing and is not made worse here; it becomes visible only because the harnesses run again.

## Flagged, deliberately not changed: `lean-build` is not a required status context

This is why none of the above blocked a merge, and it is arguably the deeper problem. Making it required affects every open PR and all org seats, and interacts with the pending merge-queue decision. **Left alone — that is a call for @stevenobiajulu, not this PR.**

## Verification

- `npm run lint:workspaces` — exit 0. (Note: a stale `dist/` produces phantom `projectSymbolRun` errors; run `TURBO_FORCE=1 npm run build` first.)
- Proof that a failing step 15 no longer suppresses 16–17: a deliberate-sabotage branch, run via `workflow_dispatch`, posted as a comment below.

### DIFF
--- a/.github/workflows/lean-build.yml
+++ b/.github/workflows/lean-build.yml
@@ -250,17 +250,23 @@ jobs:
       # NVCA acceptance step suppressed both harnesses across a window in which #802
       # edited the Lean model and its TypeScript counterparts together.
       #
-      # `always()` would be the WRONG relaxation. Both harnesses self-gate on the compiled
-      # Lean executable (`describe.skip` when it is absent) and an all-skipped vitest file
-      # exits 0 — so under `always()`, a run whose `lake build` failed would execute these
-      # steps, skip every case, and report them GREEN. That is the same defect wearing a
-      # passing checkmark, and strictly harder to notice than a skip.
+      # `!cancelled()` plus explicit prerequisite success, rather than `always()`, for two
+      # reasons. It respects cancellation — `always()` would keep launching a ~10-minute
+      # exhaustive sweep on a run an operator has already abandoned or that a newer push
+      # superseded. And it declines to run the harnesses at all when their own inputs were
+      # never built: after a failed `lake build` or `npm ci` there is nothing meaningful to
+      # compare, and a step that cannot be meaningful should not report at all. GitHub
+      # documents `!cancelled()` as the recommended form for steps that must survive an
+      # earlier failure.
       #
-      # `!cancelled()` plus explicit prerequisite success is the precise condition: run
-      # whenever the harnesses CAN be meaningful (their own inputs were built) and an
-      # operator has not cancelled the run, regardless of unrelated failures. Paired with
-      # SAFE_DOCX_REQUIRE_LEAN_DIFFERENTIAL below, which turns the self-skip into a hard
-      # failure, a green harness step now implies the harness actually ran.
+      # Historically `always()` carried a second, sharper hazard: both harnesses self-gate
+      # on the compiled Lean executable (`describe.skip` when absent) and an all-skipped
+      # vitest file exits 0, so a run whose `lake build` failed would have executed these
+      # steps, skipped every case, and reported them GREEN — the same defect wearing a
+      # passing checkmark. SAFE_DOCX_REQUIRE_LEAN_DIFFERENTIAL below now closes that
+      # independently, turning a missing executable into a hard failure, so it is defense
+      # in depth rather than the reason for the condition. Together: a green harness step
+      # implies the harness actually ran.
       #
       # Reordering ahead of the acceptance steps would only invert the problem (a harness
       # failure would then suppress the acceptance gates). A separate job would re-pay the
--- a/.github/workflows/lean-build.yml
+++ b/.github/workflows/lean-build.yml
@@ -139,6 +139,7 @@ jobs:
           exit 1
 
       - name: Build LeanSpike
+        id: lean_build
         working-directory: verification/lean
         run: |
           lake build
@@ -205,6 +206,7 @@ jobs:
           node-version: 20
           cache: npm
       - name: Install dependencies (npm ci with retry)
+        id: npm_ci
         run: |
           for attempt in 1 2 3; do
             echo "npm ci attempt ${attempt}/3"
@@ -236,9 +238,38 @@ jobs:
           SAFE_DOCX_REQUIRE_LEAN_CHECKER: '1'
         run: npm run test:run -w @usejunior/docx-core -- src/integration/nvca-structural-regression.test.ts
 
+      # The two differential harnesses below are the only checks that verify the Lean
+      # model still DESCRIBES the TypeScript engine. Every other Lean step above proves
+      # the model compiles and is internally sound, which stays true even as the engine
+      # drifts away from it.
+      #
+      # They therefore must not inherit the default `if: success()`. Under it, any earlier
+      # failure — including an acceptance test about document-comparison behavior, which
+      # says nothing about the Lean↔TS correspondence — silently skips them. That is
+      # exactly what happened between 2026-08-10 and 2026-08-11 (issue #804): a failing
+      # NVCA acceptance step suppressed both harnesses across a window in which #802
+      # edited the Lean model and its TypeScript counterparts together.
+      #
+      # `always()` would be the WRONG relaxation. Both harnesses self-gate on the compiled
+      # Lean executable (`describe.skip` when it is absent) and an all-skipped vitest file
+      # exits 0 — so under `always()`, a run whose `lake build` failed would execute these
+      # steps, skip every case, and report them GREEN. That is the same defect wearing a
+      # passing checkmark, and strictly harder to notice than a skip.
+      #
+      # `!cancelled()` plus explicit prerequisite success is the precise condition: run
+      # whenever the harnesses CAN be meaningful (their own inputs were built) and an
+      # operator has not cancelled the run, regardless of unrelated failures. Paired with
+      # SAFE_DOCX_REQUIRE_LEAN_DIFFERENTIAL below, which turns the self-skip into a hard
+      # failure, a green harness step now implies the harness actually ran.
+      #
+      # Reordering ahead of the acceptance steps would only invert the problem (a harness
+      # failure would then suppress the acceptance gates). A separate job would re-pay the
+      # ~22-minute cold Lean build for no added isolation.
       - name: Lean↔TS LCS differential harness (exhaustive)
+        if: ${{ !cancelled() && steps.lean_build.conclusion == 'success' && steps.npm_ci.conclusion == 'success' }}
         env:
           LEAN_DIFF_EXHAUSTIVE: '1'
+          SAFE_DOCX_REQUIRE_LEAN_DIFFERENTIAL: '1'
         run: npm run test:run -w @usejunior/docx-core -- src/integration/lean-differential-lcs.test.ts
 
       # Lean↔TS Tier 2-helper differential harness (Tier 2.5, second increment). `lake build`
@@ -248,6 +279,8 @@ jobs:
       # accept/reject/validate engine over a widened randomized sweep, reusing the Node setup
       # and `npm ci` above. `LEAN_DIFF_EXHAUSTIVE=1` widens this harness's random sweep too.
       - name: Lean↔TS Tier 2-helper differential harness (extended)
+        if: ${{ !cancelled() && steps.lean_build.conclusion == 'success' && steps.npm_ci.conclusion == 'success' }}
         env:
           LEAN_DIFF_EXHAUSTIVE: '1'
+          SAFE_DOCX_REQUIRE_LEAN_DIFFERENTIAL: '1'
         run: npm run test:run -w @usejunior/docx-core -- src/integration/lean-differential-helpers.test.ts
--- a/packages/docx-core/src/integration/lean-differential-helpers.test.ts
+++ b/packages/docx-core/src/integration/lean-differential-helpers.test.ts
@@ -493,6 +493,18 @@ const G5_DOC: WireDoc = [
 ];
 
 const exeExists = existsSync(LEAN_EXE);
+
+// See the matching note in lean-differential-lcs.test.ts. The skip is a developer
+// affordance; in CI a skipped harness exits 0 and is indistinguishable from a passing
+// one, which would let the Lean↔TS correspondence go unchecked behind a green step.
+// `SAFE_DOCX_REQUIRE_LEAN_DIFFERENTIAL=1` makes a missing executable fail loudly. See #804.
+if (!exeExists && process.env.SAFE_DOCX_REQUIRE_LEAN_DIFFERENTIAL === '1') {
+  throw new Error(
+    `[lean-differential-helpers] SAFE_DOCX_REQUIRE_LEAN_DIFFERENTIAL=1 but ${LEAN_EXE} is missing. ` +
+      `Refusing to skip: a skipped differential harness reports success without verifying ` +
+      `anything. Build it with: (cd verification/lean && lake build leanHelperDifferential)`,
+  );
+}
 if (!exeExists) {
   // eslint-disable-next-line no-console
   console.warn(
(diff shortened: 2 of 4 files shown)


stevenobiajulu:
## Proof: a failing step 15 no longer suppresses steps 16–17

Run on a throwaway branch (`proof/lean-harness-sabotage`) carrying this PR's fix plus one commit that deliberately breaks step 15, dispatched via `workflow_dispatch`:

**[Run 31518225561](https://github.com/UseJunior/safe-docx/actions/runs/31518225561)** — step conclusions read individually, not from the job summary:

| # | Before (run [31511870128](https://github.com/UseJunior/safe-docx/actions/runs/31511870128), `main` @ `8a7935b`) | After (this fix, step 15 sabotaged) |
|---|---|---|
| 13 Near-limit comment payload memory acceptance | success | success |
| 14 Real NVCA compared-only relationship-story mutations | success | success |
| 15 Complete NVCA selected-comment fixed-stack acceptance | **failure** | **failure** (deliberate) |
| 16 Lean↔TS LCS differential harness (exhaustive) | **skipped** | ✅ **success** |
| 17 Lean↔TS Tier 2-helper differential harness (extended) | **skipped** | ✅ **success** |

Same failing step 15, same step numbers. The harnesses now run.

The job still concludes `failure`, which is correct — step 15 genuinely failed. The point is that the failure is now *reported* rather than allowed to silence the verification harnesses behind it.

The sabotage branch has been deleted; nothing from it was merged.

---

### Note for @stevenobiajulu — this PR merged with `lean-build` still pending

Automerge squashed at `6ee0470` while `lean-build` was still running, because it is not a required status context. That is the flagged-but-deliberately-unchanged item from #804 demonstrating itself on the very PR that fixes the harness wiring: the harnesses now run, but nothing yet *waits* for them.

Still your call, and still interacting with the pending merge-queue decision — recorded here so it is not lost.

stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `6ee04703ed803986c675ac7aa3915c03c8d14825`
Built from: `main` @ `6ee0470`
Smoke: clean rebuild (`TURBO_FORCE=1 npm run build`) + lint + the real-document and differential paths this PR touches

**Steps**
- ✅ build (forced fresh — a stale `dist/` otherwise reports phantom `projectSymbolRun` errors)
- ✅ lint (`npm run lint:workspaces` — 0 errors)
- ✅ `nvca-structural-regression` — 2 passed, 22.2s, with `SAFE_DOCX_REQUIRE_LEAN_CHECKER=1`
- ✅ `lean-differential-lcs` — 2 passed, 22.1s, `LEAN_DIFF_EXHAUSTIVE=1 SAFE_DOCX_REQUIRE_LEAN_DIFFERENTIAL=1`
- ✅ `lean-differential-helpers` — 12 passed, 24.9s, same strict env
- ✅ `atomizer.test.ts` — 64 passed (the #785 regression that the old step-15 assertion contradicted)

**Real-world fixtures**
- `tests/test_documents/nvca-regression/source.docx` (real NVCA form, 140,221 bytes — not synthetic)
  - Driven through `compareDocuments` in `inplace` mode with the Lean XML verifier enabled, after `addSelectedCommentsPart` and a minimal tracked edit
  - Atomizes to exactly 41,621 atoms; `documentIntegrity.status = passed`, protocol v7, and stdout stays silent through the full comparison

## The deliverable: steps 16–17 execute on `main`

Post-merge run **[31519865828](https://github.com/UseJunior/safe-docx/actions/runs/31519865828)** (`main` @ `6ee0470`) — **conclusion `success`**, step conclusions read individually:

| # | Step | Before (`8a7935b`) | Now |
|---|---|---|---|
| 15 | Complete NVCA selected-comment fixed-stack acceptance | failure | ✅ success |
| 16 | Lean↔TS LCS differential harness (exhaustive) | **skipped** | ✅ **success** |
| 17 | Lean↔TS Tier 2-helper differential harness (extended) | **skipped** | ✅ **success** |

And they genuinely ran rather than passing vacuously — from the run log:

```
Lean↔TS LCS differential harness (exhaustive)          Tests  2 passed (2)   40.03s
Lean↔TS Tier 2-helper differential harness (extended)  Tests  9 passed | 3 skipped (12)   42.54s
```

This is the first green `lean-build` on `main` since 2026-08-03, and the first time the Lean↔TS correspondence has been checked on `main` since 2026-08-10.

**Two things stated plainly rather than glossed:**

1. Step 17's **3 skipped** are the LibreOffice oracle cases, which skip in CI because the `soffice` macro produces no marker file. Pre-existing and unrelated to this PR (#801 made those probes skip rather than abort). Locally, with LibreOffice available, the same file runs 12/12.
2. The green sweep still says **nothing about #802's `sym` constructor** — the harness cannot encode one and the Lean executable rejects one. That is **#805**, and it is a coverage gap, not a divergence. Nothing was loosened to make anything pass.

**Cleanup**
- ✅ remote branch `fix/lean-harness-skip` already gone (auto-delete enabled)
- ✅ removed worktree `.worktrees/lean-harness-skip`
- ✅ deleted local branches `fix/lean-harness-skip` and `proof/lean-harness-sabotage`; pruned origin

Log: `/tmp/automerge-smoke-806-1786470982.log` (local).

PR #811, opened 2026-08-11
fix(docx-compare): share one inplace default across every comparison front door and report the mode actually used

stevenobiajulu:
Closes #808

## What

Unifies the default comparison reconstruction mode to `inplace` across every front door and makes the `docx-comparison`/`safe-docx-compare` CLI report the mode actually used, not the mode requested.

| Front door | Before | After |
|---|---|---|
| `docx-comparison` / `safe-docx-compare` bins | `rebuild` (hardcoded literal, `compare-two.ts`) | shared `DEFAULT_RECONSTRUCTION_MODE` (`inplace`) |
| Library `compareDocuments` (option omitted) | `rebuild` (destructuring default, `pipeline.ts`) | shared `DEFAULT_RECONSTRUCTION_MODE` (`inplace`) |
| MCP `compare_documents` tool | `inplace` (own constant copy) | same constant, now re-exported from docx-compare |
| `safe-docx` / `safedocx compare` CLIs | `inplace` (own constant copy) | same constant, now re-exported from docx-compare |

## Why

Issue #649 reported the divergence and PR #669 fixed it — but only for the docx-mcp CLI. The published `@usejunior/docx-compare` bins kept the `rebuild` literal from their initial commit (`db8328ed`), whose only recorded rationale was a migration-era "default to rebuild initially for backward compatibility" note. Rebuild is the documented lower-fidelity path (#582: loses rsids, sectPr, fields, content controls), so callers reaching for the published bins silently got worse output than the MCP tool for the same input.

Separately, `compare-two.ts` emitted `mode: parsed.options.reconstructionMode` — the *requested* mode. The pipeline can silently fall back inplace→rebuild, so a caller could request inplace, receive rebuild output, and be told `"mode":"inplace"`. This contradicted the repo's own diagnosis protocol (`docs/comparison-failure-diagnosis.md` criterion 5: "the mode that ran, not the mode requested, is what gets recorded").

## How

- New `packages/docx-compare/src/comparison-defaults.ts` exports `DEFAULT_RECONSTRUCTION_MODE: ReconstructionMode = 'inplace'`, exported from the package index.
- `packages/docx-mcp/src/tools/comparison_defaults.ts` re-exports it instead of declaring its own copy — the constant cannot drift between packages again.
- The atomizer pipeline's destructuring default and the docx-compare CLI parser both consume the shared constant.
[...]

Issue #808, opened 2026-08-11
fix(docx-compare): published compare bins default to rebuild while MCP and safe-docx CLIs use inplace, and requested mode is reported as actual


stevenobiajulu:
## Summary

Two related defects in the shipped `@usejunior/docx-compare` CLI (`docx-comparison` / `safe-docx-compare` bins, published at 0.19.1), verified on `origin/main` @ `6ee04703`:

1. **Divergent default reconstruction mode across shipped front doors.** The docx-compare bins and the library API default to `rebuild`, while the MCP `compare_documents` tool and the `safe-docx`/`safedocx` CLIs default to `inplace`. Same input, same engine, different output fidelity depending on which front door you use. #582 documents that the rebuild path loses rsids, sectPr, fields, and content controls — so callers reaching for the published compare bins silently get the lower-fidelity result.
2. **The docx-compare CLI reports the mode *requested* as the mode *used*.** The atomizer pipeline can silently fall back inplace→rebuild, but `compare-two.ts` echoes the parsed option back. A caller can request `--mode inplace`, receive rebuild output, and be told `"mode":"inplace"`.

## Front-door inventory (verified on origin/main @ 6ee04703)

| Front door | Ships as | Default mode | Evidence |
|---|---|---|---|
| `docx-comparison` / `safe-docx-compare` bins | `@usejunior/docx-compare` 0.19.1, `package.json` bins → `dist/cli/index.js` | `rebuild` (hardcoded literal) | `packages/docx-compare/src/cli/compare-two.ts:50` |
| Library API (`compareDocuments` without `reconstructionMode`) | `@usejunior/docx-compare` 0.19.1 | `rebuild` (destructuring default) | `packages/docx-compare/src/baselines/atomizer/pipeline.ts:832` |
| MCP `compare_documents` tool | `@usejunior/docx-mcp` 0.19.1 | `inplace` via `DEFAULT_RECONSTRUCTION_MODE` | `packages/docx-mcp/src/tools/compare_documents.ts:127`, `packages/docx-mcp/src/tools/comparison_defaults.ts:9` |
| `safe-docx compare` / `safedocx compare` CLIs | `@usejunior/docx-mcp` 0.19.1, bins → `dist/cli.js` | `inplace` via `DEFAULT_RECONSTRUCTION_MODE` | `packages/docx-mcp/src/cli/commands/compare.ts:78` |

## Misreporting defect

[...]

PR #815, opened 2026-08-11
fix(docx-compare): pin the benchmark to rebuild and finish the #808 review follow-ups

stevenobiajulu:
Refs #808. Follow-up to #811, addressing the three verified findings from its Codex peer review (adjudication posted on #811):

1. **Benchmark strategy no longer drifts with the library default.** `packages/docx-compare/src/benchmark/runner.ts` was the one non-test caller omitting `reconstructionMode`, so #811 silently flipped the benchmark from rebuild to inplace. Benchmark scores exist for historical comparison (including against Aspose), so the runner now passes `reconstructionMode: 'rebuild'` explicitly, with a comment making any future switch to the shipped default a deliberate decision that must record requested/used mode and fallback reason in benchmark artifacts.
2. **Stale JSDoc.** `CompareResult.rebuildSafetyDiagnostics` still described rebuild as "the default mode"; reworded for the shared inplace default.
3. **The MCP handler is now pinned directly.** The `compare_documents` two-file tool test asserts `reconstruction_mode_requested === 'inplace'`, `reconstruction_mode_used === 'inplace'`, and no `fallback_reason` — previously the handler's default was pinned only transitively via constant parity.

Full pre-submit gate green locally (one unrelated LibreOffice headless PDF probe flaked under load and passes in isolation: `generation-package-structure.test.ts`, 6/6).

### DIFF
--- a/packages/docx-compare/src/benchmark/runner.ts
+++ b/packages/docx-compare/src/benchmark/runner.ts
@@ -75,7 +75,17 @@ async function produceRedline(
   asposeCliPath?: string,
 ): Promise<Buffer> {
   if (engine === 'atomizer') {
-    const result = await compareDocuments(originalBuffer, revisedBuffer, { engine: 'atomizer', author });
+    // Pin the benchmark to rebuild explicitly: benchmark scores predate the
+    // shared inplace front-door default (#808), and historical comparability
+    // requires the reconstruction strategy to stay fixed rather than drift
+    // with the library default. Switching the benchmark to measure the
+    // shipped default is a deliberate decision — record requested/used mode
+    // and fallback reason in the artifacts when making it.
+    const result = await compareDocuments(originalBuffer, revisedBuffer, {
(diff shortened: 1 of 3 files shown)


stevenobiajulu:
## Post-merge verification — PASS

Fresh clone updated to `main` @ `7db223c`, `docx-compare` + `docx-mcp` rebuilt.

MCP `compare_documents` over the real NVCA pair still requests and uses the shared default:

```
mcp-tool: {"req":"inplace","used":"inplace","fallback":null}
```

The newly pinned handler test passes against the merged code:

```
$ npm run test:run -w @usejunior/docx-mcp -- src/tools/compare_documents.test.ts
      Tests  10 passed (10)
```

And the benchmark runner now carries the explicit mode (`packages/docx-compare/src/benchmark/runner.ts:87` → `reconstructionMode: 'rebuild'`).

Issue #814, opened 2026-08-11
Construct redlines from a side-tagged tree instead of searching for a safe reconstruction


stevenobiajulu:
## Problem

`compareDocumentsAtomizerCore` does not construct a redline — it searches for one.

`packages/docx-compare/src/baselines/atomizer/pipeline.ts:1128-1231` defines four
atomization configurations, runs the **entire comparison** under each, and after each
candidate calls `evaluateRoundTripSafety`, which accept-alls and reject-alls the
serialized output and compares it against the two inputs. The first candidate that
passes wins; if all four fail, the result is discarded and `rebuild` runs as a hard
fallback.

Two consequences:

1. **The correctness oracle is load-bearing control flow.**
   `accept(output) ≡ revised ∧ reject(output) ≡ original` is the *definition* of a
   correct redline. Using it as a runtime filter to select among guesses means the
   engine is correct only to the extent that the checker is complete — every gap in
   the checker is a silently-shipped wrong redline, and every failure is a silent
   downgrade to lower-fidelity output.

2. **The emitter needs a cleanup crew behind it.** `inPlaceModifier.ts:126-145` runs
   `suppressNoOpChangePairs` over the tree it just built, to find and delete
   `<w:del>x</w:del><w:ins>x</w:ins>` pairs — content marked deleted and re-inserted,
   identically. Its own docstring (`inPlaceModifier-postprocess.ts:163-166`) calls
   these "false-positive changes". `suppressDuplicatedFormatChangesInTextReplacements`
   (#724) then compensates for a defect introduced by the coalescing passes themselves.

The root cause is representational. `atomizeTree` flattens OOXML into a flat
`ComparisonUnitAtom[]`, LCS runs over the flat list, and `documentReconstructor.ts`
(2330 lines) reconstitutes a tree from it. The tree invariants that make output valid
are destroyed by the flattening and have to be recovered afterwards — which is
precisely why they can only be *checked* afterwards.

Related: #542 (cross-run passes as candidate dead code), #469.

## Proposed approach

A side-tagged tree IR in which every node is tagged `both` / `original` / `revised`,
a `both` node carries **both** side representatives plus a scoped property delta, and
`project(tree, side)` is a total fold. The aligner's obligation is that each projection
is *isomorphic* to its input side (bijection, sibling order, containment, side-specific
content, opaque payload) — so round-trip correctness becomes a property of the
construction rather than a test on serialized output.

Full proposal, design, and staged tasks:
[...]

Issue #126, opened 2026-05-03
(Milestone) Remove comparison from the default finalization path


stevenobiajulu:
## Context

**This is a milestone gate, not a peer task.** It cannot close until all of #120, #121, #122, #123, #124, #125 are done and #119 has ratified the supported surface.

Today, tracked save in `packages/docx-mcp/src/tools/save.ts` still routes through `compareDocuments(...)` on a normalized baseline. The reconstructor wraps detected deltas in fresh `w:ins`/`w:del` carrying the session's author/date — that's what destroys pre-existing reviewer metadata. Removing this default is the **definition of done** for the architectural shift described in the umbrella (#118).

## Proposal

Switch the default save path to:

```
edits → tracked OOXML (#120) → validator (#121) → optional acceptAIEdits (#123)
```

Comparison-based reconstruction remains available as an explicit opt-in tool / flag for users who specifically want a redline against a baseline.

## Acceptance criteria

- End-to-end MCP test demonstrates a full edit → tracked-change → save flow with **no** invocation of `compareDocuments(...)` or `documentReconstructor`.
- Comparison rebuild is gated behind an explicit flag or a separate MCP tool — not the default save path.
- `packages/docx-mcp/src/tools/save.ts` no longer calls comparison code in the happy path.
- Documentation in `README.md` and the umbrella reflects the new default flow.

## Depends on

- #119 (surface definition)
- #120 (canonical tracked changes)
- #121 (validator)
- #122 (forbid untracked)
- #123 (selective accept)
- #124 (invariant corpus)
- #125 (mixed-author corpus)

## Related

- #118 (umbrella)
- Comparison package split (filed separately, follows this milestone)

stevenobiajulu:
Completed by merged PR #546, which removed comparison from the default save/finalization path and made write-time tracked markup canonical. Closing this stale milestone tracker.

PR #818, opened 2026-08-12
docs(docx-comparison): propose side-tagged tree redline construction

stevenobiajulu:
## Why

The atomizer does not construct a redline — it searches for one. `pipeline.ts:1128-1231` runs the full comparison under four atomization configurations, accept/rejects each serialized candidate against the two inputs, and keeps the first survivor; if none pass, it discards the result and falls back to `rebuild`.

That makes the correctness oracle load-bearing control flow. `accept(output) ≡ revised ∧ reject(output) ≡ original` is the *definition* of a correct redline — using it to select among guesses means the engine is correct only as far as the checker is complete, so every gap in the checker is a silently-shipped wrong redline and every failure is a silent fidelity downgrade. It also forces a cleanup crew behind the emitter: `suppressNoOpChangePairs` exists to delete `<w:del>x</w:del><w:ins>x</w:ins>` pairs that nothing should have emitted.

Root cause is representational: flattening OOXML to a flat atom list destroys the tree invariants that make output valid, so they can only be recovered — and therefore only checked — after the fact.

## What this PR contains

Documentation only. One OpenSpec change proposal, no code.

The proposal introduces a side-tagged tree where every node is tagged `both` / `original` / `revised`, a `both` node carries **both** side representatives plus a scoped property delta, and each projection must be *isomorphic* to its input side (bijection, sibling order, containment, side-specific content, opaque payload). Round-trip correctness then follows from the construction rather than from inspection.

**Scoped to stage A**: additive representation, isomorphism checks, and shadow-mode divergence evidence over the differential corpus. No behavior change, no deletions, every existing runtime check stays. The retry-ladder deletion (B/C) and the public `rebuild`-mode decision (D) are separate successor changes — nothing irreversible ships in the same change that introduces its own replacement.

## Peer review

Codex reviewed the first draft dynamically (it ran `openspec validate`, `check:spec-coverage`, targeted vitest suites, and `coverage:packages:check`) and overturned three assumptions. All corrected here, and recorded in the proposal rather than quietly dropped:

- **The coverage obligation was order-blind.** "Every input node appears exactly once" admits original `[A,B]`, revised `[B,A]`, tree `[both(B),both(A)]` — coverage holds, `project(original)` yields `[B,A]`. Replaced with a five-part isomorphism contract; the counterexample is now a required test case.
- **`rebuild` is a requested public output mode, not only a fallback.** Exposed on `CompareOptions.reconstructionMode`, the docx-compare CLI *defaults* to it (`cli/compare-two.ts:49`), and it selects a different base archive (`pipeline.ts:1239`). Removing it is a public breaking change → successor D.
- **`fail_on_rebuild_fallback` needed no migration** — it exists only on `save`, already deprecated and ignored (#126).

Also corrected: a `both` node needs two element representatives; cross-run unreachability is empirical rather than proven (the safety oracle checks bookmarks and field structure, not only text); PRESERVE needs construction invariants rather than payload transport (`preSplitInsProvenanceRuns` splits runs along provenance boundaries); and `coalesceMoveRangeMarkers` was missing from the post-processing inventory.

## Validation

- `openspec validate refactor-tagged-tree-redline-construction --strict` — valid
- `npm run check:spec-coverage` — 95/95 docx-comparison scenarios mapped (unchanged; this change adds no REMOVED delta)

## Note for review

Merging this is the approval gate — per `openspec/AGENTS.md`, implementation does not start until the proposal is approved. The substantive question is whether the staging (A → B → C → D) is the right split, and whether the isomorphism contract is now strong enough.

Refs #814

### DIFF
--- a/openspec/changes/refactor-tagged-tree-redline-construction/design.md
+++ b/openspec/changes/refactor-tagged-tree-redline-construction/design.md
@@ -1,0 +1,239 @@
+## Context
+
+The comparison engine's core loop is a generate-and-test search: `pipeline.ts`
+runs the full comparison under up to four atomization configurations, checks
+each serialized candidate by accept/reject round-trip, and keeps the first
+survivor — falling back to `rebuild` if none survive.
+
+That shape follows from the intermediate representation. `atomizeTree` flattens
+OOXML into a flat `ComparisonUnitAtom[]`; LCS runs over the flat list;
+`documentReconstructor` rebuilds a tree from atoms. Tree well-formedness — what
+makes a redline valid, and what `accept`/`reject` depend on — is not preserved
+by that pipeline, so it can only be checked at the end, and recovery from a
+failed check means retrying with different knobs.
+
+Two prior findings frame this change:
+
+- `reclassify-cross-run-rescue-as-residual` measured passes 3-4 as never
+  selected and documented them as a residual (#469, follow-up #542).
+- The Lean spike carries `compareDocumentXml_output_text_roundtrip` as a named
+  residual axiom precisely because `compareDocumentXml` is not modeled
+  definitionally — the engine's behavior has to be assumed.
+
+Both are symptoms of the same thing: correctness is asserted about the output
+instead of guaranteed by the construction.
+
+## Goals / Non-Goals
+
+**Goals (this change, stage A)**
+- A representation in which `accept`/`reject` correctness is a property of the
+  construction rather than a test on serialized output.
+- A projection contract precise enough to be falsifiable.
+- Shadow-mode divergence evidence over the differential corpus.
+
+**Non-Goals (this change)**
+- Any behavior change. Nothing is deleted, nothing is flipped, no runtime check
+  is retired.
+- **Effective formatting.** The existing detector (`format-detection.ts:299`)
+  and fidelity oracle (`formattingFidelity.ts:290`) inspect *direct* `w:rPr` /
+  `w:pPr` only, not formatting resolved through the style chain or
+  `docDefaults`. The shadow gate therefore cannot establish correctness for
+  inherited toggles, and `PropertyDelta` is scoped to direct properties.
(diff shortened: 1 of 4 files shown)


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `3af249ccb656560563c084c984bcaf5c75e612d2`
Verified against: `origin/main`

**Steps**
- ✅ `openspec validate refactor-tagged-tree-redline-construction --strict` — valid on merged main
- ✅ `openspec list` — change registered, 0/26 tasks
- ✅ `npm run build` — exit 0
- ✅ `npm run lint --workspaces --if-present` — exit 0
- ✅ `npm run test:run -w @usejunior/docx-compare` — 824 passed, 106 skipped, 62 files
- ✅ `npm run check:spec-coverage` — exit 0
- ✅ `npm run check:allure-labels` / `check:allure-quality` — 0 errors

Documentation-only PR, so there is no built artifact to drive end-to-end and no
document fixture to exercise — the meaningful smoke is that the change validates
against merged main and that the surrounding gates stay green on a checkout that
contains it. Recorded rather than substituted with a synthetic fixture.

**Correction found during post-merge verification**

One citation in the merged proposal was already stale when it landed. The
"Corrections carried from peer review" section says the docx-compare CLI
**defaults** to `rebuild`. That was true of the branch it was verified against,
but #811 (issue #808) had unified every comparison front door on `inplace`
before this merged, and #816 recorded that as a behavior change.

The substantive point is unaffected — `rebuild` is still a requested output
shape rather than merely a safety fallback, still accepted as `--mode rebuild`,
and still selects a different base archive — so successor D's scope does not
change. The default clause is withdrawn in the follow-up PR.

Separately: `CompareOptions.reconstructionMode` still advertises `Default:
'rebuild'` in its own doc comment (`packages/docx-compare/src/compare-types.ts:25`),
which is now wrong on a published API surface. Flagged, not fixed here.

Refs #814

PR #819, opened 2026-08-12
feat(docx-comparison): side-tagged comparison tree and projection contract

stevenobiajulu:
Implements stage A tasks 1.1-1.6 of `refactor-tagged-tree-redline-construction` (#818, #814).

## Why

`compareDocumentsAtomizerCore` does not construct a redline — it searches for one. `pipeline.ts:1128-1231` runs the full comparison under four atomization configurations, accept/rejects each serialized candidate against the two inputs, and keeps the first survivor; if none pass, it falls back to `rebuild`.

That makes the correctness oracle load-bearing control flow. `accept(output) ≡ revised ∧ reject(output) ≡ original` is the *definition* of a correct redline, so using it to pick among guesses means the engine is correct only as far as the checker is complete. The cause is representational: flattening OOXML to a flat atom list destroys the tree invariants that make output valid, so they can only be recovered — and therefore only checked — after serialization.

## What this adds

Additive only. No production caller, no behavior change, every existing runtime check untouched.

- `TaggedNode` = `both` | `original` | `revised`. A `both` node carries **two** element representatives, because matched is not identical — the same text under different run properties must stay one node, or the projection cannot know which side supplies the attributes and the difference has to be expressed as a delete plus an insert of the same text. That is the defect `suppressNoOpChangePairs` exists to clean up.
- `PropertyDelta` scoped by OOXML property level (run / paragraph mark / paragraph / row / cell / section), carrying **direct** snapshots. Effective (style-chain / `docDefaults`) formatting is an explicit non-goal — neither the detector nor the fidelity oracle resolves it, so a delta claiming to carry it could not be checked against anything.
- `project(tree, side)` as a total fold, and `verifyProjection` checking five obligations without serializing.

Coverage alone is not sufficient, and the gap is not academic: original `[A,B]` against revised `[B,A]` tagged `[both(B), both(A)]` has every input node present exactly once and still projects the original as `[B,A]`. OOXML text extraction is order-sensitive, so an order-blind obligation would certify a redline that reads back wrong. That counterexample is a test.

## Peer review

Codex reviewed the first cut dynamically — it ran the suite, `openspec validate`, `check:spec-coverage`, and its own adversarial probe — and **requested changes**. Four findings were real and are fixed in `61b83a8`:

1. **The opaque rule was vacuous.** A projected node with no children was read as standing for its whole subtree, then compared against the input by signature — but in the ordinary case those are *the same object*, so the check compared a thing with itself. A tree that forgot every descendant passed clean. Opacity is now explicit (`opaque: true`), and a non-opaque node must account for every input child. An empty child list is also what an incomplete construction looks like, so the two meanings cannot share a representation.
2. **Signatures were neither namespace-aware nor injective.** Identity was the lexical `tagName`, so elements from different namespaces sharing a prefix compared equal — and attributes were joined around delimiters, so `a="x b=y"` and `a="x" b="y"` collapsed to the same string. Now namespace URI + local name, with JSON-encoded attributes and text.
3. **P5 claimed byte identity it does not check.** The comparison normalizes attribute order, concatenates adjacent text, treats CDATA as text, and ignores comments/PIs. The spec now defines P5 as canonical equivalence and warns off modeling anything opaquely that depends on those distinctions.
4. **The round-trip claim overreached.** `verifyTaggedTree`'s doc comment said an empty result meant round-trip correctness held by construction — which this change's own spec explicitly forbids. IR projection fidelity is one of four layers; the runtime accept/reject checks are not made redundant by it.

Also from the review: property deltas are validated against their declared scope instead of being an unchecked bag, and the tests are falsifying rather than constructive — nine named regressions (forgotten descendants, wrong opaque payload, cross-namespace impostor, delimiter forgery, reorder-plus-edit, duplicated occurrence, promoted grandchild) plus three fast-check properties. **Every one of those cases passed the first cut.**

One earlier self-inflicted bug is also recorded in the code: sibling signatures were joined on a stray NUL in one of two places, which made identical subtrees compare unequal and silently disabled the ordering check. The separator is now a named constant on U+0001, which is not a legal XML character and so cannot be forged by content.

## Spec amendments in this PR

- P1 gains the opaque carve-out; P4 requires namespace-based identity; P5 is redefined as canonical equivalence, with a new scenario requiring unmodeled subtrees to declare themselves.
- The merged proposal's "the docx-compare CLI **defaults** to `rebuild`" bullet is withdrawn: #811 had already unified every front door on `inplace` before #818 merged, and #816 recorded it. The substantive point stands — `rebuild` is a requested output shape, not merely a fallback — so successor D's scope is unchanged.

## Validation

- `npm run build` — exit 0
- `npm run lint --workspaces --if-present` — exit 0
- `npm run test:run -w @usejunior/docx-compare` — 837 passed, 106 skipped (was 824 before this PR)
- `npm run check:spec-coverage` — exit 0
- `npm run check:allure-labels` / `check:allure-quality` — 0 errors
- `openspec validate refactor-tagged-tree-redline-construction --strict` — valid

## Known gap, flagged not fixed

`CompareOptions.reconstructionMode` still advertises `Default: 'rebuild'` in its doc comment (`packages/docx-compare/src/compare-types.ts:25`), which is wrong on a published API surface since #811. Out of scope here.

Refs #814

### DIFF
--- a/openspec/changes/refactor-tagged-tree-redline-construction/proposal.md
+++ b/openspec/changes/refactor-tagged-tree-redline-construction/proposal.md
@@ -84,12 +84,22 @@ Three claims in the first draft of this proposal were wrong and are recorded
 here rather than quietly dropped:
 
 - **`rebuild` is a first-class requested output mode, not only a fallback.**
-  `CompareOptions.reconstructionMode` exposes it (`compare-types.ts:20`), the
-  docx-compare CLI **defaults** to it (`cli/compare-two.ts:49`), and rebuild vs.
-  inplace select different base archives (`pipeline.ts:1239`). "Collapse to a
-  single path" is therefore wrong; the correct target is *one construction
-  algorithm per requested output shape*. Removing the mode is a separate public
-  breaking change (successor D), not a side effect of this refactor.
+  `CompareOptions.reconstructionMode` exposes it, both CLIs accept
+  `--mode rebuild`, and rebuild vs. inplace select different base archives
+  (`pipeline.ts:1239`). "Collapse to a single path" is therefore wrong; the
+  correct target is *one construction algorithm per requested output shape*.
+  Removing the mode is a separate public breaking change (successor D), not a
+  side effect of this refactor.
+
+  *Amended after merge:* this bullet originally said the docx-compare CLI
+  **defaults** to `rebuild`. That was true of the branch it was verified
+  against, but #811 (issue #808) had already unified every front door on
+  `inplace`, and #816 recorded that as a behavior change. The default clause is
+  withdrawn; the substantive point — `rebuild` is a requested output shape, not
+  merely a fallback — is unaffected, and successor D still owns any removal.
+  Note that `CompareOptions.reconstructionMode`'s own doc comment
+  (`compare-types.ts:25`) still advertises `Default: 'rebuild'` and is now
+  wrong on a published API surface; tracked separately, not fixed here.
 
 - **`fail_on_rebuild_fallback` needs no migration.** It exists only on `save`,
   where it is already deprecated and ignored (#126,
--- a/openspec/changes/refactor-tagged-tree-redline-construction/specs/docx-comparison/spec.md
+++ b/openspec/changes/refactor-tagged-tree-redline-construction/specs/docx-comparison/spec.md
@@ -49,15 +49,28 @@ contract for each side `s`:
 
 - **P1 bijection**: every node of input side `s` corresponds to exactly one tree
   occurrence tagged `both` or `s`, and every such occurrence corresponds to
-  exactly one node of input side `s`;
+  exactly one node of input side `s`. An explicitly opaque subtree counts as a
+  single atomic input unit and its descendants are not separately accounted;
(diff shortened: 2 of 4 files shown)


stevenobiajulu:
## Post-merge smoke passed\n\nVerified merge commit `e511bbe` from a worktree pinned to `origin/main`.\n\n- Real fixture: `~/.open-agreements/cache/nvca-investors-rights-agreement/source.docx`\n- Tagged every `w:body` element against itself: **14,193 elements**, `verifyTaggedTree(...)` returned **0 violations**.\n- Mutation probe: dropping one real tagged body child produced **13,426 violations**; swapping two real tagged body children produced **2 violations**. This confirms the verifier is live rather than vacuous.\n- Temporary smoke test was deleted after execution; the worktree has no smoke-test residue.\n- `npm run test:run -w @usejunior/docx-compare`: **841 passed, 106 skipped** (exit 0).\n- `npx openspec validate refactor-tagged-tree-redline-construction --strict`: exit 0.\n\nVisual evidence intentionally omitted: this is an unrendered data-structure contract with no production caller, so a screenshot would not provide useful evidence.

PR #825, opened 2026-08-12
feat(docx-mcp): warn when an inserted run departs from the document's formatting convention

### LINKED ISSUE #687: docx-mcp: no formatting-convention check for inserted runs — a new defined term can land off-convention and pass every gate

stevenobiajulu:
Closes #687.

## The defect class

An edit can be well-formed and still visibly wrong. A replacement that introduces an inline defined term in plain runs, inside an agreement that sets every other parenthetical definition in bold, opens in Word, extracts byte-identical text, and passes `validate_document`. Nothing in the pipeline compares an inserted run against the document's own distribution of the same construct, so the defect is invisible to well-formedness checks (`validate_document`) and to revision-markup checks (`validate_ai_revisions`) alike. It is caught, if at all, by a human reading the output.

Both of #687's prerequisites were already on `main` and unused for this purpose: `extractEffectiveRunFormatting` (#684, PR #691) and the success-path `warnings` channel (#686, PR #701).

## What this adds

`packages/docx-mcp/src/tools/formatting_convention.ts` — a structural convention check that runs on the preview document `preflightAiRevisionMutation` already parses, and returns a warning on the channel #701 built.

Two construct classes:

1. **Inline defined terms** — a quoted term inside a parenthetical carrying definitional phrasing: `(the "X")`, `(each, a "X")`, `(collectively, the "X" and the "Y")`.
2. **Proviso keywords** — `provided` / `however` introduced by a semicolon.

For each, the modal `(bold, italic, underline)` tuple is computed over every comparable instance in the document *as it stood before the edit*, and an inserted run that diverges from the mode is reported.

Five properties the design holds to:

- **Structural, never textual.** Text locates instances; only run properties judge them.
- **Effective, not declared, formatting.** Resolution goes through `extractEffectiveRunFormatting`, so a run inheriting bold italic from a character style is not flagged. That is the whole reason #684 was a prerequisite.
- **Advisory.** The check cannot block an edit and degrades to silence on any unexpected document shape. It rides on a round trip already paid for — 150–195 ms per AI-attributed mutation, measured on #687.
- **Silent without a convention.** Fewer than 5 comparable instances, or a modal tuple below 80% of them, produces nothing rather than a guess. Thresholds are the ones settled in the #687 design comment, evaluated over the population of comparable constructs rather than the document at large.
- **This mutation's insertions, not the session's.** "Inserted" is decided by differencing the pre-mutation document against the preview, using the same multiset-consumption technique `splitIntroducedDiagnostics` already uses in `ai_revision_guard.ts`. A `w:ins` an earlier edit left behind is part of the document now: it keeps its vote on the convention and is not attributed to this edit.

The trigger is gated on the inserted text — the check runs only when the text being inserted itself contains a construct, and is skipped entirely otherwise. Settled on #687 (2026-07-29): cost does not constrain the trigger, so it is chosen for precision.

### Threshold justification

`N >= 5` comparable instances with `>= 80%` dominance, from the #687 design comment. Five is the smallest population where a single outlier (1 of 5 = 20%) still leaves the mode exactly at the bar, so one anomalous instance cannot by itself manufacture a convention; below that, two instances agreeing is coincidence. 80% is high enough that a document with two competing house styles produces silence — two distinct tuples cannot each hold 80%, so a tie can never pass — and low enough to survive the handful of exceptions every real agreement carries.

### One measurement that changed the design

The construct span covers the term *between* the quote marks, not the marks themselves. Measured across the two committed corpus agreements, the opening and closing quotes almost always sit in the neighbouring runs — `[" (the "", plain]["Fund", bold]["") is made on", plain]` is the typical shape — so a span including the marks reads as three disagreeing runs and the instance is discarded as incomparable. With the span including quote marks, **1 of 26** and **2 of 43** instances were comparable and neither document yielded a convention at all. Narrowed to the term, the same documents resolve cleanly:

| Corpus document (already committed, public-source) | Defined terms | Provisos |
|---|---|---|
| `tests/test_documents/nvca-coi-regression/source.docx` | 25 of 26 bold (96%) | 11 of 13 underlined (85%) |
| `tests/test_documents/redline/ILPA-Model-Limited-Parnership-Agreement-Deal-By-Deal_v1.docx` | 41 of 42 bold (98%) | 30 of 30 underlined (100%) |

Every number above is read off the documents at runtime by the test, not written down.

## Negative control

A convention checker that never fires is indistinguishable from one that always passes. The suite ships paired controls throughout: each silence assertion has a minimally-different sibling that fires on the same fixture, differing only in the seeded formatting, the instance count, the author attribution, or the gate text.

Six mutations were run against the finished suite. Each turns it red:

| Mutation | Tests failed |
|---|---|
| `checkFormattingConvention` returns `[]` unconditionally | 12 |
| Declared `w:rPr` substituted for `extractEffectiveRunFormatting` | 2 |
| Both thresholds removed | 2 |
| Baseline differencing removed (every AI `w:ins` treated as this mutation's) | 1 |
| Only the first inserted run judged instead of every one | 1 |
| Inter-quote grammar removed | 1 |

All fixtures are synthesised from scratch — generic parties, generic obligations, no borrowed prose. The two corpus documents are used read-only, as evidence that the matcher finds real conventions in the wild.

## Plumbing

- `replace_text` — passes its inserted text; the existing `warnings` field carries the finding.
- `insert_paragraph` — gains the optional `warnings` field its siblings received in #701. Without it the check would have been computed and dropped on this path.
- `batch_edit` — preflights the whole sequence at once, so each warning is attributed back to the step whose inserted text carries the construct, and reported through the per-step `{ step_id, warning }` channel it already exposes.
- `AiRevisionPreflightResult` gains `conventionWarnings`, the structured form of the *same* findings already in `warnings` — for callers that need attribution. Reading both double-reports; this is stated on the type.

No tool input schema changed, so `docs/tool-reference.generated.md` is unchanged and `check:tool-docs` is a no-op.

## Gates (observed)

| Gate | Result |
|---|---|
| `npm ci` | clean |
| `npm run build` | pass |
| `npm run test` | pass — docx-mcp **1006 passed / 1 skipped** across 103 files; docx-core 1341 passed, 2 expected fail, 15 skipped; docx-compare 849 passed / 106 skipped; odf-core 140; google-docs-core 116 / 35 skipped; suite total green |
| `npm run lint:workspaces` | pass — 0 errors, 10 pre-existing warnings on untouched files |
| `npm run check:tool-docs` | pass (no regeneration needed) |
| `npm run preflight:ci` | pass — spec coverage, Allure label/filename/quality, conformance citations, ECMA-376 coverage, capability projection, site build + 101 internal links, 713 files with no import cycles |
| New tests | 22, all passing |

## Peer review

Reviewed with Codex on the pushed diff. `safe-docx` is public, so no code-disclosure objection applies. Verdict: **REQUEST-CHANGES**, four MAJOR and three MINOR findings. All seven are fixed in this branch:

| Finding | Severity | Disposition |
|---|---|---|
| Prior same-author insertions reported as this edit's doing | MAJOR | Fixed — baseline/preview multiset differencing; the convention is now computed from the pre-mutation document and prior insertions keep their vote. Test: `does not blame this edit for an off-convention insertion an earlier edit left behind`, both directions. |
| A construct reduced to its largest run let a divergent fragment through | MAJOR | Fixed — every text-bearing run overlapping the construct is resolved and judged; heterogeneous instances no longer vote on the convention. Test: `reports a divergent fragment of a term split across runs`, with the term split 4/4 so no largest-overlap rule could see the second half. |
| `batch_edit` silently omitted the feature | MAJOR | Fixed — wired with per-step attribution through its existing `{ step_id, warning }` channel. Test asserts the warning lands on the step that caused it and not the sibling step. |
| Arbitrary prose between two quoted strings joined the population | MAJOR | Fixed — every inter-quote segment must match a conjunction grammar. `(the "Buyer" shall notify the "Seller")` now matches nothing. |
| Nested and unbalanced parentheses treated as valid enclosures | MINOR | Fixed — depth-tracking scanner replaces the regex; `((the "Term")`, `(the "Term" (as amended))`, and both unbalanced forms are rejected. |
| Curly single quotes conflict with apostrophes | MINOR | Fixed by removal — `’` is also the apostrophe, so `(the ‘Company’s Assets’)` closed at the possessive. Double quotes only, with the reason recorded at the constant. |
| Silence tests weak against mutation | MINOR | Fixed — every silence assertion is paired with a positive control on the same fixture, and the corpus test now asserts a dominant convention is actually found rather than only that the match count is nonzero. |

Codex separately confirmed as sound: no `lastIndex` leakage on the module-level global regexes (a regression test now pins this), no catastrophic backtracking in the lead-in grammar, `dominantRun`/`overlappingRuns` offsets agreeing with `getParagraphRuns` coordinates including complex fields, the check being unable to block or throw, convention warnings correctly discarded on the blocked path, and no double-reporting in the single-edit warning merge.

## Scope

Left alone deliberately: `batch_edit` still drops `preflightAiRevisionMutation`'s validation warnings (a #686 residue, unrelated to this check — only the convention findings are routed there); no tool description or input schema changed; the check is not wired into `format_layout`, `clear_formatting` or the footnote/comment tools, none of which insert prose runs.

### DIFF
--- a/packages/docx-mcp/src/tools/ai_revision_guard.ts
+++ b/packages/docx-mcp/src/tools/ai_revision_guard.ts
@@ -6,6 +6,7 @@ import {
   type RevisionIdState,
 } from '@usejunior/docx-core';
 import { type DocxSession } from '../session/manager.js';
+import { checkFormattingConvention, type ConventionWarning } from './formatting_convention.js';
 import { err, type ToolResponse } from './types.js';
 
 function cloneRevisionIdState(state: RevisionIdState): RevisionIdState {
@@ -111,20 +112,30 @@ export function formatAiRevisionWarning(d: AiRevisionDiagnostic): string {
  * `blocked` is the historical `ToolResponse | null` contract: non-null means
  * validation rejected the mutation and the caller must return that response
  * without applying the edit. `warnings` carries the non-blocking diagnostics
- * that used to be structurally dropped on the success path (issue #686);
+ * that used to be structurally dropped on the success path (issue #686), plus
+ * any advisory findings computed off the same preview document (the
+ * formatting-convention check, issue #687);
  * callers should surface them on their success response as an optional
  * `warnings?: string[]` field. When `blocked` is set, `warnings` is empty —
  * the blocking response already carries the full diagnostics
  * (errors + warnings) in its `diagnostics` field.
+ *
+ * `conventionWarnings` is the *same* findings as the convention entries in
+ * `warnings`, in structured form — not an addition to them. A caller renders
+ * `warnings` and ignores this, or (like batch_edit, which must attribute a
+ * finding back to the step that produced it) reads this and ignores those.
+ * Reading both double-reports.
  */
 export type AiRevisionPreflightResult = {
   blocked: ToolResponse | null;
   warnings: string[];
+  conventionWarnings: ConventionWarning[];
 };
 
 const PREFLIGHT_PROCEED: AiRevisionPreflightResult = Object.freeze({
   blocked: null,
   warnings: [],
+  conventionWarnings: [],
 });
 
 const sessionBaselineDiagnostics = new WeakMap<DocxSession, Promise<Map<string, number>>>();
@@ -148,20 +159,59 @@ export function getAiRevisionBaseline(session: DocxSession): Promise<Map<string,
   return promise;
 }
 
+/**
+ * Non-validation advisories the caller wants computed off the same preview
+ * round trip. `insertedText` is the text the mutation adds; supplying it opts
+ * the mutation into the formatting-convention check (#687), which is skipped
+ * outright when the inserted text carries no construct the check knows about.
+ *
+ * It rides here rather than on its own document load because the preview load
+ * + mutate + validate round trip is already paid for on every AI-attributed
+ * edit (150–195 ms measured on #687); a full run scan adds 1–2 ms on top.
+ */
+export type AiRevisionPreflightAdvisories = {
+  insertedText?: string;
+};
+
 export async function preflightAiRevisionMutation(
   session: DocxSession,
   ctx: RevisionContext | undefined,
   mutatePreview: (doc: DocxDocument, ctx: RevisionContext | undefined) => Promise<void> | void,
   touched?: AiRevisionValidationTouchedContext,
+  advisories?: AiRevisionPreflightAdvisories,
 ): Promise<AiRevisionPreflightResult> {
   if (!session.aiAuthor) return PREFLIGHT_PROCEED;
 
   const snapshot = await session.doc.toBuffer({ cleanBookmarks: false });
+  // Two loads of the same bytes: `previewDoc` is mutated below, `baselineDoc`
+  // stays as the document was before this mutation so the convention check can
+  // difference them. Loading it from the same snapshot (rather than reusing
+  // `session.doc`) keeps both sides on identical run boundaries, so the
+  // difference reflects the edit and not a serialization artefact. It is only
+  // paid for when a caller actually requests the advisory.
   const previewDoc = await DocxDocument.load(snapshot.buffer);
+  const baselineDoc = advisories?.insertedText
+    ? await DocxDocument.load(snapshot.buffer)
+    : null;
   await mutatePreview(previewDoc, cloneRevisionContext(ctx));
   const validation = await previewDoc.validateAiRevisions(session.aiAuthor, touched);
+  // Advisory only, and never on the blocked path: a blocked mutation is not
+  // applied, so its formatting is not a finding the caller can act on.
+  const conventionWarnings =
+    advisories?.insertedText && baselineDoc
+      ? checkFormattingConvention(previewDoc, {
+          insertedText: advisories.insertedText,
+          aiAuthor: session.aiAuthor,
+          baselineDoc,
+        })
+      : [];
+  const conventionMessages = conventionWarnings.map((w) => w.message);
   if (validation.errors.length === 0) {
-    return { blocked: null, warnings: validation.warnings.map(formatAiRevisionWarning) };
+    return {
+      blocked: null,
+      warnings: [...validation.warnings.map(formatAiRevisionWarning), ...conventionMessages],
+      conventionWarnings,
+    };
   }
 
   // AI-attributed errors always fail. Unattributable errors (field structure,
@@ -186,7 +236,11 @@ export async function preflightAiRevisionMutation(
     // merge them into the surfaced warnings exactly as the failure path does.
     return {
       blocked: null,
-      warnings: [...validation.warnings, ...demoted].map(formatAiRevisionWarning),
+      warnings: [
+        ...[...validation.warnings, ...demoted].map(formatAiRevisionWarning),
+        ...conventionMessages,
+      ],
+      conventionWarnings,
     };
   }
 
@@ -196,5 +250,6 @@ export async function preflightAiRevisionMutation(
       warnings: [...validation.warnings, ...demoted],
     }),
     warnings: [],
+    conventionWarnings: [],
   };
 }
--- a/packages/docx-mcp/src/tools/batch_edit.ts
+++ b/packages/docx-mcp/src/tools/batch_edit.ts
@@ -14,6 +14,7 @@ import { replaceText, stripSearchTags } from './replace_text.js';
 import { insertParagraph } from './insert_paragraph.js';
 import { resolveSessionForTool } from './session_resolution.js';
 import { preflightAiRevisionMutation } from './ai_revision_guard.js';
+import { insertedConstructKeys, type ConventionWarning } from './formatting_convention.js';
 
 const REPLACE_TEXT_FIELDS = new Set([
   'target_paragraph_id',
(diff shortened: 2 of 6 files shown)


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `760a0d77fbb696aea421b6cf98b7bdb57a2bc684`
Built from: `main` @ `760a0d7` (clean worktree pinned to `origin/main`, not the primary checkout)
Smoke: `npm ci && npm run build && npm run test` + a real-DOCX end-to-end drive of the merged check

**Steps**
- ✅ build (`npm run build`, exit 0 — `packages/docx-mcp/dist/tools/formatting_convention.js` emitted)
- ✅ tests — docx-mcp **1006 passed / 1 skipped** (103 files); docx-core 1341 passed / 2 expected-fail / 15 skipped; docx-compare 849 / 106 skipped; odf-core 140; google-docs-core 116 / 35 skipped; mcpb 3; test-narrative 43; allure-factory 9
- ✅ post-merge `main` CI: `workspace-test (20)` and `workspace-test (22)` both success
- ✅ end-to-end: `open_document` → `insert_paragraph` → convention warning on the success channel

**Real-world fixtures**

Driven against the committed corpus document, not a synthetic fixture:

- `tests/test_documents/nvca-coi-regression/source.docx` (147,622 bytes, 228 paragraphs, copied to a working path and opened through the real session manager)
  - Convention the merged code independently derived from the document: inline defined terms **25 of 26 → bold** (96%); proviso keywords **11 of 13 → underline** (85%). These reproduce the PR body's measured figures exactly, computed from the document at smoke time rather than restated.
  - **Case 1 — off-convention insertion.** `insert_paragraph` added `Interest accrues on each late payment (each, a "Fee Item") monthly.` at a real body paragraph. Edit applied (`success=true` — the check never blocks) and one warning surfaced:
    ```
    FORMATTING_CONVENTION_DIVERGENCE: inserted inline defined term "Fee Item" is
    bold=false, italic=true, underline=false, but 25 of 26 (96%) of this document's
    inline defined terms are bold=true, italic=false, underline=false
    (construct=inline_defined_term)
    ```
  - **Case 2 — negative control.** The same tool, same anchor, same session, text carrying no construct (`The parties shall cooperate in good faith to effect the foregoing.`): edit applied, **zero** convention warnings. The silence is the gate, not an absent code path.

**Advisory-gate audit** (a green check is not a clean verdict): `Aggregate and post review` returns in 0s and dispatches to a separate repo, so its verdict text was read rather than its status. Genuinely healthy — **Overall: PASS (7 pass, 0 warn, 0 error, 10 skipped, 17 total)**, no `Unable to verify` / `provider failed` markers. Two of its passes are worth recording: *Unit-test quality (avoid tautological / change-detector tests)* and *Library stays general (no downstream-domain leakage)*.

**Note, not a blocker:** Codecov reports **91.03% patch coverage**, 19 lines uncovered (17 in `formatting_convention.ts`, 2 partials in `batch_edit.ts`) — mostly partial branches. Worth a glance if the uncovered branches are the degrade-to-silence paths, which are exactly the ones that fail quietly.

Visual evidence: skipped — the change adds an advisory warning channel and does not alter rendered document output.
Prod: ✅ Vercel deployment for `760a0d7` reports success (`Deployment has completed`).
Log: `/tmp/automerge-smoke-825-20260812-184148.log` (local).

**Cleanup**
- ✅ remote branch `687-formatting-convention-20260812` already gone (auto-delete enabled)
- ✅ removed worktree `.worktrees/687-formatting-convention-20260812`
- ✅ deleted local branch, pruned `origin`

PR #835, opened 2026-08-14
fix(docx-render-verifier): make PDF text binding pagination- and story-aware

### LINKED ISSUE #833: renderer verifier: make PDF text binding pagination- and story-aware
## Problem

`verifyRenderedMarkup()` requires normalized `pdftotext` output to equal one
caller-supplied logical markup string. For a paginated Word document,
`pdftotext` includes repeated headers, repeated footers, page numbers, and text
ordered according to renderer-created page positions. A logical DOCX body/story
projection cannot predict that sequence without independently reproducing
LibreOffice pagination.

Consequently, a render can have independently calibrated visible blue
insertions and red deletions, yet receive a generic failure solely because the
logical expected text omits pagination-derived repetitions.

## Minimal reproduction

Create a synthetic multi-page tracked DOCX with:

- one repeated header;
- one footer page-number field;
- one visible insertion and one visible deletion in the body.

Supply the independently derived logical tracked-text projection. LibreOffice's
PDF contains the expected body text plus repeated header/footer/page-number
material, so exact normalized equality fails even though both configured colour
bands exceed the by-author control.

## Expected invariant

- Keep text binding independent from the redline generator.
- Separate logical content binding from renderer-created pagination artifacts.
- Do not obtain a pass by simply dropping text verification.
- Report text-binding failure separately from colour visibility; do not label a
  render `insufficient-contrast` when calibrated blue and red evidence is
  present and only text binding failed.
- Add single-page, multi-page, repeated-header/footer, page-field, and reordered
  text-box fixtures.

Possible designs include story-scoped subsequence/multiset evidence with
explicit occurrence bounds, or a separately rendered clean/control operand
whose pagination artifacts can be compared independently. The issue should
select an invariant before implementation rather than prescribing one heuristic.

## Privacy

Use only a generated multi-page fixture. Do not commit private corpus documents,
text, paths, hashes, logos, or screenshots.

stevenobiajulu:
Fixes #833

## Problem

`verifyRenderedMarkup()` required normalized `pdftotext` output to equal one caller-supplied logical markup string. For a paginated Word document, LibreOffice's PDF contains repeated headers, repeated footers, page-number field renderings, and text in renderer-created page/float order. A logical DOCX body/story projection cannot predict that sequence without reimplementing pagination, so a render with correctly calibrated blue insertions and red deletions failed anyway — and the failure was mislabeled `insufficient-contrast`.

## Selected invariant: story-scoped per-page reservation, then exact multiset coverage

1. **Pagination-first reservation** — extracted PDF text is split into rendered pages (form feeds). On each page, pagination-owned material is reserved FIRST and MAXIMALLY: each referenced header/footer story token up to its occurrence count in those stories (cached field results excluded), and numeric page-number renderings up to the header/footer PAGE-family field count per page plus a whole-document budget for body-story fields. Numeric reservation is value-constrained to the rendered page-number range (`w:pgNumType` start offsets included) and prioritizes each page's own number and the NUMPAGES total.
2. **Completeness lower bound** — every token of the caller's logical projection must be covered by the tokens remaining after reservation. Because reservation is maximal, repeated header/footer vocabulary can never substitute for genuinely missing logical content.
3. **Zero residue** — every remaining token beyond the projection fails, so duplicated or hallucinated rendered content also fails.

The pagination profile is derived from the rendered DOCX package (via the existing `referencedRenderedStories()` resolver, now story-kind-aware) and the rasterized page count of the configured render — never from the caller's projection and never from any Safe DOCX generator — so the package remains an independent verifier.

**Why it can still fail on genuinely missing content:** a dropped logical token is not covered by post-reservation remainders (verified by fake-harness regressions and a real-render negative control), a body token that only header repetition could supply is still reported missing (header-masking regression), an out-of-range numeral such as an amount is never absorbed by a PAGE field, and any unattributable residue fails outright. If the rendered package is unreadable, reservation degrades to zero — strict multiset equality — rather than to leniency.

**Documented limitations (strict direction or separately covered):** reading order is outside the automated verdict (review PNGs support optional human placement review; no automated placement comparison); header/footer story text is pagination-owned and not itself bound (header revision visibility is covered by the pixel/revision-markup path); a bare in-range integer in body text may be attributed to pagination, which can only cause a failure, never a false pass.

## Separate failure reporting

- `RenderVerdict.textBinding` is new structured evidence (`matched`, `pageCount`, `missingTokenSample`, `unexplainedTokenSample`); `markupTextMatchesPdf` keeps its meaning for existing consumers (docx-release-verifier).
- `revisionVisibility` is now classified purely from pixel + revision-markup evidence. A text-binding failure produces its own reason string and never relabels calibrated colour evidence as `insufficient-contrast`.

## Fixtures

New generated fixtures (all synthetic, leak-scanned, registered in `fixtures/index.json`): `single-page-body.xml`, `multi-page-body.xml` (explicit page break), `repeated-header.xml`, `page-field-footer.xml` (PAGE field), `reordered-text-box.xml`. Tests drive the existing fake `RendererTools` harness (extended with a simulated page count) — no real binaries needed in CI. No private corpus content is included anywhere.

## Peer review + real-render validation

Codex peer review (dynamic, with executed probes) returned REQUEST-CHANGES on the first-cut design; all four findings were fixed in the follow-up commit — see the adjudication comment. Real-binary validation (`soffice`/`pdftotext`/`pdftoppm`/`magick`) on a synthetic 2-page tracked DOCX with repeated header and `Page <PAGE>` footer: correct projection passes (`textBinding.matched: true`, `pageCount: 2`, blue 29 / red 30 vs by-author control 0/0, `revisionVisibility: "visible"`); a corrupted projection fails with the text-binding reason and `missingTokenSample: ["phantom-token-that-was-never-rendered"]` while `revisionVisibility` stays `"visible"`.

### DIFF
--- a/packages/docx-render-verifier/README.md
+++ b/packages/docx-render-verifier/README.md
@@ -5,14 +5,27 @@ no dependency on Safe DOCX generators, mutators, comparison engines, or IR.
 
 `verifyRenderedMarkup` renders a disposable configured LibreOffice profile
 (blue/underlined insertions, red/struck deletions) and a same-input by-author
-control. It compares `pdftotext` output to caller-supplied independent markup
+control. It binds `pdftotext` output to caller-supplied independent markup
 text, measures broad blue/red pixel bands after downsampling, and writes only
 selected PDF-page PNGs to the requested output path. Required missing tools are
 reported as `not_run`, not pass.
 
+Text binding is story-scoped and pagination-aware rather than exact equality:
+every token of the caller's logical projection must appear in the PDF text at
+least as often as in the projection, and every extra PDF token must be
+attributable to renderer-created pagination artifacts with explicit occurrence
+bounds — referenced header/footer story tokens up to `pageCount x` their story
+occurrence, and numeric page-number residue only when a PAGE-family field
+exists, capped at `pageCount x` the field count. Missing logical content or
+unattributable residue still fails; reading order is deliberately left to the
+image-review domain. The structured outcome is reported in
+`verdict.textBinding` separately from colour visibility.
+
 If Writer displays configured insertions but suppresses deletions, verification
 fails with `revisionVisibility: "hidden-deletions"` rather than reporting only
-a generic colour-contrast failure.
+a generic colour-contrast failure. A text-binding failure keeps
+`revisionVisibility` truthful to the pixel and revision-markup evidence; it is
+never relabelled `insufficient-contrast`.
 
 An optional render transform receives only a copied input and a disposable
 workspace. The verifier hashes the authoritative DOCX before and after and
--- a/packages/docx-render-verifier/README.md
+++ b/packages/docx-render-verifier/README.md
@@ -10,16 +10,20 @@ text, measures broad blue/red pixel bands after downsampling, and writes only
 selected PDF-page PNGs to the requested output path. Required missing tools are
 reported as `not_run`, not pass.
 
-Text binding is story-scoped and pagination-aware rather than exact equality:
-every token of the caller's logical projection must appear in the PDF text at
-least as often as in the projection, and every extra PDF token must be
-attributable to renderer-created pagination artifacts with explicit occurrence
-bounds — referenced header/footer story tokens up to `pageCount x` their story
-occurrence, and numeric page-number residue only when a PAGE-family field
-exists, capped at `pageCount x` the field count. Missing logical content or
-unattributable residue still fails; reading order is deliberately left to the
-image-review domain. The structured outcome is reported in
-`verdict.textBinding` separately from colour visibility.
+Text binding is story-scoped and pagination-aware rather than exact equality.
+The extracted PDF text is split into rendered pages; on each page,
+pagination-owned material is reserved first and maximally — referenced
+header/footer story tokens up to their story occurrence count, and numeric
+page-number renderings (values within the rendered page-number range) up to
+the header/footer PAGE-family field count per page plus a whole-document
+budget for body-story fields. The caller's logical projection must then be
+covered by the remaining tokens, and any leftover token fails. Because
+reservation is maximal, repeated header vocabulary can never substitute for
+missing logical content, and duplicated or hallucinated rendered text is
+never absorbed. Reading order is deliberately outside this automated verdict:
+the emitted review PNGs support optional human placement review, and no
+automated placement comparison is performed. The structured outcome is
+reported in `verdict.textBinding` separately from colour visibility.
 
 If Writer displays configured insertions but suppresses deletions, verification
 fails with `revisionVisibility: "hidden-deletions"` rather than reporting only
--- a/packages/docx-render-verifier/fixtures/index.json
+++ b/packages/docx-render-verifier/fixtures/index.json
@@ -11,6 +11,11 @@
     { "id": "color-control", "path": "color-control.xml", "purpose": "renderer colour-control text" },
     { "id": "corrupt-zip", "path": "corrupt.docx", "purpose": "unreadable package negative control" },
     { "id": "header-wrap", "path": "header-wrap.xml", "purpose": "minimized de-identified header-wrap shape" },
-    { "id": "signature-table", "path": "signature-table.xml", "purpose": "minimized de-identified table layout shape" }
+    { "id": "signature-table", "path": "signature-table.xml", "purpose": "minimized de-identified table layout shape" },
+    { "id": "single-page-body", "path": "single-page-body.xml", "purpose": "single-page tracked body for exact text binding" },
+    { "id": "multi-page-body", "path": "multi-page-body.xml", "purpose": "explicit page-break body spanning two rendered pages" },
+    { "id": "repeated-header", "path": "repeated-header.xml", "purpose": "header story repeated once per rendered page" },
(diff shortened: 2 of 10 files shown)


stevenobiajulu:
## Codex peer review: verdict + adjudication

Codex CLI reviewed the initial commit (6ff6edb) dynamically — it read every affected file, ran the package and consumer test suites (exit 0), built the package, and executed adversarial probes against the built `dist/render.js`. **VERDICT: REQUEST-CHANGES**, with four findings. Adjudication:

| # | Finding | Adjudication | Resolution (commit 2e22a8b) |
|---|---|---|---|
| 1 | Header/footer text could satisfy a missing body token: completeness compared expected counts against the *global* PDF multiset, so a repeated header token substituted for a dropped body occurrence (probe: expected `Body Shared`, PDF `Shared Body / Shared` → matched) | **Valid** — reproduced against the built dist; it contradicted the JSDoc completeness claim | Redesigned to per-page **maximal reservation**: pagination-owned tokens are reserved first on every rendered page; the projection must be covered by what remains. The probe now fails with `missingTokenSample: ["Shared"]`. Regression test: "fails when repeated header vocabulary would otherwise mask a missing body token". |
| 2 | A PAGE-family field licensed arbitrary all-digit residue (probe: `999` accepted as an "amount" because one PAGE field existed) | **Valid** | Numeric reservation is now value-constrained to the rendered page-number range (`1..pageCount + pgNumType start − 1`) and prioritized to each page's own number + NUMPAGES. `999` now fails as unexplained. Regression tests: "does not let a PAGE field launder a meaningful out-of-range number", "rejects numeric residue outside the rendered page-number range". |
| 3 | A body PAGE field received a per-page allowance (`pageCount × pageFieldCount`) though it renders once | **Valid** | Field counts are now story-kind-aware: header/footer fields grant a per-page budget; body-story fields grant a whole-document budget. Regression test: "grants a body PAGE field one rendered occurrence rather than one per page". |
| 4 | Docs claimed order/placement was covered by "the image-review domain", but `reviewPages()` only emits PNGs — no automated placement comparison | **Valid (documentation)** | JSDoc + README now state that order insensitivity is an accepted limitation of the automated verdict and that review PNGs support optional *human* placement review. Regression test documents the accepted insertion/deletion order-swap pass. |

Additional fix found during real-render revalidation of the redesign (not in the review): the footer's cached `<w:fldSimple>` result (`1`) double-reserved with the numeric page-field budget and consumed a legitimate body token `1` ("Synthetic clause 1 …") — a false FAIL on real LibreOffice output. Cached field results are now excluded from the story quota (`hasFieldResultAncestor`), since the rendered value replaces them.

Verified-holding claims from the review (unchanged): zero-allowance strict fallback for unreadable packages, truthful `revisionVisibility` under text mismatch, and docx-release-verifier compatibility (still gates on `configuredContrastPassed === true && markupTextMatchesPdf === true`).

Post-fix evidence: 29/29 package tests pass; Codex's probes re-run against the rebuilt dist give `matched:false` for findings 1–2 and the documented pass for the order swap; full pre-submit gate (build, lint:workspaces, test:run, spec-coverage, conformance-citations, conformance-doc) green; real-render pass/fail controls re-confirmed (details in the PR body).

stevenobiajulu:
## Post-merge real-render smoke (merged main `bfae8a8`)

Built merged main in the worktree, generated a synthetic 2-page tracked DOCX (60 numbered body clauses, one tracked insertion, one tracked deletion, a repeated default header, and a `Page <PAGE>` footer field), and ran the actual verifier end-to-end with real binaries (`/opt/homebrew/bin/soffice` LibreOffice 25.8, `pdftotext`, `pdftoppm`, `magick` — all present, nothing faked).

**Positive control** — correct logical projection (body text incl. both revisions, headers/footers/page numbers omitted as pagination-owned):

```json
{
  "status": "pass",
  "reason": null,
  "markupTextMatchesPdf": true,
  "textBinding": {
    "matched": true,
    "pageCount": 2,
    "missingTokenSample": [],
    "unexplainedTokenSample": []
  },
  "configured": {
    "sampledPixels": 17384,
    "bluePixels": 29,
    "redPixels": 30
  },
  "byAuthorControl": {
    "sampledPixels": 17384,
    "bluePixels": 0,
    "redPixels": 0
  },
  "configuredContrastPassed": true,
  "revisionVisibility": "visible"
}
```

Text binding passes across both pages (repeated header + per-page page numbers fully reserved), and both colour bands exceed the by-author control (blue 29 / red 30 vs 0 / 0).

**Negative control** — the same projection with one phantom token appended:

```json
{
  "status": "fail",
  "reason": "PDF text binding failed: expected logical markup content is missing from the rendered PDF.",
  "markupTextMatchesPdf": false,
  "textBinding": {
    "matched": false,
    "pageCount": 2,
    "missingTokenSample": [
      "phantom-token-that-was-never-rendered"
    ],
    "unexplainedTokenSample": []
  },
  "configured": {
    "sampledPixels": 17384,
    "bluePixels": 29,
    "redPixels": 30
  },
  "byAuthorControl": {
    "sampledPixels": 17384,
    "bluePixels": 0,
    "redPixels": 0
  },
  "configuredContrastPassed": true,
  "revisionVisibility": "visible"
}
```

It fails with the text-binding reason and names the missing token, while `revisionVisibility` stays truthful at `"visible"` and `configuredContrastPassed` stays `true` — the mislabeling as `insufficient-contrast` that issue #833 reported is gone.

Smoke inputs were generated synthetic content only; no private corpus material was used.

Issue #839, opened 2026-08-14
fix(docx-compare): result metadata names deleted engines and reconstruction modes


stevenobiajulu:
## Problem

The sole surviving comparison spine returns public metadata naming deleted implementations:

```ts
engine: 'atomizer',
reconstructionModeRequested: 'inplace',
reconstructionModeUsed: 'inplace',
```

`CompareResult`, exported from `packages/docx-compare/src/index.ts`, still admits `engine: 'wmlcomparer' | 'atomizer'` and documents fallback fields as explaining rejection in favor of "validated legacy output." Unsafe tagged publication now throws, and neither the legacy engine nor inplace/rebuild reconstruction modes remain.

Callers using these fields for telemetry or branching receive false statements about how the document was produced.

## Acceptance criteria

- Inventory every exported result field tied to the deleted engine, reconstruction modes, and legacy fallback.
- Remove obsolete fields in the intended breaking release or replace them with truthful tagged-spine concepts.
- Do not return `atomizer`, `wmlcomparer`, `inplace`, `rebuild`, or legacy-fallback metadata for the sole surviving implementation.
- Add a public API test covering actual `compareDocuments` output and exported types.
- Document the migration for callers that consumed these fields.

This issue supersedes the original rebuild-mode decision in #839 because the modes were removed before the decision issue was resolved.

Premise verified against `origin/main` at `99109405` on 2026-08-21.

stevenobiajulu:
Read-only research pass (worktree off `origin/main` @ `8cbd8ae`). Evidence and a recommendation for a later human decision — the gating precondition ("tagged serialization supports original- and revised-side package skeletons") is **not met today**, so nothing here authorizes executing an option change now.

Two premise corrections to #814, verified on current main:

- **The docx-compare CLI no longer defaults to `rebuild`.** PR #811 (`f76739a`) unified every front door on `DEFAULT_RECONSTRUCTION_MODE = 'inplace'` (`packages/docx-compare/src/comparison-defaults.ts:17`; CLI at `packages/docx-compare/src/cli/compare-two.ts:66`). The `cli/compare-two.ts:49` citation in #814 is stale (the openspec proposal already withdraws it).
- **But the flip is unreleased.** The published `@usejunior/docx-compare` 0.19.1 on npm still carries the old rebuild default for the library and bins; the change sits in CHANGELOG "Unreleased" (`CHANGELOG.md:5-18`) with an explicit keep-old-behavior migration line.
[...]

Issue #838, opened 2026-08-14
tagged-tree successor: delete legacy reconstruction after rollback window


stevenobiajulu:
Follow-up to #814 and the tagged-tree production-default-flip issue.\n\nAfter the production default has completed its rollback window, remove legacy reconstruction, suppression, and coalescing machinery one concern per PR.\n\nKeep field, move-range, multi-author, and formatting regressions as permanent behavior contracts. Do not combine the default flip and legacy deletion in one change.\n\nThis is successor C from the staged tagged-tree plan.

stevenobiajulu:
Audit re-triage: leaving this issue closed rather than reopening a deletion task that has already shipped. The problem is now evidence/remediation, not whether to perform the deletion.

Successors:

- #919 repairs and executes the durable rollback recipe.
- #920 records the skipped between-phase soak gate and replacement evidence.
- #918 corrects the tautological standalone-assembler evidence.

The original requirement not to combine the authority flip and deletion was not honored; #920 is the explicit record for that gap.

Issue #837, opened 2026-08-14
tagged-tree successor: gate the production default flip on full shadow equivalence


stevenobiajulu:
Follow-up to #814.

Make tagged-tree redline construction authoritative only after all blocking offline-corpus divergences and the accept/reject, structural, formatting, move, field, multi-author, ancillary-story, and cross-reader gates pass.

Requirements:

- preserve the legacy output path for an explicit rollback window;
- publish the measured corpus-equivalence evidence;
- do not delete legacy reconstruction in this issue; and
- keep the new path out of the ordinary comparison pipeline until every blocking gate passes.

This is successor B from the staged tagged-tree plan.

stevenobiajulu:
Post-migration audit update: #918 consolidates the two Phase 6 evidence gaps relevant to this parent issue. On current `main`, the standalone observer compares one buffer with itself and its corpus suite is absent from the real-corpus CI command. #917 separately tracks the fail-open/vacuous differential harness.

This issue remains open as the parent record because its requested equivalence and rollback-window gates were not all evidenced before the flip/deletion. Follow-ups: #917, #918, #919, and #920.

stevenobiajulu:
Closing. The tagged-tree default flip happened in #841 and the legacy spine was deleted in #898/#907. As the 2026-08-21 comment records, not every gate here was evidenced before the flip. The four follow-ups that repaired that record are all closed: #917 (#930, fail-closed differential corpus), #918 (#929, removed false shadow evidence), #919 (#933, executable rollback from main) and #920 (#934, postmortem for the skipped soak). Nothing is left for this parent record to track.

Issue #836, opened 2026-08-14
tagged-tree offline evaluator: preserve direct paragraph formatting in tracked serialization


stevenobiajulu:
Stage-A offline comparison for #814 preserves accept/reject text on a synthetic paragraph replacement, but its serialized tagged tree does not match the legacy candidate's direct-formatting projection.

A committed de-identified synthetic regression changes paragraph justification and run emphasis. Both accept-all and reject-all text projections are exact, but the formatting-fidelity score remains below 1. The same divergence class blocks opaque corpus fixture `cd2f69960d5f13cc6292a138` with score `0.6287170885149017`.

Completion gate:

- the synthetic regression reaches formatting fidelity 1;
- the opaque fixture reaches formatting fidelity 1;
- accept/reject projections remain exact; and
- the legacy pipeline stays authoritative until the separate default-flip gate is met.

This issue records one formatting defect only. It does not authorize a production-default change or runtime telemetry.

stevenobiajulu:
Fixed before the tagged-tree default flip (#841). The synthetic regression (justification center→right, bold→italic, text change) now reaches formatting fidelity 1 with exact accept/reject projections, pinned in `packages/docx-compare/src/tagged/taggedTreeShadow.test.ts` ("proves direct paragraph and run formatting against source projections"; re-run on main @ fb0bd3af, passes). The serializer repair and paragraph-property canonicalization that brought opaque fixture `cd2f69960d5f13cc6292a138` to exactly 1.0 in both directions are recorded in `openspec/changes/archive/2026-08-17-refactor-tagged-tree-redline-construction/tasks.md`. The fixture itself was not re-run in this triage.

PR #841, opened 2026-08-14
feat(docx-comparison): default to tagged-tree comparison

### LINKED ISSUE #814: Construct redlines from a side-tagged tree instead of searching for a safe reconstruction

stevenobiajulu:
## What changed

- adds a side-tagged tree intermediate representation that retains original and revised representatives plus scoped direct-property deltas;
- adds construction, projection-isomorphism, move-relation, provenance, multi-author, and serialization checks;
- exposes the new construction to tests and controlled corpus jobs through an offline evaluator, without wiring it into the ordinary comparison pipeline;
- emits fixture-keyed divergence and formatting-fidelity evidence;
- adds synthetic and corpus-backed tests, including Word, Pages, and Google Docs reader checks; and
- records separately gated successor work in #837, #838, and #839.

## Why

The current comparator flattens OOXML, reconstructs a candidate, and then relies on accept/reject checks and fallback passes to determine whether the candidate is safe. This Stage A change preserves side and tree structure during alignment so correctness properties can be asserted before serialization, while retaining the existing implementation as the sole authoritative output path. Because Safe DOCX is local-only, evidence collection stays in local/CI corpus runs rather than a runtime shadow mode that could not report results upstream.

## Current limitation

The offline tagged evaluator reproduces accept/reject text exactly across the reviewed corpus, but a pinned direct-formatting divergence remains for one opaque fixture and its de-identified synthetic control. That blocker is tracked in #836. This PR does not flip the production default.

## Validation

- `npm run build`
- `npm run test:run -w @usejunior/docx-compare` — 847 passed, 27 skipped
- `openspec validate refactor-tagged-tree-redline-construction --strict`
- cross-reader opening/import in Microsoft Word, Apple Pages, and Google Docs using a de-identified synthetic DOCX

Closes #814 only for the approved Stage A shadow implementation; production activation remains separately gated.


stevenobiajulu:
## ✅ Post-merge smoke passed

Tested merged `main` at `c284d9f6d57671759d94e0b11b7f315e859d6871`.

### Automated gates

- ✅ Workspace build passed.
- ✅ Full workspace tests passed: 3,757 passed, 62 skipped, and 2 expected failures.
- ✅ OpenSpec strict validation passed before merge.

### Real-document end-to-end smoke

Ran the shipped `docx-comparison` CLI against a real Common Paper Mutual NDA DOCX, changing the Purpose field from `{purpose}` to `evaluating a potential supply agreement`.

- ✅ Default comparison requested and used `tagged-tree`; no legacy fallback.
- ✅ Produced 1 inserted range and 1 deleted range (9 inserted atoms / 3 deleted atoms).
- ✅ Accept All exactly matched the revised DOCX projection.
- ✅ Reject All exactly matched the original DOCX projection.
- ✅ Emitted `word/document.xml` passed the Transitional WordprocessingML schema.
- ✅ LibreOffice rendered both tracked and clean outputs successfully. Visual inspection confirmed the tracked output shows only the intended placeholder replacement and the clean output shows only the revised text, with the surrounding form layout preserved.

Production deployment: not applicable; this is a library/package repository and the comparison runtime is not deployed on merge.

Issue #226, opened 2026-05-25
Direct `reconstructionMode: 'rebuild'` bypasses evaluateRoundTripSafety entirely


stevenobiajulu:
## Context

Surfaced during PR #225 (per-story field-closure check, #212) peer review by Codex.

In \`packages/docx-core/src/baselines/atomizer/pipeline.ts\`, \`evaluateRoundTripSafety\` (and therefore the entire safety-check suite: text round-trip equality, bookmark diagnostics, **field structure**) only runs when \`reconstructionMode === 'inplace'\` is requested. When a caller explicitly passes \`reconstructionMode: 'rebuild'\`, the direct rebuild branch invokes \`runComparisonPass\` and returns the result without ever evaluating safety.

Concretely (line numbers as of \`b1f0cc4\`):
- The \`inplace\` request path threads attempts through \`evaluateRoundTripSafety\` and falls back to rebuild if any safety check fails.
- The else-branch direct \`rebuild\` path (~\`pipeline.ts:884\`) runs \`runComparisonPass(..., 'rebuild')\` and immediately returns the candidate XML — no safety evaluation.

## Why this matters

PR #225 adds a per-story field-closure check to defend against cross-story field issues that would break Word's field state machine. **That defense never executes in direct rebuild mode.** The same is true for the pre-existing text round-trip and bookmark checks — a caller who requests rebuild explicitly gets zero safety screening on the output.

This is broader than #212's scope (which was specifically about partitioning the existing safety check) but is now an obvious latent gap: if/when the \`add-footnote-support\` change lands and starts emitting field-bearing footnote XML in rebuild output, malformed output can ship unchecked.

## What needs to change

Route the direct-rebuild path through \`evaluateRoundTripSafety\` (or a rebuild-specific equivalent), so the field-structure / text-equality / bookmark checks apply uniformly. Decide what to do on failure — there's no further fallback available, so options are:
1. Return the result anyway but surface \`failedChecks\` in the diagnostics (caller-visible warning).
2. Throw an error.
3. Add a \`safetyCheckMode\` option so callers can opt in/out.

Recommend (1) for back-compat: callers that explicitly choose rebuild may be doing so deliberately and shouldn't have output hard-blocked.

## Acceptance

- [ ] \`evaluateRoundTripSafety\` (or equivalent) runs on direct rebuild output.
- [ ] Failure surfacing decided and documented.
[...]

PR #844, opened 2026-08-14
fix(docx-compare): rebind relationships across merged packages

stevenobiajulu:
## Problem

OPC relationship ids are **part-local** — ECMA-376 Part 2 §6.5.3.4 makes `Id` unique only within its own Relationships part. Two independently authored documents both number from `rId1`, so the same `rId9` routinely means an image in one and a header in the other.

The comparison output clones **one** side's package but carries a `document.xml` merged from **both**. Any reference inherited from the other side therefore either dangles or silently binds to an unrelated part of the wrong type.

Observed on the committed ILPA pair in `reconstructionMode: 'rebuild'`:

```
w:headerReference relationship 'rId9'  has type '.../image'
w:headerReference relationship 'rId10' has type '.../footer'
w:footerReference relationship 'rId12' has type '.../header'
```

`AncillaryStorySafetyError` then rejected the whole comparison.

## Approach

Mirrors the strategy `auxiliaryIdCollision.ts` already established for comment/footnote/endnote `w:id` — renumber before comparison rather than detect-and-flag, because failing closed would reject any comparison of two independently authored documents.

1. **`renumberCollidingRelationshipIds`** (pre-comparison) makes the two id spaces disjoint. Only ids that mean *different* things on the two sides are rewritten; ids resolving identically are left alone so unchanged references stay byte-stable.
2. **`importReferencedRelationships`** (package assembly) pulls the relationships the merged document still references but the base lacks, copying target parts transitively and registering content types.

**Which side gets renumbered is mode-aware.** The output clones the base, so renumbering the base is pure churn: in-place clones the revised side, rebuild clones the original. Getting this backwards renumbers the very table the result inherits — an earlier revision of this change did exactly that and churned all 33 ids on the default in-place path.

Cross-checked against Word 16.112, which rebuilds its output relationship table from scratch (inputs with header `rId8`/`rId2` → output `rId6`, with `rId8` reassigned to fontTable). This change is deliberately more conservative — it preserves the base's ids and patches in what's missing — which keeps the default path byte-stable.

## Verification

- **In-place relationship table byte-identical to `main`** on the ILPA pair — no churn on the default path.
- 832 passing in `docx-compare` (8 new), 1325 in `docx-core`, 0 failures.
- `tsc --noEmit` clean; allure-labels, conformance-citations, and the coverage ratchet all pass.

## Peer review

Codex review returned **BLOCK** on the first pass and caught two defects, both in the "silently binds to the wrong part" class this change exists to eliminate. Both are fixed, each with a committed regression test:

1. **Part-name collision** — `copyPartClosure` skipped copying when the path was occupied but added the relationship anyway, so two packages both defining `word/header7.xml` bound the import to the *base's* unrelated content. Now copied aside to a free name with the relationship retargeted; identical bytes still reuse the existing part.
2. **Hyperlink id race** — hyperlink ids were reserved only from the *original* rels table, so a minted id could take one a renumbered merge-source relationship already used. Because hyperlinks are appended first, the import then saw the id as present and skipped, leaving `r:embed` bound to a **hyperlink** instead of a header. Now reserved against both tables, plus a guard that throws rather than silently picking when an id means different things on the two sides.

## Scope

Deliberately excluded and still on separate branches: the opaque-passthrough guard relaxation (boundary-count and content-control fail-closed behavior) and the rebuild field-deletion conformance fix. Those need a decision on two pinned tests and are entangled with the tagged-tree work.

The in-place → rebuild fallback flips which package is the base after the renumbering direction is chosen. Codex's probe showed references still resolve and ids stay unique in that case (correct, but with id churn); a full-pipeline fallback integration test is a follow-up.

Refs #582

### DIFF
--- a/packages/docx-compare/src/baselines/atomizer/pipeline.ts
+++ b/packages/docx-compare/src/baselines/atomizer/pipeline.ts
@@ -119,6 +119,10 @@ import {
   restampCollidingCommentParaIds,
   type AuxiliaryPartDescriptor,
 } from './auxiliaryIdCollision.js';
+import {
+  importReferencedRelationships,
+  renumberCollidingRelationshipIds,
+} from './relationshipIdCollision.js';
 import { maybeCaptureEmittedDocumentXml } from '@usejunior/docx-core';
 import {
   AncillaryStorySafetyError,
@@ -839,6 +843,23 @@ async function compareDocumentsAtomizerCore(
   await renumberCollidingAuxiliaryIds(originalArchive, revisedArchive);
   await restampCollidingCommentParaIds(originalArchive, revisedArchive);
 
+  // Step 1c: Resolve relationship ID collisions for the same reason. `rId9`
+  // means an image in one document and a header in the other, so a merged
+  // reference resolved against the base package's table can silently bind to
+  // the wrong part. Disjoint id spaces make the merged references unambiguous;
+  // `importReferencedRelationships` then supplies the entries at assembly.
+  //
+  // Which side gets renumbered follows the base package: the output clones the
+  // base, so renumbering it would churn every id in the table the result
+  // inherits, for no correctness gain. In-place clones the revised side, rebuild
+  // clones the original, so the merge source is the opposite one each time.
+  // A later in-place -> rebuild fallback flips the base after this point; that
+  // costs id churn in the fallback output but stays correct, because both
+  // tables remain internally consistent and the two id spaces stay disjoint.
+  await (reconstructionMode === 'rebuild'
+    ? renumberCollidingRelationshipIds(revisedArchive, originalArchive)
+    : renumberCollidingRelationshipIds(originalArchive, revisedArchive));
+
   // Step 2: Extract document.xml
   const originalXml = canonicalizeWordprocessingPrefixes(await originalArchive.getDocumentXml());
   const revisedXml = canonicalizeWordprocessingPrefixes(await revisedArchive.getDocumentXml());
@@ -1222,6 +1243,11 @@ async function compareDocumentsAtomizerCore(
 
     await appendHyperlinkRelationships(resultArchive, candidate.hyperlinkRelationships);
 
+    // The merged document carries references from both sides, but the result
+    // archive is a clone of one. Import what the base package lacks, with the
+    // target parts and content types those references depend on.
+    await importReferencedRelationships(mergeSourceArchive, resultArchive, newDocumentXml);
+
     const noteMergeResults = new Map<'footnote' | 'endnote', AuxiliaryMergeResult>();
     for (const descriptor of AUXILIARY_PARTS) {
       let mergeResult: AuxiliaryMergeResult;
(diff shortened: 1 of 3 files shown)


stevenobiajulu:
CI caught a defect in my **test fixture**, not the production code:

```
element drawing: attribute 'r:embed': The attribute 'r:embed' is not allowed.
check_emitted_document_schema: 1 of 962 document.xml instances failed schema validation
```

I had put `r:embed` directly on `<w:drawing>`. That is invalid OOXML — the reference belongs on `<a:blip>` inside a `wp:inline` picture. Vitest was happy; the emitted-document schema gate, which validates every `document.xml` the pipeline produces, was not.

Fixed by using the schema-valid inline-picture shape already established in `documentReconstructor-block-sdt.test.ts`, factored into an `inlinePicture()` helper with a comment explaining why the shortcut fails.

The end-to-end assertion also got stronger while I was in there. It now checks the resolved **target**, not just the relationship type — the base side also has an image at `rId2`, so type alone did not prove the reference resolved to the merge source's picture.

Reproduced the gate locally this time (`SDX_SCHEMA_CORPUS_DIR=… npx vitest run` then `node scripts/check_emitted_document_schema.mjs`): **151 of 153 valid**, the only 2 failures being the pre-existing `#226` pins. That is the check I should have run before the first push.

Force-pushed `9b948da` → `a44ef83`. Automerge stays armed.

stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `5c986ddf5df4f221db88e2d8d775bb18de65c94d`
Built from: `main` @ `5c986dd`
Smoke: clean `npm install` + build, full suites, emitted-schema gate, real-document end-to-end

**Steps**
- ✅ clean build from `main` HEAD (no feature-branch artifacts)
- ✅ `docx-compare` — 832 passed, 27 skipped
- ✅ `docx-core` — 1325 passed, 2 expected-fail, 1 skipped
- ✅ emitted `document.xml` schema gate — 150 of 152 valid (2 pinned to #226)

**Real-world fixtures**

`tests/test_documents/redline/ILPA-Model-Limited-Partnership-Agreement-WOF_v2.docx` → `…Deal-By-Deal_v1.docx` — the committed pair whose `w:headerReference`-bound-to-an-image failure motivated this change. Driven through `compareDocumentsAtomizer` in both modes, auditing the emitted package:

| | result |
|---|---|
| references audited | 23 |
| dangling `r:id` | **0** |
| wrong relationship type | **0** (`headerReference`→header, `footerReference`→footer, `hyperlink`→hyperlink) |
| internal targets missing a part | **0** |
| relationships in table | 33 |

- **Anti-churn verified**: the in-place relationship table is identical to pre-merge `main` — the default path is byte-stable, which was the property an earlier revision of this change violated by renumbering the wrong side.
- The three `rebuild` / reversed-direction combinations still stop at the opaque-passthrough guard (`boundary count changed`). That is **pre-existing `main` behavior, unchanged by this PR** — the guard relaxation was deliberately scoped out and remains on a separate branch.

Visual evidence: skipped — this change alters package-level relationship binding, which is structurally invisible on a rendered page.
Prod: n/a — library package, no deploy-on-merge.
Log: `/tmp/automerge-smoke-844-1786743221.log` (local).

**Cleanup**
- ✅ remote branch already deleted (repo auto-deletes head branches)
- ✅ removed smoke worktree
- ✅ deleted local branch

Issue #845, opened 2026-08-14
docx-core: accept/reject strip row-level revision markers and keep the row, reporting success


stevenobiajulu:
`acceptChanges` and `rejectChanges` strip row-level revision markers by local name, leaving the row in the document and reporting success.

## Behavior

`removeAllByLocalName(root, 'del')` (`packages/docx-core/src/primitives/accept_changes.ts:338`) and `removeAllByLocalName(root, 'ins')` (`packages/docx-core/src/primitives/reject_changes.ts:441`) collect by local name only — `collectByLocalName(container, localName).filter(filter)`, with no check on the parent element (`accept_changes.ts:73-87`).

Row-level revision markers live at `w:tr/w:trPr/w:del` and `w:tr/w:trPr/w:ins`. They are *property markers on a row*, not wrappers around deleted or inserted content. Both sweeps match them, so the marker element is removed and the `w:tr` it described is left untouched.

Two consequences:

1. The projected document diverges from the host application's. A row marked deleted survives `acceptChanges`; a row marked inserted survives `rejectChanges`.
2. The divergence is unrecoverable and unreported. The marker carrying the `w:id`/`w:author`/`w:date` evidence is gone, so nothing downstream can tell the row was ever tracked — and the returned stats claim the revision was resolved.

## Reproduction

Against `packages/docx-core/src` (`npx tsx`), with a two-row table whose first row carries a row marker:

```
--- acceptChanges on a row marked DELETED
result: {"insertionsAccepted":0,"deletionsAccepted":1,"movesResolved":0,"propertyChangesResolved":0}
rows remaining: 2 | ROWTEXT present: true
<w:tr><w:trPr/><w:tc><w:p><w:r><w:t>ROWTEXT</w:t></w:r></w:p></w:tc></w:tr>

--- rejectChanges on a row marked INSERTED
result: {"insertionsRemoved":1,"deletionsRestored":0,"movesReverted":0,"propertyChangesReverted":0}
rows remaining: 2 | ROWTEXT present: true
<w:tr><w:trPr/><w:tc><w:p><w:r><w:t>ROWTEXT</w:t></w:r></w:p></w:tc></w:tr>
```

`deletionsAccepted: 1` and `insertionsRemoved: 1` are the important part: the counters report a resolved revision, so a caller checking the return value is told the operation succeeded. What it actually did was delete the audit trail and keep the row.

## The codebase already knows

[...]

Issue #764, opened 2026-07-29
docx-core: cannot make structural table changes (rows, columns, cells, merged cells) or compare table topology


stevenobiajulu:
## Goal
Safe Docx can edit and compare structural table changes (rows, columns and cells, including merged cells) with native tracked changes that accept and reject correctly. Legal documents use merged label/value tables, signature grids, schedules and form layouts, so this needs a logical grid model rather than ad hoc XML insertion.

## Tasks
- [x] Row insertion and deletion for rectangular body-level tables, with clean and tracked output and whole-row accept/reject semantics (#1001, `51af59d6`)
- [ ] #868 extract_revisions reports table-row revisions
- [ ] #1040
- [ ] #1041
- [ ] #1042
- [ ] #1043
- [ ] Markdoc syntax and compiler integration for table rows (tracked in #998)

Merged-cell safety model for every child: address operations in logical grid coordinates, not raw `w:tc` indexes; normalise admitted tables into an occupancy map without rewriting unchanged XML; require a unique rectangular transformation for edits that intersect merged regions; fail closed with structured diagnostics when ambiguous; keep unaffected rows and cells byte-equivalent where practical. Non-goals: repairing malformed source tables, inferring intent for non-rectangular merged-cell edits, and pixel-level layout equivalence.

Related: #761, #845, #868, #998

---
<details><summary>Original report (preserved from before the 2026-09-23 triage)</summary>

## Parent

Follow-up to #761.

## Why

Safe Docx can edit and compare text inside table cells, preserve table structure, and track selected table property changes. It does not yet provide a general surface for inserting or deleting rows, columns, or cells, nor can comparison reliably express arbitrary table-topology changes.

Legal and business documents frequently use merged label/value tables, signature grids, schedules, and form layouts. These require a logical grid model rather than ad hoc XML insertion.

## Proposed scope

Create an OpenSpec proposal and phased implementation for structural table editing and comparison:

1. Build a canonical logical occupancy grid for each table.
2. Add row insertion/deletion with native tracked-change semantics and accept/reject round trips.
3. Add column insertion/deletion as coordinated grid and per-row transformations.
4. Add cell insertion/deletion where the requested transformation has one deterministic rectangular interpretation.
5. Extend two-file comparison to correlate table topology and emit supported structural revisions.

The logical model must account for:

- `w:tblGrid` and grid-column widths;
- `w:gridSpan` horizontal spans;
- `w:vMerge` restart/continuation chains;
[...]

Issue #125, opened 2026-05-03
Mixed-author preservation corpus for AI accept/reject


stevenobiajulu:
## Context

This is the headline correctness property of the new design — the property that distinguishes SafeDocX from any compare-based redline tool. It's isolated from the broader invariant corpus (#124) because it's the most failure-prone area and deserves a dedicated fixture set with sharp pass/fail signal.

The current reconstruction code (`packages/docx-core/src/baselines/atomizer/documentReconstructor.ts`) is known to overwrite reviewer attribution by wrapping all detected deltas in fresh `w:ins`/`w:del` tagged with the session's author/date. This issue codifies the invariant that the new pipeline must satisfy.

## Proposal

Define and commit:

> **Invariant.** After `acceptAIEdits(doc)` in normalized non-overlap mode, every tracked-change element whose `w:id` or `w:author` is not the AI actor must be **byte-identical** (modulo namespace/whitespace normalization) to its pre-acceptance form.

Corpus must include AI + human + third-party reviewer revisions interleaved across:
- body text
- tables (rows, cells, table-grid changes)
- headers and footers
- footnotes and endnotes

Plus an **ambiguous-overlap fixture** that exercises the hard-error path from #123: an AI revision structurally containing a human revision (or vice versa). The fixture asserts the operation hard-errors with a structured failure mode listing the offending revision pairs.

## Acceptance criteria

- Fixtures committed under `packages/docx-core/test/fixtures/` (or sibling accept-test dir).
- CI fails loudly on any non-AI revision being mutated in non-overlap mode.
- Ambiguous-overlap fixture asserts hard-error response.
- Accompanying test names map clearly to the surface (#119) and selective accept (#123).

## Depends on

- #123 (selective accept)
- #124 (invariant corpus infrastructure)

## Related

- #118 (umbrella)
- #120 (canonical tracked changes — defines what "AI actor" means)

stevenobiajulu:
[...]

PR #848, opened 2026-08-14
fix(docx-primitives): preserve row-level revision markers accept/reject cannot resolve

stevenobiajulu:
## What

`acceptChanges` and `rejectChanges` swept revision wrappers by local name only, with no check on the parent element. Row-level revision markers live at `w:tr > w:trPr > w:del` and `w:tr > w:trPr > w:ins` and describe the **row** rather than wrapping content, so both sweeps matched them: the marker was removed and the `w:tr` it described was left behind.

This preserves the marker in the two directions the engine cannot resolve and reports the count. It does **not** implement the row semantics — see *Out of scope*.

## Why it mattered

The marker carries `w:id`, `w:author` and `w:date`. Destroying it left no residual record that the row had ever been tracked, so nothing downstream could detect the divergence. And the marker was counted as a resolved revision:

```
acceptChanges on a row marked DELETED
  result: {"deletionsAccepted":1, ...}
  rows remaining: 2        <- the row Word removes survives
  trPr marker survives: false
```

A caller checking the return value was told the operation succeeded while holding a document that does not match what a word processor projects.

The repository already knew. `cli/conformance-adapter.ts` classifies accept-over-`trPr>del` and reject-over-`trPr>ins` as `supported: false`, and `[XIMPL-08]` pins that. The honesty stopped at the conformance harness — `DocxDocument` and the `accept_changes` MCP tool called the primitives unconditionally.

## Approach

Preserve-and-report, following the convention already applied to other unresolved advanced revision records (`ADV-UNRESOLVED-RECORDS-01`, `ADV-TOPOLOGY-PRESERVATION-01`): a record left in place is a visible gap a caller can detect; removing it while keeping the content is silent divergence.

Only the two genuinely unresolvable directions are guarded. Accepting a `trPr>ins` and rejecting a `trPr>del` already produce the correct document — keep the row, drop the marker — so those paths are untouched.

- `AcceptChangesResult` / `RejectChangesResult` gain `unresolvedRowRevisions`, aggregated across body and side stories. It is deliberately separate from the resolved counters, and `hasAcceptedChanges` / `hasRejectedChanges` do not treat a preserved marker as a change.
- Rejecting a `w:trPrChange` restores row properties by replacing the whole `w:trPr`; that replacement now carries surviving markers across (see below).

## Peer review (Codex) — one real bug caught

**Phase-ordering leak (P1).** A `w:trPr` can hold both a row marker and a `w:trPrChange`. Phase C preserved the marker, then Phase F replaced the entire `w:trPr` with the snapshot and destroyed it — while `unresolvedRowRevisions` still reported it. Reproduced, then fixed:

```
before: reject ins+trPrChange -> {"unresolvedRowRevisions":1}, w:ins in trPr: 0, w:id="7" absent
after:  same input           -> {"unresolvedRowRevisions":1}, w:ins in trPr: 1, attributes intact
```

This also protects a **foreign** marker during a selective reject, which the byte-untouched invariant (#125) requires. The accept side needed no guard — Phase D removes the change record without replacing its parent.

**MCP contract overpromised (P1/P2).** The `mcp-server` requirement, `tool_catalog.ts`, and the generated tool reference all said `accept_changes` produces "a clean document with no revision markup" — no longer true. Fixed via a MODIFIED requirement delta plus a regenerated tool reference.

**Source compatibility (P2).** `unresolvedRowRevisions` is required, not optional, on two exported types. Runtime-additive, but a source-level break for external TypeScript that *constructs* or mocks them. Documented in the proposal with the reasoning for keeping it required.

Codex also confirmed the two load-bearing assumptions against the vendored strict and transitional schemas: `CT_TrPr` admits only `ins`/`del`/`trPrChange` as revision children, paragraph-mark markers live under `pPr > rPr`, cell topology uses distinct `cellIns`/`cellDel` names, and nested tables do not change the direct-parent test.

## Verification

Red/green confirmed rather than assumed. With the source edits stashed, the four original tests fail on `expected null not to be null` (marker stripped) and `expected undefined to be +0` (no counter). The two `trPrChange` tests were separately verified red against the un-fixed Phase F.

| Gate | Result |
|---|---|
| `npm run build` | pass |
| `npm run test:run` | pass |
| `npm run lint:workspaces` | pass |
| `npm run check:spec-coverage` | pass — 6/6 primitives + 4/4 mcp-server scenarios mapped |
| `check:tool-docs` | pass |
| `check:conformance-citations` / `-doc` | pass |
| `check:allure-labels` / `-filenames` / `-quality` | pass |
| `openspec validate --strict` | valid |

## Out of scope

Resolving row revisions semantically (accept-deletes-the-row, reject-removes-the-inserted-row). That needs an oracle pass over Word's real projection — `w:vMerge` chains and cells carrying their own content revisions have to be settled first — and it overlaps #764, which owns structural table changes and lists accept/reject round trips in its scope. No Word or LibreOffice oracle was run here, so application-specific normalization is unverified; this change makes the gap visible instead of guessing.

Extending `extract_revisions` to report row and cell topology, so a caller can inspect a document *before* accepting, is also left out. Today the counter reports the gap after the fact.

## Provenance

Surfaced while triaging recurring public support signals for tracked-change handling, then independently reproduced against safe-docx. No fixture derives from any external document.

Ref #845

### DIFF
--- a/openspec/changes/guard-row-level-revision-resolution/oracle-probe.md
+++ b/openspec/changes/guard-row-level-revision-resolution/oracle-probe.md
@@ -1,0 +1,80 @@
+# Oracle probe: can LibreOffice adjudicate row-level revisions?
+
+Recorded so the claim in `proposal.md` (`## Out of scope`) is reproducible rather than asserted.
+Run from the worktree root with `npx tsx`, in a **real terminal** — `soffice` aborts (SIGABRT) under a
+sandboxed shell, which is a known environment limitation, not a regression.
+
+## Probe
+
+```ts
+import { resolveSoffice, runLibreOfficeOracle } from './packages/docx-core/src/integration/libreoffice-oracle.js';
+const W = 'http://schemas.openxmlformats.org/wordprocessingml/2006/main';
+
+// Row 1 carries the row-level marker AND a content deletion inside its cell.
+// The content deletion is the POSITIVE CONTROL: if LibreOffice resolves it but
+// leaves the row marker alone, LO imported content revisions and ignored the row one.
+const doc = (marker: 'ins' | 'del') =>
+  `<w:document xmlns:w="${W}"><w:body><w:tbl>`
+  + `<w:tblPr/><w:tblGrid><w:gridCol w:w="4680"/></w:tblGrid>`
+  + `<w:tr><w:trPr><w:${marker} w:id="7" w:author="Reviewer" w:date="2026-01-01T00:00:00Z"/></w:trPr>`
+  + `<w:tc><w:tcPr/><w:p><w:r><w:t>ROWMARKED</w:t></w:r>`
+  + `<w:del w:id="9" w:author="Reviewer" w:date="2026-01-01T00:00:00Z">`
+  + `<w:r><w:delText>CONTENTDEL</w:delText></w:r></w:del></w:p></w:tc></w:tr>`
+  + `<w:tr><w:tc><w:tcPr/><w:p><w:r><w:t>CONTROLROW</w:t></w:r></w:p></w:tc></w:tr>`
+  + `</w:tbl><w:p><w:r><w:t>TAIL</w:t></w:r></w:p></w:body></w:document>`;
+
+const out = await runLibreOfficeOracle([
+  { op: 'identity', documentXml: doc('del') },
+  { op: 'identity', documentXml: doc('ins') },
+  { op: 'accept',   documentXml: doc('del') },
+  { op: 'reject',   documentXml: doc('ins') },
+], resolveSoffice()!);
+// then report, per result: w:tr count, ROWMARKED/CONTROLROW/CONTENTDEL presence, and the trPr block
+```
+
+## Captured output (2026-08-14, LibreOffice via `/opt/homebrew/bin/soffice`)
+
+```text
+--- IDENTITY (load+save) with trPr>del  — does LO preserve the row marker at all?
+  rows: 2 | ROWMARKED: true | CONTROLROW: true
+  CONTENTDEL text present: true | any w:del elem: true | any w:ins elem: false
+  trPr block: <w:trPr></w:trPr>
+
+--- IDENTITY (load+save) with trPr>ins  — same question
+  rows: 2 | ROWMARKED: true | CONTROLROW: true
+  CONTENTDEL text present: true | any w:del elem: true | any w:ins elem: false
+  trPr block: <w:trPr></w:trPr>
+
+--- ACCEPT over trPr>del
+  rows: 2 | ROWMARKED: true | CONTROLROW: true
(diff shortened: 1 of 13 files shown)


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `303cedf440e4467b6c359b0c34a81df2bc38a67b`
Built from: `main` @ `303cedf` (clean worktree, fresh `npm install`, no feature-branch artifacts)
Smoke: `npm run build && npm run lint:workspaces && npm run test:run` + real-document end-to-end

**Steps**
- ✅ build
- ✅ lint
- ✅ tests (full workspace suite, exit 0)
- ✅ accept/reject over a real agreement, via `DocxDocument`

**Real-world fixtures**
- `nvca-stock-purchase-agreement` (via `ensureSourceDocx` cache, SHA-256-verified, 140,221 bytes, 6 real table rows)
  - **Regression:** the untouched real agreement accepts cleanly — `unresolvedRowRevisions=0`, all 6 rows survive. No phantom counter on real input.
  - **Feature (accept):** a row-deletion marker placed in the agreement's own first `<w:tr>` → `unresolvedRowRevisions=1`, `deletionsAccepted=0`, `w:id="90001"`/`w:author` intact, marker still a `w:trPr` child, all 6 rows present.
  - **Feature (reject):** row-insertion marker → `unresolvedRowRevisions=1`, `insertionsRemoved=0`, evidence preserved.
  - Note on method: neither safe-docx nor either available oracle can *author* a row-level revision, so the marker is injected into the real document's own table — the shape Word emits when a reviewer deletes a table row with track changes on. The document, its tables and its package are real; only the marker under test is synthetic.

Visual evidence: skipped, and for a reason worth recording — our only headless renderer (LibreOffice) **discards row-level revision markers on import** (see `oracle-probe.md` in this change), so a rendered page cannot show the pre-fix/post-fix difference. A screenshot here would be evidence of nothing.

**Prod**: ✅ Production deployment for `303cedf` completed successfully. Library-only change, not user-visible on the site, so SHA match is the verification.

⚠️ **Advisory gate finding (non-blocking, not swallowed):** the LLM gate returned **Overall: WARN** (7 pass, 1 warn, 0 error) even though its check-run is green. The warning is legitimate — `isRowPropertyRevisionMarker` and `countUnresolvedRowRevisions` are duplicated verbatim across `accept_changes.ts` and `reject_changes.ts` instead of being shared. The two files already duplicate several helpers, so this follows the local pattern rather than breaking it, and `reject_changes.ts` already imports `RevisionFilter` from `accept_changes.ts` — so consolidation is straightforward. Worth a follow-up; it does not affect correctness.

**Cleanup**
- ⚠️ remote branch already deleted — repo has auto-delete enabled
- ✅ removed worktree `.worktrees/845-row-revisions-20260814`
- ✅ deleted local branch `845-fail-closed-row-revisions-20260814`

Log: `/tmp/automerge-smoke-848-20260814-183609.log` (local).

PR #856, opened 2026-08-15
fix(docx-compare): keep revision boundaries outside parentheses

### LINKED ISSUE #851: docx-compare: deletion boundary splits inside a parenthetical enumerator, orphaning the closing bracket
## Symptom

Comparing the committed ILPA pair, §14.7.1 renders with a deletion that starts *inside* a parenthetical, leaving a stray closing bracket:

```
Ours:   If, upon[- any of-] (i[-) the first anniversary … (iii-])
```

The reader sees `(i`, then a long strike-through, then an orphaned `)`. The enumerator `(i)` is visually broken apart even though it survives the edit.

## What the oracles do

Both reference implementations insert the new enumerator and delete the whole old one, rather than splitting at the bracket:

```
Word 16.112:        If, upon {+(i+}[-any of (i) the first anniversary … (iii-])
Aspose.Words 25.10: If, upon [-any of (i) … (iii-]{+(i+})
```

(`[-…-]` deleted, `{+…+}` inserted. Word and Aspose differ only in the order they emit the pair.)

## Cause

We anchor on the literal `(i` prefix shared by both sides and split the deletion immediately after it, so the `)` that belongs to `(i)` falls at the head of the deleted span. The common-prefix match is real but not a good place to cut.

## Expected

Match Word: do not split inside a parenthetical enumerator. Deleting the whole old enumerator and inserting the new one reads correctly and is what both mature implementations produce.

## Notes

- An earlier suggestion was a bracket-matching heuristic. The oracle evidence points at the simpler answer: do not split at that boundary at all.
- The neighbouring `…Giveback),` case was reviewed and judged acceptable as-is: we and Word both delete `),` and re-insert it. Aspose is more surgical there, which is an oracle divergence rather than the behavior to follow.
- Reproduce with `reconstructionMode: 'inplace'` on `tests/test_documents/redline/ILPA-Model-Limited-Partnership-Agreement-WOF_v2.docx` → `ILPA-Model-Limited-Parnership-Agreement-Deal-By-Deal_v1.docx`, then read the paragraph containing "the first anniversary following the end of the Commitment Period".

stevenobiajulu:
## Summary
- keep ordinary token equality, including unchanged markers such as `(i)`
- remove only orphaned parenthesis matches whose mates are not matched to each other
- prevent deletion wrappers from ending immediately before an unchanged closing `)`
- remove the superseded Roman/contextual-marker classification and similarity policy

## Concrete behavior

For:

```text
(i) deleted text (parenthetical) retained tail
```

compared with:

```text
(i) retained tail
```

the marker remains equal and the deletion is exactly `deleted text (parenthetical)`. The revision boundary no longer splits either parenthetical pair.

## Verification
- focused boundary suite: 9 passed, including the exact synthetic repro and the real ILPA pair
- full docx-compare suite: 848 passed, 27 expected corpus skips
- workspace TypeScript lint passed; only 6 pre-existing ESLint warnings remain
- ECMA-376 citation hygiene and diff checks passed

Automerge remains disabled pending review.

Fixes #851


stevenobiajulu:
Closing after fresh Word and Aspose oracle checks showed the proposed behavior would be a regression in parity.

For the reduced ILPA §14.7.1 pair:

- Microsoft Word 16.112 emits bare deleted `iv` and bare inserted `ii`, leaving the surrounding parentheses equal.
- Licensed Aspose.Words 25.10 emits the same boundary.
- Current safe-docx emits the same boundary.
- This PR would instead delete `(iv)` and insert `(ii)`.

For the synthetic case closer to the reported concern:

```text
(i) deleted text (parenthetical) retained tail
→
(i) retained tail
```

Word emits one deletion containing exactly `deleted text (parenthetical) `: the unchanged `(i)` remains outside the deletion and the trailing parenthetical closes inside it. Current safe-docx already produces that same result.

The Word automation opened and closed only its staged documents; the pre-existing Word document count remained 5 before and after.

Conclusion: there is no demonstrated correctness defect to fix. The proposed atomic-marker policy is aesthetic, changes revision semantics, and diverges from both reference implementations.

Issue #858, opened 2026-08-15
docx-release-verifier: requireNativeComments passes on a DOCX with zero native comments


stevenobiajulu:
## Summary

A release verification configured with `requireNativeComments: true` can pass when the tracked DOCX contains zero native Word comments. The current check validates consistency of an empty set but does not enforce the requested nonempty invariant.

## Reduced reproduction

1. Build a tracked DOCX containing at least one revision and no comment definitions or references.
2. Run the release verifier with `requireNativeComments: true`.

## Actual behavior

The native-comment check can report success with `count: 0`.

## Expected behavior

`requireNativeComments: true` means at least one valid native comment must exist. Zero comments must fail closed with an actionable diagnostic.

## Acceptance criteria

- [ ] A tracked DOCX with zero comments fails when `requireNativeComments: true`.
- [ ] The same DOCX may pass the comment-presence gate when the option is false or omitted.
- [ ] One structurally valid native comment satisfies the nonempty requirement.
- [ ] Dangling references, duplicate definitions, or malformed ranges still fail independently.
- [ ] Add a negative control proving the required-comments gate turns red when the final comment is removed.
- [ ] Report the actual comment count against the required minimum in the certificate. The manifest carries no caller-supplied expected count, so the actionable invariant is "expected minimum 1, actual 0" rather than a comparison against a declared number.

## Related nonvacuity audit

Audit other opt-in completeness gates for empty-set success. If a caller explicitly requires rationales, requirements, assertions, or change sets, an empty collection should not satisfy that requirement unless the API names and documents an allow-empty mode.

## Relationship to contextual-anchor alignment

No overlap. This is a release-certificate nonvacuity defect; contextual-anchor alignment affects only comparison atom eligibility.

## Verification at `7bd35c8`

Confirmed reachable end-to-end through the public `verifyRelease()` API, not merely in the internal helper.

`commentIntegrity` in `packages/docx-release-verifier/src/archive.ts:44-56` derives four ID lists from the document and comments parts. When the tracked DOCX has no comments, all four are empty: every `.every(Boolean)` is vacuously true and the local `exact()` helper returns true for equal empty arrays, so the gate returns `pass`.

There is no upstream nonempty check — `packages/docx-release-verifier/src/verifier.ts` installs `tracked.commentVerdict` directly into the required gates.

Observed output for a tracked DOCX with one revision and zero comments, run with `requireNativeComments: true`:

```text
[...]

PR #866, opened 2026-08-16
fix(docx-release-verifier): compare comment ID multiplicities, not membership

### LINKED ISSUE #863: docx-release-verifier: comment integrity passes when IDs are duplicated identically across all four collections
## Summary

`commentIntegrity` in `packages/docx-release-verifier/src/archive.ts` compares comment ID lists by membership and total length, not by multiplicity. A tracked DOCX whose comment IDs are duplicated *identically* across all four sides — `commentRangeStart`, `commentRangeEnd`, `commentReference`, and the `w:comment` definitions — satisfies every consistency check and returns a required `pass`.

This is the sibling of #858. That issue fixed the empty-set vacuous pass; this is the same helper passing on a different degenerate shape it cannot distinguish.

## Reduced reproduction

Build a tracked DOCX where each of the four ID collections is `['1', '1']`, then run the verifier with `requireNativeComments: true`.

## Actual behavior

Reachable end-to-end through the public `verifyRelease()` API:

```text
{"comments":{"status":"pass","required":true,"details":{"expectedMinimum":1,"count":2}},
 "delivery":{"status":"pass","required":true},
 "exitCode":0}
```

Mechanism:

- `archive.ts:53` — `exact()` is `left.length === right.length && left.every((id) => right.includes(id))`, which is set-style and ignores multiplicity. Symmetric duplication keeps the lengths equal and every member present.
- `archive.ts:61` — the minimum added by #858 only rejects zero records, and a duplicate set of length ≥ 1 clears it.
- `xml.ts:240` — `commentIds()` maps every matching element and does not deduplicate, so duplicates genuinely reach the comparison.

## Existing coverage does not catch it

`verifier.test.ts:277` duplicates only the definitions while leaving a single range and reference. That asymmetry produces a length mismatch, so it correctly fails. The fully symmetric duplicate shape is uncovered.

## Conformance note (CORRECTED)

**An earlier version of this issue claimed duplicate comment IDs are non-conformant under ECMA-376's normative prose. That claim was wrong and is retracted.**

A dynamic review extracted the official ECMA-376 fifth-edition Part 1 text for §17.13.4.2 (`comment (Comment Content)`). What it actually says:

- `w:id` is described as "a unique identifier for an annotation within a WordprocessingML document" — descriptive prose.
- The same section **explicitly anticipates duplicates and defines consumer behavior for them**: "If more than one comment shares the same value for the id attribute, then only ...". A section that prescribes how a consumer handles a shape is not declaring that shape non-conformant.
- The same section *does* explicitly call an **omitted** `id` non-conformant — demonstrating that it says so when it means it.

Two things therefore do NOT support this fix, and must not be used to justify it:

1. **The schema.** The vendored transitional `wml.xsd` permits unbounded `w:comment` children and carries no `xs:key` uniqueness constraint. This shape is not structurally XSD-invalid.
2. **ECMA-376 conformance.** The spec does not prohibit duplicate comment IDs; it defines behavior for them.

**The correct justification is a policy one.** Duplicate IDs make comment linkage ambiguous, and the spec's disambiguation rule means one comment's content is silently disregarded by the consumer. A release verifier whose job is to fail closed should reject that ambiguity in a deliverable rather than depend on a consumer's recovery behavior. That is a deliberately stricter-than-spec integrity invariant, and it should be described that way in code and in the changelog.

Do not add a `@conformance` tag for this. `check:conformance-citations` verifies only that a cited section resolves in the registry — it cannot verify that the claim is semantically true, so a passing lint is not evidence the rule exists.

## Severity

Low — hardening rather than an active false assurance. Word and safe-docx both allocate distinct sequential IDs, so coordinated duplication across all four sides is more plausibly the product of a broken producer, hand-edited XML, or a faulty merge than of ordinary authoring. It is still a real integrity hole in a gate whose entire job is to fail closed.

## Acceptance criteria

- [ ] Comment integrity compares ID multiplicities, not only membership and total length.
- [ ] A DOCX with IDs duplicated identically across starts, ends, references, and definitions fails the required comments gate.
- [ ] One structurally valid comment continues to pass.
- [ ] Existing dangling-reference, asymmetric-duplicate, malformed-range, and zero-comment cases continue to fail independently.
- [ ] Add a regression test that goes through the public `verifyRelease()` API for the symmetric duplicate shape.
- [ ] Preserve the dependency boundary — `dependency-boundary.test.ts` forbids importing Safe DOCX parsers, mutators, comparers, or generators to build fixtures.

## Provenance

Surfaced while fixing #858 and deliberately left out of that PR to keep the diff scoped. Premise verified dynamically against `1e9785a` (`npm ci`, `npm run build`, and a public-API probe); the output above is the actual probe result.

Refs #858

stevenobiajulu:
Fixes #863

## Problem

`commentIntegrity` in `packages/docx-release-verifier/src/archive.ts` compared comment ID lists with a set-style helper (`left.length === right.length && left.every((id) => right.includes(id))`), which ignores multiplicity. A tracked DOCX whose comment IDs are duplicated *identically* across all four collections — `commentRangeStart`, `commentRangeEnd`, `commentReference`, and the `w:comment` records, each e.g. `['1','1']` — satisfied every consistency check and returned a required `pass` through the public `verifyRelease()` API. The #858 zero-comment minimum does not catch it, because a duplicate set of length >= 1 clears the minimum.

## Fix

Two halves, because multiset equality alone is not sufficient — identical duplication on all four sides yields four *equal* multisets:

1. `exact()` now compares per-ID counts. This also rejects equal-length, set-equal shapes the old comparison passed outright (e.g. starts `['1','1']` against records `['1','2']`), keeping the existing "disagree" verdict for them.
2. The comment records must be duplicate-free. This rejects the fully symmetric duplicate shape with a diagnostic that names the duplicated IDs.

The #858 zero-comment guard is untouched and still fails closed with its existing message and `{expectedMinimum, count}` details.

## Framing (revised after peer review)

Duplicate rejection is a **conservative release-verifier integrity invariant** — deliberately stricter than the spec — because unambiguous comment linkage requires each identifier to resolve to exactly one record. It is *not* framed as an ECMA-376 conformance violation: the transitional `wml.xsd` carries no uniqueness constraint (unbounded `w:comment` children, no `xs:key`), and ECMA-376 Part 1 § 17.13.4.2 itself anticipates duplicate id values by defining consumer fallback behavior for them rather than prohibiting them. A gate whose job is to fail closed chooses not to rely on that fallback. (An earlier revision of this branch carried a `@conformance` citation claiming a normative uniqueness rule; peer review checked the actual spec text and refuted it — see the adjudication comment below.)

## Severity

Hardening, not an active false assurance. Word and safe-docx both allocate distinct sequential comment IDs, so coordinated duplication across all four sides most plausibly comes from a broken producer, hand-edited XML, or a faulty merge. It is still a real integrity hole in a gate whose job is to fail closed.

## Tests

Extended `verifier.test.ts` (no new test file) with one regression test through the public `verifyRelease()` API covering the symmetric duplicate (fails with the duplicate diagnostic, `duplicates: ['1']`, exit 1), the equal-length multiplicity skew (fails with the existing "disagree" reason), and an order-insensitivity positive control (two distinct IDs in different orders per collection pass with `count: 2`, exit 0). Existing comment-integrity tests now assert the complete diagnostic strings and exact details objects. Fixtures are built with the file's local JSZip helper, preserving the `dependency-boundary.test.ts` rule against importing Safe DOCX parsers/mutators/comparers/generators.

## Pre-submit

`npm run build && npm run lint:workspaces && npm run test:run && npm run check:spec-coverage && npm run check:conformance-citations && npm run check:conformance-doc` — exit 0 (rerun after the review revision).

### DIFF
--- a/packages/docx-release-verifier/src/archive.ts
+++ b/packages/docx-release-verifier/src/archive.ts
@@ -56,13 +56,17 @@ function idMultiset(ids: string[]): Map<string, number> {
  * multiset, and each identifier must be defined exactly once. Comparing
  * multiplicities — not just membership and total length — rejects documents a
  * set-style comparison cannot distinguish from valid ones, such as identifiers
- * duplicated identically across all four collections. The transitional
- * `wml.xsd` alone does not reject duplicate identifiers (it permits unbounded
- * `w:comment` children and carries no uniqueness constraint); the violated
- * rule is the normative prose defining the annotation identifier as a unique
- * identifier for the comment.
+ * duplicated identically across all four collections.
  *
- * @conformance ECMA-376 edition 5, Part 1 § 17.13.4.2
+ * Rejecting duplicate identifiers is a conservative release-verifier
+ * integrity invariant, deliberately stricter than the spec: unambiguous
+ * comment linkage requires each identifier to resolve to exactly one comment
+ * record. The transitional `wml.xsd` carries no uniqueness constraint (it
+ * permits unbounded `w:comment` children), and the spec prose anticipates
+ * duplicate identifiers by defining consumer fallback behavior for them
+ * rather than prohibiting them. A release gate whose job is to fail closed
+ * does not rely on that fallback.
+ *
  * @see https://github.com/UseJunior/safe-docx/issues/863
  */
 function commentIntegrity(documentXml: string, commentsXml: string | null, required: boolean): Verdict {
--- a/packages/docx-release-verifier/src/archive.ts
+++ b/packages/docx-release-verifier/src/archive.ts
@@ -44,17 +44,53 @@ async function boundedText(zip: JSZip, name: string): Promise<string> {
 /** Requiring native comments means at least this many valid comments must exist. */
 const REQUIRED_COMMENT_MINIMUM = 1;
 
+function idMultiset(ids: string[]): Map<string, number> {
+  const counts = new Map<string, number>();
+  for (const id of ids) counts.set(id, (counts.get(id) ?? 0) + 1);
+  return counts;
+}
+
+/**
+ * Verifies that native comment markup is internally consistent: range starts,
+ * range ends, references, and comment records must carry the same identifier
+ * multiset, and each identifier must be defined exactly once. Comparing
+ * multiplicities — not just membership and total length — rejects documents a
+ * set-style comparison cannot distinguish from valid ones, such as identifiers
+ * duplicated identically across all four collections. The transitional
+ * `wml.xsd` alone does not reject duplicate identifiers (it permits unbounded
+ * `w:comment` children and carries no uniqueness constraint); the violated
+ * rule is the normative prose defining the annotation identifier as a unique
+ * identifier for the comment.
+ *
+ * @conformance ECMA-376 edition 5, Part 1 § 17.13.4.2
+ * @see https://github.com/UseJunior/safe-docx/issues/863
+ */
 function commentIntegrity(documentXml: string, commentsXml: string | null, required: boolean): Verdict {
   if (!required) return { status: 'not_run', required: false, reason: 'Native comments not required.' };
   const starts = commentIds(documentXml, 'commentRangeStart');
   const ends = commentIds(documentXml, 'commentRangeEnd');
   const references = commentIds(documentXml, 'commentReference');
   const records = commentsXml ? commentIds(commentsXml, 'comment') : [];
-  const exact = (left: string[], right: string[]) => left.length === right.length && left.every((id) => right.includes(id));
+  const exact = (left: string[], right: string[]) => {
+    if (left.length !== right.length) return false;
+    const expected = idMultiset(right);
+    return [...idMultiset(left)].every(([id, count]) => expected.get(id) === count);
+  };
   const consistent = starts.every(Boolean) && ends.every(Boolean) && references.every(Boolean) && records.every(Boolean)
     && exact(starts, ends) && exact(starts, references) && exact(starts, records);
   if (!consistent) {
     return { status: 'fail', required: true, reason: 'Native comment records, range starts, range ends, and references disagree.', details: { starts, ends, references, records } };
+  }
+  // Identical duplication across all four collections survives the multiset
+  // comparison, so the definitions must additionally be duplicate-free.
+  const duplicates = [...idMultiset(records)].filter(([, count]) => count > 1).map(([id]) => id);
+  if (duplicates.length > 0) {
+    return {
+      status: 'fail',
+      required: true,
+      reason: `Native comment IDs are duplicated across range starts, range ends, references, and records: ${duplicates.join(', ')}. Each annotation identifier must be unique.`,
+      details: { duplicates, starts, ends, references, records },
+    };
   }
   // Consistency of an empty set is vacuous: requiring native comments demands
   // at least one valid comment, so zero comments fails closed (issue #858).
--- a/packages/docx-release-verifier/src/verifier.test.ts
+++ b/packages/docx-release-verifier/src/verifier.test.ts
@@ -300,6 +300,33 @@ describe('independent release verifier', () => {
     expect(malformed.exitCode).toBe(1);
   });
 
+  itAllure('fails comment IDs duplicated identically across starts, ends, references, and records', async () => {
+    const { manifest } = await fixture();
+    const revision = '<w:r><w:t xml:space="preserve">Hello </w:t></w:r><w:del w:id="1"><w:r><w:delText>old</w:delText></w:r></w:del><w:ins w:id="2"><w:r><w:t>new</w:t></w:r></w:ins>';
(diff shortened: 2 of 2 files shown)


stevenobiajulu:
## Codex peer-review verdict and adjudication

**Reviewer:** Codex CLI (`codex exec --sandbox workspace-write`), dynamic review against this worktree. Raw output retained in the gitignored `.peer-review/` directory.

**Verdict: changes requested — one finding, adopted.**

### What the reviewer verified dynamically (all confirmed, no changes needed)

- Ran `npx vitest run src/verifier.test.ts`: 16/16 pass.
- Every pre-existing failure shape keeps its verdict and branch: dangling reference, asymmetric duplicate definitions (records `['1','1']` vs singleton ranges — still fails on the length check with the "disagree" reason, not re-routed to the new duplicate branch), malformed range, and the #858 zero-comment case with its exact "expected minimum 1, actual 0" diagnostic.
- Order-insensitivity of the multiset comparison confirmed (no false-fail regression for valid multi-comment documents).
- `npm run check:conformance-citations` passed on the original revision.

### The finding (P1, adopted): the ECMA-376 conformance claim was unsupported

The original revision justified duplicate rejection with `@conformance ECMA-376 edition 5, Part 1 § 17.13.4.2`, claiming the section's normative prose defines the annotation identifier as unique. The reviewer retrieved the official ECMA-376 fifth-edition Part 1 text and read § 17.13.4.2 directly:

- The `w:id` description ("a unique identifier for an annotation within a WordprocessingML document") is descriptive prose, not a mandatory constraint.
- The same section **explicitly anticipates duplicates** and defines consumer behavior for them ("If more than one comment shares the same value for the id attribute, then only …"). A spec that tells consumers how to handle a shape is not declaring that shape non-conformant.
- By contrast, the section *does* use explicit non-conformance language for an **omitted** id — demonstrating it says so when it means it.

The citation lint only proves a tag resolves against the registry; it cannot check whether the claim is semantically true. Under this repo's standing rule that a conformance claim with no quotable mandatory spec text is itself a bug, the claim had to go. Note: issue #863's own "Conformance note" section carries the same overclaim and is being corrected separately — it should not be cited as justification either.

**Change made (commit 53f4c27):** the JSDoc now frames duplicate rejection as a conservative release-verifier integrity invariant — deliberately stricter than the spec, required for unambiguous comment linkage — with no ECMA-376 citation or mention. It keeps the accurate statement that transitional `wml.xsd` has no uniqueness constraint, and notes that the spec defines consumer fallback behavior for duplicates which a fail-closed gate chooses not to rely on. The same commit strengthens the tests to assert complete diagnostic strings (not substrings), pins the exact zero-comment details object, and adds the order-insensitivity positive control.

Full pre-submit gate rerun after the revision: exit 0.

stevenobiajulu:
## Post-merge smoke: PASS

Ran against merged main (`447be1b`), worktree detached onto the squash commit and `@usejunior/docx-release-verifier` rebuilt from it. Each scenario builds a tracked DOCX with JSZip only (dependency boundary preserved) and drives the packaged CLI (`dist/cli.js --manifest …`, i.e. the public `verifyRelease()` path) with `requireNativeComments: true`, asserting on the real process exit code. Actual output:

```text
--- symmetric-duplicate (all four collections [1,1])
exit=1 (expected 1) OK
comments gate: {"status":"fail","required":true,"reason":"Native comment IDs are duplicated across range starts, range ends, references, and records: 1. Each annotation identifier must be unique.","details":{"duplicates":["1"],"starts":["1","1"],"ends":["1","1"],"references":["1","1"],"records":["1","1"]}}
--- one valid comment
exit=0 (expected 0) OK
comments gate: {"status":"pass","required":true,"details":{"expectedMinimum":1,"count":1}}
--- zero comments
exit=1 (expected 1) OK
comments gate: {"status":"fail","required":true,"reason":"Native comments were required but the tracked DOCX contains none: expected minimum 1, actual 0.","details":{"expectedMinimum":1,"count":0}}
--- asymmetric duplicate (records [1,1], ranges/refs [1])
exit=1 (expected 1) OK
comments gate: {"status":"fail","required":true,"reason":"Native comment records, range starts, range ends, and references disagree.","details":{"starts":["1"],"ends":["1"],"references":["1"],"records":["1","1"]}}
--- order-insensitivity control (two distinct IDs, different orders per collection)
exit=0 (expected 0) OK
comments gate: {"status":"pass","required":true,"details":{"expectedMinimum":1,"count":2}}
SMOKE PASS
```

All acceptance criteria of #863 confirmed end-to-end on the merged artifact: the symmetric duplicate now fails closed with a diagnostic naming the duplicated ID, one valid comment still passes, the #858 zero-comment minimum is intact with its exact message and details, the pre-existing asymmetric-duplicate shape still fails on disagreement, and valid multi-comment documents remain order-insensitive.

Post-merge gate audit: Vercel commit status `success` on both the PR head (`53f4c27`) and the merge commit (`447be1b`); LLM review gate Overall: PASS (5 pass, 0 warn, 0 error).

PR #869, opened 2026-08-16
fix(docx-core): preserve embedded objects through text replacement

### LINKED ISSUE #739: replace_text: blanking a paragraph's text silently deletes a sibling image run (data loss)
## Summary

Blanking a paragraph's text with `replace_text` (complete visible text as `old_string`, empty `new_string`) can **silently delete an image in the same paragraph**. The tool returns `success: true` and says nothing about the removed drawing.

`mergeRunsOnly` folds an image run into an adjacent text run when the image run carries no `w:rPr`. Blanking the text then removes the merged run, taking the `<w:drawing>` with it.

## Reproduction

A numbered paragraph containing a caption run followed by an image run with no `w:rPr`:

```xml
<w:p>
  <w:pPr><w:numPr><w:ilvl w:val="0"/><w:numId w:val="1"/></w:numPr></w:pPr>
  <w:r><w:t xml:space="preserve">Figure 1 caption.</w:t></w:r>
  <w:r><w:drawing><wp:inline>…<a:blip r:embed="rId9"/>…</wp:inline></w:drawing></w:r>
</w:p>
```

```
safe-docx edit img.docx --replace <paraId> 'Figure 1 caption.' '' -o out.docx
```

Result (clean save), counting elements in `word/document.xml`:

```
img.docx   w:drawing=1  a:blip=1
out.docx   w:drawing=0  a:blip=0
```

The whole `<w:p>` is emitted as:

```xml
<w:p><w:pPr><w:numPr>…</w:numPr></w:pPr></w:p>
```

## Scope

- Not a save-time artifact: the image survives open + save untouched.
- Survives any edit that leaves **some** text behind. Only full blanking destroys it.
- Only occurs when the image run is format-identical to the text run (no `w:rPr`), so it is merge-dependent — a format-distinct image run is unaffected.

## Impact

Silent, unreported data loss in a documented workflow. An orphan list label is visible on the page; a missing exhibit, org chart or signature graphic is not. Callers have no signal to check.

## Suggested fix

- `mergeRunsOnly` should not merge a run containing `w:drawing`/`w:pict`/`w:object` into a text run, regardless of `w:rPr` equality.
- Independently, any edit that removes a run carrying embedded content should surface a warning (see #686 re: `replace_text` having no `warnings` field).

Found while dogfooding an agent-driven document edit.

stevenobiajulu:
Fixes: #739

## Summary

Blanking or replacing a paragraph's visible text with `replace_text` silently destroyed any inline image (`w:drawing`), VML picture (`w:pict`), or embedded OLE object (`w:object`) caught in the replaced range — `success: true`, no warning, image gone.

This PR:

1. **Preserves embedded-content runs through `replaceParagraphTextRange()`** (`packages/docx-core/src/primitives/text.ts`) — the actual fix, covering `w:drawing`, `w:pict`, `w:object`, and `w:contentPart` (all relationship-bearing `EG_RunInnerContent`). An embedded-only run inside the replaced range stays in place untouched; content co-resident with replaced text is split into its own run (cloned formatting, original child order) before the text is removed. Preserved content stays **live** — never wrapped in `w:del` — because the caller's approved text match never covered it. Tracked deletions are emitted as **in-place segments** (one `w:del` per contiguous removed stretch, at its original position) so `rejectChanges()` restores the original content order exactly; ranges without embedded content keep the historical single-deletion emission unchanged. The paragraph-mark deletion emitted for a tracked full blanking is suppressed when content was preserved, so the paragraph that still visibly carries the image is not merged into its neighbor on accept.
2. **Adds `drawing`/`pict`/`object`/`contentPart` to `BARRIER_LOCALS` in `merge_runs.ts`** as defensive hardening (separately tested): on-open normalization can no longer fold an rPr-less image run into an adjacent text run, which was the path that made the issue's exact repro shape lossy.
3. Widens `makeDocxWithDocumentXml`/`openSession` `extraFiles` to accept `Buffer` values so package fixtures can carry real binary media (`createZipBuffer` already accepted them).

## Root cause — corrected from the issue

The issue blames `mergeRunsOnly` and proposes only the barrier change. That change alone would **not** fix the data loss. The primary defect is in `replaceParagraphTextRange()`: the run-removal loop detaches every `w:r` in the replaced range but only records a run in `removedRuns` when `getRunVisibleLength(runEl) > 0`. `visibleLengthForEl()` returns a length only for `w:t`/`w:tab`/`w:br` and `0` for everything else — so an image-only run is detached from the DOM and never re-emitted by either the clean or the tracked branch.

XML-level probe, run against origin/main sources (8f84295) and re-run against this branch (paragraph `[w:r "before"][w:r w:drawing r:embed="rId9"][w:r "after"]`, replace the full visible text with `"replacement"`, clean mode):

```
=== PRE-FIX (origin/main sources) ===
# replaceParagraphTextRange() ONLY, no merge:
REPLACE_ONLY_XML = <w:p><w:r><w:t>replacement</w:t></w:r></w:p>          <- drawing and rId9 GONE
# mergeRuns() ONLY:
MERGE_RESULT = {"runsMerged":2,"proofErrRemoved":0}
MERGE_XML = <w:p><w:r><w:t>before</w:t><w:drawing>...r:embed="rId9"...</w:drawing><w:t>after</w:t></w:r></w:p>   <- drawing PRESERVED

=== POST-FIX (this branch) ===
REPLACE_ONLY_XML = <w:p><w:r><w:drawing>...r:embed="rId9"...</w:drawing></w:r><w:r><w:t>replacement</w:t></w:r></w:p>
MERGE_RESULT = {"runsMerged":0,"proofErrRemoved":0}   <- barrier holds; drawing keeps its own run
```

The merge is a *contributing* path (it converts the issue's `[caption][image-run]` shape into a co-resident `[caption+drawing]` run that the removal loop then destroys), which is why it gets the separate hardening in (2). An independent Codex premise check on the unmodified tree reached the same conclusion: "Adding those embedded-content elements to the merge barrier list alone would **not** fix the loss."

## Premise-gate evidence (end-to-end through the real `replace_text` MCP path)

Full generated DOCX package: 3 paragraphs, the second `[w:r "Figure 1 caption."][w:r w:drawing r:embed="rId9"]`, `word/_rels/document.xml.rels` carrying `rId9 → media/image1.png`, and a real 70-byte PNG at `word/media/image1.png`. Flow: `open_document` → `replace_text` (old_string `"Figure 1 caption."`, new_string `""`) → `save`.

**Before the fix (origin/main @ 8f84295), both clean and tracked saves:**

```
--- BEFORE ---                          --- AFTER (clean AND tracked) ---
w:drawing count : 1                     w:drawing count : 0
a:blip count    : 1                     a:blip count    : 0
r:embed="rId9"  : 1                     r:embed="rId9"  : 0
rels has rId9   : true                  rels has rId9   : true   (orphaned)
media/image1.png: 70 bytes              media/image1.png: 70 bytes (orphaned)
replace_text: success:true, no warning; paragraph emitted with no drawing at all
```

**After the fix, same probe:** `w:drawing`/`a:blip`/`r:embed="rId9"` all `1` in both clean and tracked output; the paragraph retains the drawing as a live run.

## Tests (red → green verified by stashing the source fix)

- `packages/docx-core/src/primitives/text.test.ts` — 13 new preservation tests (**8 of the original 9 red pre-fix**; the before/after-guard passes by construction): drawing before / between / after text runs; drawing co-resident with text in the same run; with and without `w:rPr`; partial replacement and full blanking; clean and tracked; `w:pict`, `w:object`, and `w:contentPart`; tracked full blanking keeps the paragraph mark undeleted and the content outside `w:del`; and — added after peer review — structural-order regressions that call the real `rejectChanges()`/`acceptChanges()` primitives: reject restores `text → drawing → text` exactly, accept keeps the drawing live before the replacement text, and a span with two preserved objects emits three correctly-placed deletion segments.
- `packages/docx-core/test-primitives/merge_runs.test.ts` — 6 new barrier tests (**5 red pre-fix**), kept separate from the preservation tests, including a `contentPart` barrier and a control that plain text runs still merge in a paragraph containing a drawing run.
- `packages/docx-mcp/src/tools/replace_text_embedded_image.regression.test.ts` — new end-to-end package regression (**2 red pre-fix**): clean and tracked flows through the real `replace_text` tool assert the `w:drawing` node, the `rId9` relationship entry, byte-identical `word/media/image1.png`, drawing outside any revision wrapper, tracked `delText` for the caption, and a successful package reopen.

## Scope discipline

Pure preservation of intended behavior — no OpenSpec proposal needed, no new `UNSUPPORTED_EDIT` refusal, no new warning semantics. Surfacing a runtime warning for "embedded object preserved outside the replaced range" needs a core→MCP warnings channel that does not exist today (`replaceParagraphTextRange` returns `void`, and `preflightAiRevisionMutation` invokes the mutation callback before the `warnings` const is initialized); that is split into a follow-up issue rather than grown into this PR.

## Pre-submit

`npm run build && npm run lint:workspaces && npm run test:run && npm run check:spec-coverage && npm run check:conformance-citations && npm run check:conformance-doc` — exit 0.

## Peer review

Dynamic Codex review requested changes; the blocker (single terminal `w:del` broke `rejectChanges()` ordering around a preserved drawing) was accepted and fixed with in-place deletion segments in 26f2e01, and `w:contentPart` was added to the preservation/barrier sets per the review's schema finding. Full adjudication in the PR comments.

### DIFF
--- a/packages/docx-core/src/primitives/merge_runs.ts
+++ b/packages/docx-core/src/primitives/merge_runs.ts
@@ -3,7 +3,7 @@
  *
  * Safety barriers: never merge across fldChar/instrText, comment range,
  * bookmark, or tracked-change wrapper boundaries, and never merge runs
- * carrying embedded objects (drawing/pict/object).
+ * carrying embedded content (drawing/pict/object/contentPart).
  */
 
 import { XMLSerializer } from '@xmldom/xmldom';
@@ -37,13 +37,14 @@ const BARRIER_LOCALS = new Set([
   W.bookmarkEnd,
   'commentRangeStart',
   'commentRangeEnd',
-  // Embedded objects: folding an image/picture/object run into an adjacent
-  // text run makes later text edits treat the object as part of the text
-  // payload, so removing that text drags the object with it. Keep
-  // object-bearing runs unmerged (issue #739).
+  // Embedded content: folding an image/picture/object/content-part run into
+  // an adjacent text run makes later text edits treat the embedded content as
+  // part of the text payload, so removing that text drags the content with
+  // it. Keep embedded-content runs unmerged (issue #739).
   W.drawing,
   W.pict,
   W.object,
+  W.contentPart,
 ]);
 
 const TC_WRAPPER_LOCALS = new Set([
@@ -299,7 +300,7 @@ function mergeParagraphRuns(paragraph: Element, preserveRsidIdentity: boolean): 
  * - Comment range boundaries (commentRangeStart, commentRangeEnd)
  * - Bookmark boundaries (bookmarkStart, bookmarkEnd)
  * - Tracked-change wrapper boundaries (ins, del, moveFrom, moveTo)
- * - Embedded-object runs (drawing, pict, object) — see #739
+ * - Embedded-content runs (drawing, pict, object, contentPart) — see #739
  */
 export function mergeRuns(doc: Document, opts: MergeRunsOptions = {}): MergeRunsResult {
   const body = doc.getElementsByTagNameNS(OOXML.W_NS, W.body).item(0);
--- a/packages/docx-core/src/primitives/merge_runs.ts
+++ b/packages/docx-core/src/primitives/merge_runs.ts
@@ -2,7 +2,8 @@
  * merge_runs — merge adjacent format-identical runs within paragraphs.
  *
  * Safety barriers: never merge across fldChar/instrText, comment range,
- * bookmark, or tracked-change wrapper boundaries.
+ * bookmark, or tracked-change wrapper boundaries, and never merge runs
+ * carrying embedded objects (drawing/pict/object).
  */
 
 import { XMLSerializer } from '@xmldom/xmldom';
@@ -36,6 +37,13 @@ const BARRIER_LOCALS = new Set([
   W.bookmarkEnd,
   'commentRangeStart',
   'commentRangeEnd',
+  // Embedded objects: folding an image/picture/object run into an adjacent
+  // text run makes later text edits treat the object as part of the text
+  // payload, so removing that text drags the object with it. Keep
+  // object-bearing runs unmerged (issue #739).
+  W.drawing,
+  W.pict,
+  W.object,
 ]);
 
 const TC_WRAPPER_LOCALS = new Set([
@@ -104,7 +112,7 @@ function runRsidKey(run: Element): string {
     .join('|');
 }
 
-/** Check whether a run contains barrier content (fldChar, instrText). */
+/** Check whether a run contains barrier content (fldChar, instrText, embedded objects, …). */
 function runContainsBarrierContent(run: Element): boolean {
   let child = run.firstChild;
   while (child) {
@@ -291,6 +299,7 @@ function mergeParagraphRuns(paragraph: Element, preserveRsidIdentity: boolean): 
  * - Comment range boundaries (commentRangeStart, commentRangeEnd)
  * - Bookmark boundaries (bookmarkStart, bookmarkEnd)
  * - Tracked-change wrapper boundaries (ins, del, moveFrom, moveTo)
+ * - Embedded-object runs (drawing, pict, object) — see #739
  */
 export function mergeRuns(doc: Document, opts: MergeRunsOptions = {}): MergeRunsResult {
   const body = doc.getElementsByTagNameNS(OOXML.W_NS, W.body).item(0);
--- a/packages/docx-core/src/primitives/namespaces.ts
+++ b/packages/docx-core/src/primitives/namespaces.ts
@@ -182,6 +182,7 @@ export const W = {
   drawing: 'drawing',
   pict: 'pict',
   object: 'object',
+  contentPart: 'contentPart',
 
   // Revision wrappers
   del: 'del',
--- a/packages/docx-core/src/primitives/text.test.ts
+++ b/packages/docx-core/src/primitives/text.test.ts
@@ -6,6 +6,8 @@ import { OOXML, W } from './namespaces.js';
 import { SafeDocxError } from './errors.js';
 import { getDirectChildrenByName } from './dom-helpers.js';
 import { createRevisionContext, createRevisionIdState } from './track-changes-emitter.js';
+import { rejectChanges } from './reject_changes.js';
+import { acceptChanges } from './accept_changes.js';
 import { revisionEvidence, revisionEvidenceCases } from '../testing/revision-evidence.js';
 import {
   fldChar,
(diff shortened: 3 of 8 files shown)


stevenobiajulu:

Ran a dynamic Codex review (workspace-write sandbox, executed the test suites and the actual revision primitives; full methodology trace retained). **Verdict: changes requested — one blocker, one scope question, one minor.** Adjudication below; the fixes landed in 26f2e01.

### Blocker (accepted, fixed): single terminal `w:del` broke `rejectChanges()` ordering

Codex proved by executing the real `rejectChanges()` primitive — not by reading the diff — that the first preservation approach reordered content on reject. For original `text("before") → drawing → text("after")`, a tracked full replacement emitted one `w:del` **after** the preserved drawing, so reject-all restored `drawing → before → after`. That silently breaks the repo's reject-recovers-the-original-exactly guarantee, and no required check would have caught it (concatenated `delText` still matched).

**Fix (26f2e01):** tracked deletions around preserved embedded content are now emitted as **in-place segments** — one `w:del` per contiguous stretch of removed text, inserted at that stretch's original position, segment broken whenever preserved embedded content interrupts it. A run holding both text and embedded content is split into consecutive homogeneous runs in original child order first. Ranges without embedded content keep the historical single-deletion emission unchanged.

**New regression tests** (in `packages/docx-core/src/primitives/text.test.ts`) call the real `rejectChanges()` / `acceptChanges()` primitives and assert **structural order** via a paragraph content-sequence fingerprint, exactly as the review demanded:

- tracked replace of `text → drawing → text` emits `['del:before', '[drawing]', 'del:after', 'ins:X']`; reject restores `['before', '[drawing]', 'after']`;
- accept leaves `['[drawing]', 'X']`;
- a span with a drawing **and** a pict produces three correctly-placed deletion segments (`['del:aa', '[drawing]', 'del:bb', '[pict]', 'del:cc']`) and reject restores all five positions.

### Scope question (accepted, option (a)): `w:contentPart`

Codex is right that `w:contentPart` sits in `EG_RunInnerContent` with type `CT_Rel` in the vendored strict and transitional schemas — relationship-bearing run content exactly like the three handled forms. Rather than narrow the claim, 26f2e01 adds `contentPart` to the preservation set (renamed `EMBEDDED_CONTENT_LOCALS`) and to the `merge_runs` barrier list, each with its own test.

### Minor (rebutted): trailing whitespace at `session-test-utils.ts:55`

The flagged "trailing whitespace" is a carriage return: that file is checked into the repo **entirely CRLF** (verified `git show origin/main:packages/docx-mcp/src/testing/session-test-utils.ts | file -` → CRLF throughout; the untouched neighboring lines 54 and 56 also end `\r\n`). The edited line matches the file's existing convention; "fixing" it would produce a single LF line inside a CRLF file, i.e. a mixed-EOL file and a noisier diff. `git diff --check` flags only changed lines, which is why the pre-existing convention surfaces on exactly this line. Left as-is deliberately.

### Post-fix status

Full pre-submit chain (`build && lint:workspaces && test:run && check:spec-coverage && check:conformance-citations && check:conformance-doc`) re-run after 26f2e01, gated on exit code: **0**.

PR #879, opened 2026-08-16
fix(docx-render-verifier): pin corpus expectations

### LINKED ISSUE #865: docx-render-verifier: decide whether an empty expectedMarkupText may satisfy markupTextMatchesPdf
## Summary

`packages/docx-render-verifier/src/render.ts:163` computes `matched: missingTokens.length === 0 && unexplainedTokens.length === 0`. When both the expected markup projection and the extracted PDF text tokenize to zero tokens, that is trivially true and `markupTextMatchesPdf` reports a match.

**This is filed as a contract question, not as a confirmed bug.** The mechanism is real; whether it is *wrong* depends on a policy decision this repo has not made. Read the qualification before implementing anything.

## Mechanism (verified)

- `types.ts:104` — `expectedMarkupText` is an unrestricted required string, with no non-empty precondition.
- `render.ts:103` — it is tokenized with no non-empty guard.
- `render.ts:667` — PDF text reaches the binding.
- `render.ts:163` — `matched` is true when both the deficit and surplus sets are empty.
- `render.ts:698` — an overall pass is possible if the separate contrast gate also passes.

Public `verifyRenderedMarkup()` probe against a real DOCX with executed renderer adapters:

```text
{"status":"pass","markupTextMatchesPdf":true,"configuredContrastPassed":true,
 "textBinding":{"matched":true,"pageCount":1,"missingTokenSample":[],"unexplainedTokenSample":[],
 "declaredRevisionBoundaryCount":1,"revisionBoundaryNormalizationCount":0}}
```

## Why this may not be a defect

The empty/empty match is valid multiset equality, and it is not as loose as it first sounds:

- An empty PDF extraction does **not** pass when the expected projection has tokens — those become deficits and fail.
- Unexpected extracted tokens with an empty expectation become surpluses and fail.
- The independent color/contrast gate stops ordinary blank documents from passing overall.

So a green result requires *both* sides to be genuinely empty, which may be legitimate for a blank document.

Calling this a bug requires adopting a premise that is currently neither encoded nor documented: **that an empty logical markup projection is never valid, or must be explicitly authorized by the caller.** That is an API policy decision — plausibly an `allowEmptyMarkupText` opt-in — not a surgical fix.

A dynamic review recommended dropping this in its current form until that decision is made. Filed so the question is tracked rather than lost.

## Decision needed first

Is an empty `expectedMarkupText` (a) always a caller error, (b) valid only for a genuinely blank document, or (c) valid and unremarkable? The answer determines whether there is any work here.

## Acceptance criteria (only if the policy is adopted)

- [ ] Empty expected markup cannot produce `markupTextMatchesPdf: true` unless explicitly allowed.
- [ ] Empty PDF extraction with nonempty expected markup continues to fail as missing content.
- [ ] Nonempty PDF extraction with empty expected markup continues to fail as unexplained content.
- [ ] Legitimate blank-document behavior is defined and tested.
- [ ] An empty binding cannot accidentally satisfy required renderer evidence.

## Provenance

Surfaced as an unverified candidate while fixing #858, then verified dynamically against `1e9785a`; the output above is an actual probe result. The mechanism is confirmed; the defect status is not.

Refs #858

stevenobiajulu:
## Summary

- reject unreadable, empty, and whitespace-only expected-markup files as bounded case-level failures
- optionally pin the expected-markup file by raw-byte SHA-256 while preserving manifests that omit the pin
- document the optional pin and keep summary output limited to labels, tracked hashes, statuses, and bounded literal reasons

## Why

The private corpus already pins its tracked DOCX but previously trusted the separate untracked expectation file without a hash or non-empty guard. Local truncation or drift could therefore weaken the corpus gate without leaving any tracked evidence.

This does **not** close #865 and does not resolve the public `render.ts` empty-projection policy question. `render.ts`, `RenderRequest`, and matcher semantics are unchanged. #865 should remain open for the repository owner's `allowEmptyMarkupText` / empty-projection contract decision.

## Validation

- full required pre-submit gate passed: build, workspace lint, all workspace tests, spec coverage, conformance citations, and conformance document
- focused renderer-verifier suite: 56 tests passed
- independent dynamic Claude review executed the focused and full package suites, confirmed `render.ts` byte-identical, and verified no expectation text or local path reaches `summary.json`
- review fixes incorporated: bounded unreadable-expectation handling and discoverable optional-pin documentation

Refs #865

### DIFF
--- a/packages/docx-render-verifier/private-corpus/README.md
+++ b/packages/docx-render-verifier/private-corpus/README.md
@@ -8,6 +8,14 @@ gitignored) and keep both input paths outside the Safe DOCX worktree. A fully
 local manifest outside the worktree is also accepted because it cannot be
 committed to this repository.
 
+Set `expectedMarkupTextSha256` to the expected-markup file's 64-character
+lowercase SHA-256 when the local corpus should fail closed on expectation-file
+drift. Replace the example placeholder when pinning, or delete that property
+from the copied manifest when not pinning. The field is optional so existing
+private manifests remain compatible.
+Every expectation file must contain non-whitespace text whether or not this pin
+is present.
+
 The runner rejects a manifest or output directory that Git does not ignore,
 and rejects outputs under public `fixtures/`. Its committed summary format
 contains only a caller label, SHA-256, verdict, and bounded reason—never DOCX,
--- a/packages/docx-render-verifier/private-corpus/manifest.example.json
+++ b/packages/docx-render-verifier/private-corpus/manifest.example.json
@@ -7,6 +7,7 @@
       "trackedDocxPath": "/absolute/local/matter/tracked.docx",
       "expectedMarkupTextPath": "/absolute/local/matter/expected-markup.txt",
       "expectedTrackedSha256": "replace-with-64-lowercase-hex",
+      "expectedMarkupTextSha256": "optional-replace-with-64-lowercase-hex",
       "requireRender": true
     }
   ]
--- a/packages/docx-render-verifier/src/corpus.test.ts
+++ b/packages/docx-render-verifier/src/corpus.test.ts
@@ -53,4 +53,98 @@ describe('private renderer corpus', () => {
 
     await expect(runPrivateCorpus(manifestPath)).rejects.toThrow('must not be Git-tracked');
   });
+
+  itAllure('fails empty expectations without stopping other corpus cases', async () => {
+    await mkdir(externalDir, { recursive: true });
+    const docx = path.join(externalDir, 'tracked.docx');
+    const empty = path.join(externalDir, 'empty.txt');
+    const whitespace = path.join(externalDir, 'whitespace.txt');
+    const valid = path.join(externalDir, 'valid.txt');
+    await writeFile(docx, 'private binary surrogate');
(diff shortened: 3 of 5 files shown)


stevenobiajulu:
## Post-merge smoke: PASS

- Merged commit: `04de60894f05ac21c09cb4c43e4efcfb0cc4e565`
- Verified as part of combined `origin/main` at `6a356bcbf68f69582513c4e4fc607dd6991adb62` in a fresh detached worktree.
- Full repository gate passed: build, workspace lint, all workspace tests, spec coverage, conformance citations, and generated conformance documentation.
- Feature-specific coverage passed in the render-verifier suite (56 tests), including unreadable, empty, and SHA-256-mismatched expected-markup handling.
- GitHub `main` CI completed successfully, including CodeQL, workspace matrices, and the real-corpus gate.
- Vercel reports the production deployment completed successfully.
- Visual evidence: not applicable; the change affects corpus verification behavior and documentation only.
- Combined smoke log: `/private/tmp/automerge-smoke-878-879-20260816.log`

Cleanup: merged branch and dedicated worktree removed after successful verification.

Issue #886, opened 2026-08-17
docx-markdoc: allow internal and external rationales for one operation


stevenobiajulu:
## Problem

The Markdoc schema currently rejects more than one `rationale` for an operation with `MULTIPLE_RATIONALES`, even when the records have different visibility values.

A revision workflow may need both:

- an internal decision record retained only in the Markdoc source; and
- an external-facing explanation eligible for a Word comment.

The compiler already treats those visibility classes differently at export time, but the parser prevents representing both.

## Proposed behavior

- Enforce rationale uniqueness by `(operation, visibility)` rather than operation alone.
- Permit at most one `internal` and at most one `external-facing` rationale per operation.
- Continue rejecting duplicate rationales within the same visibility class.
- Preserve the existing safe default: internal rationale never renders unless the CLI-only dangerous capability is supplied.
- When compiling external comments only, prove that internal rationale text is absent from every output DOCX part.

## Acceptance criteria

- One internal plus one external-facing rationale for the same operation validates and compiles.
- Two internal rationales for the same operation fail before mutation.
- Two external-facing rationales for the same operation fail before mutation.
- Default/external-only output contains the external comment and no internal rationale content.
- Existing dangerous internal-export safeguards remain unchanged.

PR #888, opened 2026-08-17
fix(docx-markdoc): preserve paired rationale consumers

stevenobiajulu:
## Summary
- group attribution discovery by operation so dangerous review exports can attach both visibility comments without marker overlap
- preserve paired rationale records in structured edit exports instead of collapsing them through an operation-only map
- add direct tests for dual native comments and lossless structured export

## Why
The advisory review on #887 correctly found that parser support alone was incomplete: the dangerous dual-comment path and structured export still assumed one rationale per operation.

## Validation
- mandatory repository pre-submit suite passed
- OpenSpec strict validation passed
- confidential two-mode real-document smoke passed locally; no fixture identity, contents, metrics, or artifacts are included

Ref #886
Follow-up to #887

### DIFF
--- a/openspec/changes/allow-dual-visibility-rationales/proposal.md
+++ b/openspec/changes/allow-dual-visibility-rationales/proposal.md
@@ -1,0 +1,20 @@
+# Change: Allow paired rationale visibility records
+
+## Why
+
+An operation may need a private decision record and a distinct external-facing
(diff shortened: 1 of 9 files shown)


stevenobiajulu:
Superseded by the clean current-main branch in #889; no force-push was used after review began.

Issue #945, opened 2026-08-23
fix(docx-compare): lift non-text deletions to schema-valid revision boundaries


stevenobiajulu:
## Consequence first

The tagged serializer emits `w:del` below schema-illegal parents when a side-only non-text atom is found inside an aligned structural container. On the public ILPA comparison this produces **16 independent schema diagnostics**:

- 13 `w:r > w:del > w:footnoteReference`
- 2 `w:r > w:del > w:instrText`
- 1 `w:drawing > w:del > wp:anchor`

Tracked run-content deletions must be represented at a legal revision boundary (typically by splitting/cloning the containing run and wrapping that run), not by placing the revision wrapper directly around an arbitrary leaf.

## Reproduction on current `main`

Reproduced at `6a7bb7a35178153cc0e971285b2dc7421a8478df` using only the repository's public ILPA fixtures:

- `tests/test_documents/redline/ILPA-Model-Limited-Partnership-Agreement-WOF_v2.docx`
  - SHA-256 `7880ad250513d23be064a5c71ef79aecb2bf3d870c45fa070256d5388ac74455`
- `tests/test_documents/redline/ILPA-Model-Limited-Parnership-Agreement-Deal-By-Deal_v1.docx`
  - SHA-256 `46be28b42853da50eba3e24df944d04dd34893f56539258d53bd69b4b33547d7`

Compare those files with `compareDocuments()` using a fixed author/date, extract `word/document.xml`, apply the repository's MCE preprocessing, and validate with:

```sh
xmllint --noout --nonet \
  --schema spec-compliance/ecma-376/validation/wml-document-transitional.xsd \
  document.xml
```

A DOM classification of every `w:del` whose direct parent is `w:r` or `w:drawing` returns exactly 16 elements. Their generated revision IDs are:

- drawing/anchor: `1347`
- footnote references: `1957`, `2225`, `5678`, `6074`, `8927`, `10640`, `12155`, `12250`, `12279`, `16575`, `19327`, `20064`, `20094`
- field instruction fragments: `9874`, `16929`

Representative invalid shapes:

```xml
<w:r>
  <w:del ...><w:footnoteReference w:id="..."/></w:del>
</w:r>

<w:r>
  <w:del ...><w:instrText>...</w:instrText></w:del>
</w:r>

<w:drawing>
  <w:del ...><wp:anchor>...</wp:anchor></w:del>
</w:drawing>
```

## Root-cause localization

[...]

PR #953, opened 2026-08-25
fix(docx-comparison): lift non-text run revisions

stevenobiajulu:
## Outcome

Word now opens the public minimized comparison cleanly, and tagged comparison no longer emits revision wrappers directly inside `w:r` or `w:drawing` for changed field instructions, footnote references, or DrawingML anchors.

## What changed

- keep differing non-text run content as opaque original/revised runs during tagged-tree construction;
- continue using the existing text-refinement path for ordinary `w:t`, `xml:space`, and volatile rendered-page-break behavior;
- translate deleted `w:instrText` to `w:delInstrText` without changing move-from field vocabulary;
- add minimized regression coverage for field instructions, footnote references, and anchored DrawingML;
- add a two-direction public ILPA corpus gate for illegal `del`/`ins`/`moveFrom`/`moveTo` parentage.

## Verification

- Red-first: the three new minimized tests failed on the prior implementation; 34 existing tests passed.
- Focused: 58/58 tagged construction + serializer tests passed.
- Public ILPA corpus: 2/2 tests passed in both directions, including exact existing projection checks and the new four-kind boundary assertion.
- Fixed minimized DOCX: vendored Transitional schema validation passed.
- Fixed minimized + full public ILPA outputs: zero revision wrappers directly parented by `w:r` or `w:drawing` across all four revision kinds.
- Microsoft Word 16.112.1: fixed minimized comparison opened cleanly with no repair/recovery dialog. The prior output produced “Word found unreadable content.”
- Mandatory pre-submit passed:
  `npm run build && npm run lint:workspaces && npm run test:run && npm run check:spec-coverage && npm run check:conformance-citations && npm run check:conformance-doc`
- Dynamic Claude review executed the focused tests and schema/classifier probes: **APPROVE**, no findings.

## Scope boundary

The full ILPA comparison has zero #945-class wrappers, but Word still reports repairs and the complete document still has vendored-schema errors belonging to the separately tracked move-range and section-property defects (#941 and #944). This PR intentionally does not broaden into those concerns.

Only repository-public fixtures were used. No private or purchased-template documents were accessed.

Fixes #945

Issue #944, opened 2026-08-23
docx-compare: section-property revisions contain header/footer references, so the output fails the ECMA-376 schema


stevenobiajulu:
## Summary
When a section with header/footer relationships changes, the comparison writes `w:headerReference` / `w:footerReference` into the historical `w:sectPr` nested under `w:sectPrChange`. That inner element is `CT_SectPrBase`, which cannot contain header/footer references, so the document fails schema validation. Word's behaviour on these files is unverified; schema-invalid output risks a repair prompt on a client redline.

Not an ordering defect and not part of #941's paragraph-move scope.

## Reproduction
Checked on main @ fb0bd3af: still reproduces.

Public fixtures:
- `tests/test_documents/redline/ILPA-Model-Limited-Partnership-Agreement-WOF_v2.docx` (SHA-256 `7880ad250513d23be064a5c71ef79aecb2bf3d870c45fa070256d5388ac74455`)
- `tests/test_documents/redline/ILPA-Model-Limited-Parnership-Agreement-Deal-By-Deal_v1.docx` (SHA-256 `46be28b42853da50eba3e24df944d04dd34893f56539258d53bd69b4b33547d7`)

Compare them with `compareDocuments()` using a fixed author/date, then run `node scripts/check_emitted_document_schema.mjs <dir-with-output.docx>` (or extract `word/document.xml`, apply MCE preprocessing and run `xmllint --noout --nonet --schema spec-compliance/ecma-376/validation/wml-document-transitional.xsd document.xml`).

Representative emitted shape:

```xml
<w:sectPrChange w:id="1480" w:author="Public Investigation" w:date="2026-08-23T00:00:00Z">
  <w:sectPr>
    <w:footerReference w:type="even" r:id="rId56"/>
    <w:footerReference w:type="default" r:id="rId50"/>
    <w:pgSz w:w="12240" w:h="15840"/>
  </w:sectPr>
</w:sectPrChange>
```

Originally reproduced at `6a7bb7a35178153cc0e971285b2dc7421a8478df` (other first failures in revisions `1915`, `20075`, `20161`, `20446`, `20485`) and again on `ca618780` and the #979 branch `efa602b2` on 2026-09-17: both source files validate; the tracked output and its native Reject fail on historical section header/footer references; native Accept validates.

## Expected
[...]

Issue #941, opened 2026-08-23
fix(docx-compare): emit schema-valid paragraph move markup


stevenobiajulu:
## Problem

The tagged comparison engine emits paragraph move markup that fails the repo's
vendored ECMA-376 Transitional WML schema gate. This is observable in a real
`compareDocuments` result, not only the lower-level tagged publication helper.

For a two-paragraph reorder with `detectMoves: true`, the output includes:

```xml
<w:moveFromRangeStart w:id="1" w:name="taggedMove1"/>
<w:moveFrom w:id="1" w:author="Range Statistics" w:date="2026-08-23T12:00:00Z">
  <w:p>...</w:p>
</w:moveFrom>
<w:moveFromRangeEnd w:id="1" w:name="taggedMove1"/>
```

with an analogous `moveTo` sequence.

`scripts/check_emitted_document_schema.mjs` reports:

- `moveFromRangeStart` / `moveToRangeStart` are missing required `w:author` and
  `w:date` attributes;
- `moveFromRangeEnd` / `moveToRangeEnd` carry disallowed `w:name` attributes;
- block-level `w:p` is emitted inside `w:moveFrom` / `w:moveTo`, where the
  vendored schema rejects it.

This shape is not listed in `coverage/emitted-schema-known-failures.json`.

## Acceptance criteria

- Emit paragraph moves using an ECMA-376-conformant structure accepted by the
  repo's vendored Transitional WML schemas.
- Add the appropriate `@conformance ECMA-376 edition 5, Part 1 § ...` citation
  to the implementation and Allure conformance metadata to the tests after the
  exact governing section is verified.
- Preserve a balanced, ID-consistent move source/destination pair.
- Preserve Accept All and Reject All projections and Word/LibreOffice opening
  behavior.
- Add an end-to-end generated-DOCX schema fixture; do not merely allow-list the
  current invalid output.
- Confirm the fixed markup in Microsoft Word and at least one independent
  consumer oracle.

Premise independently reproduced at `065eef59` on 2026-08-23 with a synthetic
public DOCX and:

```bash
node /tmp/revscratch/e2e.mjs
node scripts/check_emitted_document_schema.mjs /tmp/revscratch/e2e
```

The schema command failed with the errors listed above. No duplicate issue was
found by searching for `moveFromRangeStart author date schema` or move-schema
terms.

stevenobiajulu:
[...]

PR #954, opened 2026-08-25
test(docx-comparison): enforce revision topology corpus-wide

stevenobiajulu:
## Summary

- promote the `w:del` / `w:ins` / `w:moveFrom` / `w:moveTo` parent-topology check into the shared strategy-differential safety harness
- fail every checked-in, ILPA, and public NVCA corpus row if a revision wrapper is nested directly under `w:r` or `w:drawing`
- record the four deterministic reviewed-manifest deltas caused by #953 (`checked-in/ILPA` and three public NVCA paragraph-deletion rows)
- reuse the shared checker in the bidirectional ILPA regression test

## Why

PR #953 fixed issue #945 and merged after all required checks passed, but its advisory real-corpus job correctly detected an intentional stale manifest. Claude's post-fix peer review also identified that the illegal-parent assertion was scoped only to ILPA. This follow-up records the reviewed baseline and turns that invariant into a corpus-wide fail-closed safety gate.

The three NVCA byte deltas are exactly `w:instrText` to `w:delInstrText` conversions (2, 1, and 2 nodes). ILPA converts 112 such nodes, removes all 32 illegal nested insertion/deletion wrappers, and changes one drawing replacement from a separate property change to an opaque delete/insert run pair.

## Verification

- full real-corpus gate: 3 files, 21 tests passed (7 public NVCA documents, 23 strategy rows, both ILPA directions)
- focused strategy/construction/serializer tests: 66 passed
- mandatory pre-submit chain passed: build, workspace lint, all workspace tests, spec coverage, conformance citations, and conformance document
- deterministic updater check: `strategy differential manifest is current`
- Claude dynamic peer review: core fix approved; its corpus-wide topology-coverage finding is addressed here

Ref: #945
Follow-up to: #953

### DIFF
--- a/packages/docx-compare/src/integration/strategy-differential-harness.ts
+++ b/packages/docx-compare/src/integration/strategy-differential-harness.ts
@@ -91,6 +91,7 @@ export interface StrategyEvidence {
   integrity: {
     fieldIssues: string[];
     bookmarkIssues: string[];
+    revisionTopologyIssues: string[];
     revisionIdIssues: string[];
     moveBalanceIssues: string[];
   };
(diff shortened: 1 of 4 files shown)


stevenobiajulu:
## Post-merge smoke: PASS (Word oracle currently indeterminate)

Verified the product change on current `main` at `482b0df5c8793005eac9c3f83fe8bcfb784de293`; the only commit after this PR is the test-only topology coverage follow-up #955.

- Exact merged tree builds successfully.
- Built CLI comparison of the public/anonymized minimized pair succeeds.
- Emitted DOCX validates against the Transitional WML schema (1/1).
- The shared four-kind illegal revision-topology classifier returns `[]`.
- Focused topology regression passes 9/9.
- CI real-corpus comparison gates pass for all 21 reviewed pairs; all required workspace, static, security, packaging, coverage, and advisory LLM gates are green.

Microsoft Word oracle caveat: the isolation-safe probe returned `INDETERMINATE` for both the new output and a previously Word-clean control because Word currently reports 15 open documents but can resolve only 14, consistent with a blocking dialog belonging to a user document. It did not close, save, or alter any existing user document. This is not evidence of an artifact failure; the exact production change had already opened cleanly in Word before merge. LibreOffice and corpus regression coverage are green.

The unrelated full-ILPA Word repair findings remain tracked separately in #941 and #944.

